PR #26, opened 2020-12-08
Adding Action to Apply carvel-triage Label for Closed/Reopened Issues

danielhelfand:
As part of the new triage process for carvel repos, when an issue is closed but reopened or a comment is made on a closed issue, the `carvel-triage` label should be applied.

This pull request adds an action to automatically apply the `carvel-triage` label for the above cases.

### DIFF
--- a/.github/workflows/closed-issue-label.yml
+++ b/.github/workflows/closed-issue-label.yml
@@ -1,0 +1,17 @@
+name: Closed issue labeling
+on:
+  issues:
+    types: ['reopened']
+  issue_comment:
+    types: ['created']
+
+jobs:
+  add-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: docker://k8slt/closed-issue-label-action@sha256:84360552559e0de4dfaaeefcf1ea0a1070a6250621f98539caf3eaad9395feb7
+        if: github.event_name == 'issues' || ( github.event_name == 'issue_comment' && github.event.issue.state == 'closed')
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          ignore-comments: true
+          default-labels: '["carvel-triage"]'


.github/workflows/closed-issue-label.yml
```diff
@@ -0,0 +1,17 @@
...
+    types: ['reopened']
+  issue_comment:
+    types: ['created']
+
+jobs:
+  add-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: docker://k8slt/closed-issue-label-action@sha256:84360552559e0de4dfaaeefcf1ea0a1070a6250621f98539caf3eaad9395feb7
+        if: github.event_name == 'issues' || ( github.event_name == 'issue_comment' && github.event.issue.state == 'closed')
```
StevenLocke, line 13: I think `github.event_name == 'issues'` is too broad. 

The [reference I found](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#issues) implies this would run on a lot of actions, including editing, assigning, and labelling (like stale?). Does the `on` key protect against these?

danielhelfand: Yeah, I think it is a fair call out and something I am now wondering about (whether `on` covers this case). We can test it out and see if there is a simpler approach here.

joaopapereira: if you look at this line: https://github.com/vmware-tanzu/carvel-imgpkg/pull/26/files#diff-1332567eb0d1b394868f6c46c3ab3794422f1a7a5f59a7ec0cc61fd5516a0d27R4
you will see that we only trigger this job when an issue is reopenned

danielhelfand: Do we necessarily need the `if` statement if `on` already covers this case? More trying to understand at this point if it's already covered.

joaopapereira: The if is protecting us from the right side of the or statement.
We want to trigger on reopening issues or if the event is a new comment on a closed issue

danielhelfand: Yeah, I see. I think it makes sense then.

StevenLocke (approved):

PR #28, opened 2020-12-09
Update the file excluding flag to "--file-exclusion"

StevenLocke:
* Deprecate the old flag
* Change the wording of the help text
* Add a test to ensure it works

### DIFF
--- a/pkg/imgpkg/cmd/file_flags.go
+++ b/pkg/imgpkg/cmd/file_flags.go
@@ -10,11 +10,14 @@ import (
 type FileFlags struct {
 	Files []string
 
-	FileExcludeDefaults []string
+	ExcludedFileBasenames []string
 }
 
 func (s *FileFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().StringSliceVarP(&s.Files, "file", "f", nil, "Set file (format: /tmp/foo, -) (can be specified multiple times)")
 
-	cmd.Flags().StringSliceVar(&s.FileExcludeDefaults, "file-exclude-defaults", []string{".git"}, "Excluded file paths by default (can be specified multiple times)")
+	cmd.Flags().StringSliceVar(&s.ExcludedFileBasenames, "file-exclude-defaults", []string{".git"}, "Excluded file paths by default (can be specified multiple times)")
+	cmd.Flags().MarkDeprecated("file-exclude-defaults", "use '--file-base-exclude' instead")
+
+	cmd.Flags().StringSliceVar(&s.ExcludedFileBasenames, "file-base-exclude", []string{".git"}, "Exclude all files whose base name matches (format: bar.yaml) (can be specified multiple times)")
 }
--- a/pkg/imgpkg/cmd/file_flags.go
+++ b/pkg/imgpkg/cmd/file_flags.go
@@ -10,14 +10,14 @@ import (
 type FileFlags struct {
 	Files []string
 
-	ExcludedFileBasenames []string
+	ExcludedFilePaths []string
 }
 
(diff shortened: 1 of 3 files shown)


cari-lynn (approved):

pkg/imgpkg/cmd/file_flags.go
```diff
@@ -10,11 +10,14 @@ import (
...
 }
 
 func (s *FileFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().StringSliceVarP(&s.Files, "file", "f", nil, "Set file (format: /tmp/foo, -) (can be specified multiple times)")
 
-	cmd.Flags().StringSliceVar(&s.FileExcludeDefaults, "file-exclude-defaults", []string{".git"}, "Excluded file paths by default (can be specified multiple times)")
+	cmd.Flags().StringSliceVar(&s.ExcludedFileBasenames, "file-exclude-defaults", []string{".git"}, "Excluded file paths by default (can be specified multiple times)")
+	cmd.Flags().MarkDeprecated("file-exclude-defaults", "use '--file-base-exclude' instead")
+
+	cmd.Flags().StringSliceVar(&s.ExcludedFileBasenames, "file-base-exclude", []string{".git"}, "Exclude all files whose base name matches (format: bar.yaml) (can be specified multiple times)")
```
cppforlife: it seems that https://github.com/vmware-tanzu/carvel-imgpkg/blob/87884a5de02e2d4cb0e409cc0e5bf64f88129bab/pkg/imgpkg/image/tar_image.go#L73 is actually using a relative path to the "root directory". so if .git was within a nested dir, exclude would have to be `nested-dir/.git`.

`--file-exclusion` might be a better name imho (since file in our uses implies a path).

joaopapereira: I believe that this help should inform the user that this is a relative path, something like:
`Exclude all files whose base name matches (format: bar.yaml) relative to the bundle root (can be specified multiple times)`

cppforlife: "base name" typically refers to filename without the path components (eg https://golang.org/pkg/path/filepath/#Base).

danielhelfand:
Something I noticed while running this locally. Is this supposed to work with `imgpkg push -b` as well? I actually could only use it with `-i`.

StevenLocke:
It's definitely supposed to work with -b as well.

This is my bundle test directory. It contains a .imgpkg folder and ./imgpkg/images.yml file marking it as a bundle.
```
tree -a ../bundleTest
 
../bundleTest
├── .imgpkg
│   └── images.yml
├── a.txt
├── b.txt
└── nested
    └── b.txt
```

If I run the command `imgpkg push -b abc/123 -f ../bundleTest --file-exclusion b.txt`

I get the expected files listed (and not bundleTest/b.txt)

```
dir: .
dir: .imgpkg
file: .imgpkg/images.yml
file: a.txt
dir: nested
file: nested/b.txt
```

danielhelfand:
I was using the bundle under examples [here](https://github.com/vmware-tanzu/carvel-imgpkg/tree/cf7c0bcf384705caa71bf510515a00e6ff39a7a4/examples/basic-bundle) and couldn't seem to get `--file-exclusion` to work. I was trying to just run against the `.imgpkg` dir to avoid going through the push:

```
./imgpkg push -b k8slt/dhelfand-bundle -f examples/basic-bundle/ --file-exclusion examples/basic-bundle/.imgpkg
```

danielhelfand:
Oh I see. Just the base name is the reason. Works as expected. My bad.

danielhelfand (approved):
lgtm

PR #31, opened 2020-12-10
Enforce the ImagesLock `kind` when validating bundles

StevenLocke:
Unsure on the error message here. Too verbose? It outputs the entire path to your images.yml file.

### DIFF
--- a/pkg/imgpkg/cmd/push.go
+++ b/pkg/imgpkg/cmd/push.go
@@ -236,6 +236,9 @@ func (o *PushOptions) validateBundle(registry ctlimg.Registry) error {
 	if err != nil {
 		return fmt.Errorf("Unmarshalling image lock: %s", err)
 	}
+	if imgLock.Kind != lf.ImagesLockKind {
+		return fmt.Errorf("Invalid `kind` in lockfile at %s. Expected: %s, got: %s", filepath.Join(bundlePaths[0], lf.ImageLockFile), lf.ImagesLockKind, imgLock.Kind)
+	}
 
 	bundles, err := imgLock.CheckForBundles(registry)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/imgpkg/cmd/push.go
```diff
@@ -236,6 +236,9 @@ func (o *PushOptions) validateBundle(registry ctlimg.Registry) error {
 	if err != nil {
 		return fmt.Errorf("Unmarshalling image lock: %s", err)
 	}
+	if imgLock.Kind != lf.ImagesLockKind {
```
cppforlife, line 239: should also check apiVersion.

joaopapereira: I think this was purposeful because there is the idea that we will want to update the apiVersion by changing the domain from `imgpkg.k14s.io/*` to `imgpkg.carvel.dev/*`.
If we add the API check right now we will have to change it.

gcheadle-vmware:
> Unsure on the error message here. Too verbose? It outputs the entire path to your images.yml file.

I think that the verbose error message is appropriate here since `images.yml` is kept in the hidden `.imgpkg` dir. Makes it much easier to locate for the user and make a change.

pkg/imgpkg/cmd/push.go
cppforlife: this is new code, why wouldnt be using imgpkg.k14s.io?

separately from that, the whole point of having apiVersion is so that tool can ensure that it's seeing input in format it supports. input today is in `imgpkg.carvel.dev/v1alpha1` format, so we should check on that. if you are saying that it's in both `imgpkg.k14s.io/v1alpha1` to `imgpkg.carvel.dev/v1alpha1` then we should check on it being either of them. (also note there is no such thing as `/*`, has to be explicit version per points above).

joaopapereira: Sounds good.
Do you think we can merge this PR as is and add a new one to check the API version?

cppforlife: whatever makes sense for yall. as long as apiVersion check makes it in, im ok.

PR #41, opened 2020-12-14
Update help text for copy command

gcheadle-vmware:
* include example moving a bundle to another repo


StevenLocke:
The need for this help text was demonstrated during usability testing. Thanks for conducting and making improvements!

In terms of content: 
Should we continue to use `dkalinin` as the example user? Maybe `carvel` should be the replacement? Alternatively, we could host a simple bundle/image on DockerHub and the example commands could be entirely useable. Right now, trying to copy either app1-bundle or app1-image results in an (expected) error.

Edit: After discussion with @gcheadle-vmware, we decided that keeping a valid username is better than referencing a docker user which is not under our control (carvel). We may want to investigate uploading something to `dkalinin/app1-*` but that can be addressed outside of this PR.

Looks good to me!

PR #43, opened 2020-12-18
Suppress output when skipping file link or symlink

gcheadle-vmware:
Removed repetitive "TODO Skipping file link"...  when encountering a file link or Symlink.

Q: Is there a situation where the user would need to know that the symlinks are being skipped? If so, should we document this behavior? or include a "Skipping all file and sym links" in the output when these links are detected in a layer?

### DIFF
--- a/pkg/imgpkg/image/dir_image.go
+++ b/pkg/imgpkg/image/dir_image.go
@@ -143,8 +143,7 @@ func (i *DirImage) extractTarEntry(header *tar.Header, input io.Reader) error {
 		}
 
 	case tar.TypeLink, tar.TypeSymlink:
-		// TODO currently not implemented
-		fmt.Printf("TODO Skipping file link '%s'\n", header.Name)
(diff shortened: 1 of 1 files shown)


cari-lynn (approved):

joaopapereira (commented):
Answering your question I believe we should inform the user that we skipped symlinks.
But to decrease the noise maybe it makes sense to do it only once.

cari-lynn:
> Answering your question I believe we should inform the user that we skipped symlinks.
> But to decrease the noise maybe it makes sense to do it only once.

Does the imgpkg user need this information to complete their workflow? What reasons would they need to know about symlinks?

cari-lynn:
For context, we spoke to João and we thought it may come up where a user has symlinks in their manifests in a bundle, but we felt like this was an unusual case. So we are deferring adding any output until it comes up that this is something users want.

PR #47, opened 2020-12-22
Correct translation of OCI Images URL on copy

joaopapereira:
This PR contains the following changes:
1. Extract functionality that collects URLs from Bundle and Image lock files to the `lockfiles` package. The objective was to enable us to further refactor the cmd/copy.go and allow us to create more focused tests for this collection of URLs
1. Use `Reference` instead of a plain string URL on cmd/copy.go. This corrects the problem of translation of URLs from `k8slt/basic-bundle` to `index.docker.io/k8slt/basic-bundle`. Previously we were comparing translated with not translated URL.
1. Created interface `ImageRetriever` to create a more clear separation between `lockfiles` package and `image` package. The only tidy bit missing for a complete decoupling is a constant that is being used in both the packages


joaopapereira:
Closing this PR because #48 solves this issue and also refactors the code in a better way

PR #52, opened 2021-01-12
Improve message  given by pull command

joaopapereira:
The pull command when finishes pulling the images gave the following message:
```
All images found in bundle repo; updating lock file: /tmp/local-bundle/.imgpkg/images.yml
```
We noticed this was the source of some confusion so this PR will update that message.

Also, we fixed the TODO related to the usage of the UI and the knowledge of command flags inside the bundle package


pkg/imgpkg/cmd/pull.go
```diff
@@ -75,7 +75,16 @@ func (o *PullOptions) Run() error {
...
+		imagesLockWriter := &ImagesLockWriter{ui: o.ui}
+		err := bundle.NewPullBundle(bundleRef, registry, imagesLockWriter).Pull(o.OutputPath, o.ui)
+		if err == nil {
+			return nil
+		}
+
+		if bundle.IsNotBundleError(err) {
+			return fmt.Errorf("Expected bundle image but found plain image (hint: Did you use -i instead of -b?)")
+		}
+		return err
```
cppforlife: i think err != nil is more readable/less surprsing:

```
if err != nil {
	if bundle.IsNotBundleError(err) {
		return fmt.Errorf("Expected bundle image ...")
	}
	return err
}

return nil
```

pkg/imgpkg/cmd/images_lock_writer.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 github.com/cppforlife/go-cli-ui/ui.UI
+
+func NewImagesLockWriter(ui ui.UI) *ImagesLockWriter {
+	return &ImagesLockWriter{ui: ui}
+}
+
+type ImagesLockWriter struct {
```
cppforlife: Instead of having a do-er object (i.e. doesnt really represent a real noun), it would be good to consider what if this object was ImagesLock (a bundle specific one, since it's in bundle package). you could have something like this:

```
NewImagesLock(*lockconfig.ImagesLock, imgRetriever, relativeToRepo) bundle.ImagesLock
ImagesLock#WriteToPath(path, ui)
ImagesLock#localize(*lockconfig.ImagesLock) (bool, error)
```

this way we are encapsulating behaviour of what it means to be a bundle's imagelock in one class. this also avoid backward ref to the bundle.

cppforlife: (i dont think we need to depedency inject it into Bundle but if we really want to we can create tiny factory/function that knows how to spin one up).

pkg/imgpkg/bundle/bundle.go
```diff
@@ -67,7 +48,11 @@ func (o *Bundle) Pull(outputPath string, ui ui.UI) error {
 		return fmt.Errorf("Extracting bundle into directory: %s", err)
 	}
 
-	err = o.rewriteImagesLockFile(outputPath, ui)
+	imagesLock, err := lockconfig.NewImagesLockFromPath(filepath.Join(outputPath, ImgpkgDir, ImagesLockFile))
+	if err != nil {
+		return err
+	}
+	err = NewImagesLock(&imagesLock, o.imgRetriever, o.Repo()).WriteToPath(outputPath, ui)
```
cppforlife, line 56: nit: newline before.

joaopapereira: done

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -0,0 +1,100 @@
...
+	} else {
+		ui.BeginLinef("The bundle repo (%s) is hosting every image specified in the bundle's Images Lock file (.imgpkg/images.yml)\n", o.relativeToRepo)
+		ui.BeginLinef("\nUpdating all images in the ImagesLock file: %s\n", imagesLockPath)
+		ui.BeginLinef("+ Changing all image registry/repository references in %s to %s\n", imagesLockPath, o.relativeToRepo)
+	}
+	return o.imagesLock.WriteToPath(imagesLockPath)
+
+}
+
+func (o *ImagesLock) localizeImagesLock() (bool, error) {
```
cppforlife: i see that we are using this private method in Bundle object. i would recommend turning this into `Localized() (*lockconfig.ImagesLock, bool, error)` i.e just return new lockconfig.ImagesLock with localized images.

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if skipped {
+		ui.BeginLinef("One or more images not found in bundle repo; skipping lock file update\n")
+	} else {
+		ui.BeginLinef("The bundle repo (%s) is hosting every image specified in the bundle's Images Lock file (.imgpkg/images.yml)\n", o.relativeToRepo)
+		ui.BeginLinef("\nUpdating all images in the ImagesLock file: %s\n", imagesLockPath)
+		ui.BeginLinef("+ Changing all image registry/repository references in %s to %s\n", imagesLockPath, o.relativeToRepo)
+	}
+	return o.imagesLock.WriteToPath(imagesLockPath)
+
```
cppforlife, line 44: nit: extra newline

PR #48, opened 2020-12-22
Refactor of pull/push/copy

cppforlife:
do not merge

PR #56, opened 2021-01-14
Refactor e2e tests

joaopapereira:
After #48 we talked about the possibility of also refactor the tests.


test/e2e/assets.go
```diff
@@ -0,0 +1,140 @@
...
+	"strings"
+	"testing"
+)
+
+type assets struct {
+	t              *testing.T
+	createdFolders []string
+}
+
+func (a assets) simpleAppDir() string {
```
cppforlife: even though assets object is private to the package, its methods arguably should be public (ie public to users within this package). otherwise it reads like eveything is accessing privates of the test supporting objects.

test/e2e/bundle_factory.go
```diff
@@ -0,0 +1,83 @@
...
+	assets       *assets
+	t            *testing.T
+	bundleFolder string
+}
+
+func newBundleDir(t *testing.T, assets *assets) bundleFactory {
+	return bundleFactory{assets: assets, t: t}
+}
+
+func (b *bundleFactory) isImageABundle(imgRef string) bool {
```
cppforlife: this method seems to be extracted only for two tests: TestBundlePushPullAnnotation and TestPushWithFileExclusion. arguably TestPushWithFileExclusion shouldnt be testing that image is a bundle (covered already in first test), hence i am thinking it might be better to inline this into TestBundlePushPullAnnotation and remove reference to that in TestPushWithFileExclusion.

test/e2e/assertion.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/k14s/imgpkg/pkg/imgpkg/imagetar"
+	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
+)
+
+type assertion struct {
+	t *testing.T
+}
+
+func (a *assertion) isBundleLockFile(path, bundleImgRef string) {
```
cppforlife: this seems to be main guts of the TestBundleLockFile test. afaik isBundleLockFile is only used in that test. i would recommend inlining it to keep "primary test" directly within Test function.

test/e2e/copy_from_bundle_image_test.go
```diff
@@ -0,0 +1,213 @@
...
+	// create a bundle with ref to generic
+	imageLockYAML := fmt.Sprintf(`---
+apiVersion: imgpkg.carvel.dev/v1alpha1
+kind: ImagesLock
+images:
+- image: %s%s
+`, env.Image, imageDigest)
+	bundleDir := env.BundleFactory.createBundleDir(bundleYAML, imageLockYAML)
+
+	// create bundle that refs image and a random tag based on time
```
cppforlife, line 33: what do you think of switching from comments to https://github.com/vmware-tanzu/carvel-kapp/blob/develop/test/e2e/create_update_delete_test.go#L67 for separation? (possibly as a follow up PR)

test/e2e/helpers.go
```diff
@@ -73,31 +73,21 @@ kind: ImagesLock
...
 	match := regexp.MustCompile("@(sha256:[0123456789abcdef]{64})").FindStringSubmatch(out)
 	if len(match) != 2 {
 		t.Fatalf("Expected to find digest in output '%s'", out)
 	}
 	return match[1]
 }
+
+var letterRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
+
+func randStringRunes(n int) string {
```
cppforlife: randStringRunes -> randString? (not sure what rune is describing to the consumer of this function).

test/e2e/copy_from_bundle_image_test.go
joaopapereira: I like the way the tests would read, looking from a 30k feet view because it would allow us to be more explicit on what each test is really testing.
Maybe this exercise would display even other places where we are doing redundant test assertions.
Nevertheless, I think we should do this in a separate PR.

PR #58, opened 2021-01-15
Adding golint and goheader linters via golangci-lint

danielhelfand:
This pull requests adds a linting tool ([golangci-lint](https://github.com/golangci/golangci-lint)) to `imgpkg`. `golangci-lint` allows users to add go linters to a project via a config file (`.golangci.yml`).

For this initial introduction of linting to the project, [`golint`](https://github.com/golang/lint#purpose) and [`go-header`](https://github.com/denis-tingaikin/go-header#go-header) are being added. `golint` helps to address styling issues associated with go projects. `go-header` checks if copyright headers are present for files in a project.

The introduction of `go-header` will help prevent checking in files with no license headers. All `golint` issues were fixed as part of introducing this change as well.

For now, linting is optional via a linter script under `/hack` but will eventually be added to CI. There is a blocker on adding it to CI that is being captured in a separate story.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,18 @@
+name: golangci-lint
+
+on:
+  push:
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.34
+          args: -v
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,17 +1,9 @@
-
-# all available settings of specific linters
-linters-settings:
-  goheader:
-    values:
-      const:
-        COMPANY: Copyright 2020 VMware, Inc.
-      # regexp:
-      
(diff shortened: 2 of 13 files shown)


joaopapereira (changes requested):
This looks good.
I would like the linter to run as part of our hack/build.sh or at least as part of our workflows. Can you do that change?

danielhelfand:
@joaopapereira From a CI standpoint, I think it would make sense to not run this as part of the test-gh action. It should be run in parallel and results should be reported separately than the action that builds/tests.

If we want to introduce it to the build script, it should then have some condition to not run in that test-gh action. We have a separate story for adding this to CI where we could maybe further explore what makes sense.

danielhelfand:
A potential compromise here would be to have a separate build-and-lint.sh script that calls build and also runs the linting process.

joaopapereira:
In my opinion, the user value from having the linter is to have it run automatically, so I would prefer to have the integration of the linter in our CI in this same PR.

danielhelfand:
Yeah, I think it's fair. Let's put this on hold then as we are waiting on a fix for go-header to be available in the next golangci-lint release. That is the hold up on introducing in CI.

danielhelfand:
The linter has been added to CI.

joaopapereira (approved):
LGTM

Issue #8, opened 2020-08-27
Add `--include-non-distributable` flag to imgpkg copy


ewrenn8:
Imgpkg should support a flag to allow the user to relocate images with foreign layers to airgapped envs.

Slack thread for some discussion: https://kubernetes.slack.com/archives/CH8KCCKA5/p1598013615011900

pivotaljohn:
This is a rather critical feature for our users on Windows. Pulling in for this upcoming release.

aaronshurley:
Consolidating https://github.com/vmware-tanzu/carvel-imgpkg/issues/71 into this issue.

## Story

**As a** operator wishing to collect a bundle that includes software that includes non-distributable bits (e.g. Windows-based products) for use in an air-gapped scenario
**I want to** be able to _override_ the default behavior of skipping transferring so-called non-distributable layers ("foreign layers" in the docker manifests, "non-distributable" in OCI manifests)
**and** I want the bits of those layers to be pulled and included in the tarball of the bundle
**and** I want those bits to be loaded into the registry I copy the tarballed bundle.

## Context
Original GitHub Issue: [imgpkg#8](https://github.com/vmware-tanzu/carvel-imgpkg/issues/8)

Requested by the .NET Build team (Malini Valliath, PM) multiple times since this summer.

PCS/MAPBU leadership has pointed to the importance of servicing our Windows-using customers — Ajay pointed out the importance of Windows support during an all-hands meeting, noting that a significant percentage of our customers use Windows.

## Acceptance Criteria

_Note: we say, "foreign layers", here. We wish for all this to be true for [any non-distributable layer](https://github.com/google/go-containerregistry/blob/master/pkg/v1/types/types.go#L47-L53)._

### 🟢 Copying foreign layers to a tarball

**Given** a bundle which includes a reference to an image containing a foreign layer (e.g. a bundle that references Microsoft's nanoserver — `mcr.microsoft.com/windows/nanoserver:1809-amd64`)
**When** I copy that bundle to a tarball
**And** I indicate that non-distributable layers (including foreign layers) should be included
```
$ imgpkg copy -b index.docker.io/k8slt/bundle-with-foreign-layers --to-tar the-football.tar --include-non-distributable
```
**Then** the tarball includes the contents of the layer

### 🟢 Copying a tarball containing one or more foreign layers to a registry

**Given** a tarball containing a bundle which includes one or more foreign layers
**And** a registry enabled to receive foreign layers (⁉️)
[...]

PR #62, opened 2021-01-28
Add `--include-non-distributable` flag to copy command

DennisDenuto:
### Why
To enable an operator wishing to collect a bundle that includes software that includes non-distributable bits (e.g. Windows-based products) for use in an air-gapped scenario

Implements https://github.com/vmware-tanzu/carvel-imgpkg/issues/8


gcheadle-vmware (approved):
Looks great 👍  Let's be sure to squash some of these commits before the merge.

PR #89, opened 2021-02-19
Bump go compile version to 1.16

DennisDenuto:
See https://golang.org/ref/mod#go-mod-file-go for more details

- It would be useful to use go embed for test assets
- start getting ahead of deprecation changes to ioutil
 


joaopapereira (approved):
LGTM

PR #90, opened 2021-02-22
Add e2e test for airgap workflow

DennisDenuto:
Backfill test to 'simulate' an airgapped workflow.

It adds assertions that the 'source' registry is not used during a copy to a different 'target' registry when a tarball is used as an intermediate.

### DIFF
--- a/test/e2e/copy_from_image_test.go
+++ b/test/e2e/copy_from_image_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


test/e2e/imgpkg.go
```diff
@@ -20,13 +20,14 @@ type Imgpkg struct {
...
 
 type RunOpts struct {
-	AllowError   bool
-	StderrWriter io.Writer
-	StdoutWriter io.Writer
-	StdinReader  io.Reader
-	CancelCh     chan struct{}
-	Redact       bool
-	EnvVars      []string
+	ContainerImage string
```
cppforlife: seems to be unused.

test/e2e/copy_from_image_test.go
```diff
@@ -261,3 +335,14 @@ func TestCopyErrorsWhenCopyToTarAndGenerateOutputLockFile(t *testing.T) {
 		t.Fatalf("expected copy to fail when --lock-output is provided with a tar destination, got %v", err)
 	}
 }
+
+func stopNetworkAccessToSourceRegistry(t *testing.T, env *Env) {
```
cppforlife: are we assuming here that $IMGPKG_E2E_AIRGAPPED_TEST_REPO is somehow related to `docker stop registry-for-airgapped-testing`? it seems that if we going to take on docker dependency that it should be container to the test instead of splitting it over test-e2e.sh and go test.

test/e2e/copy_from_image_test.go
```diff
@@ -126,31 +126,63 @@ func TestCopyImageInputToTarWithNonDistributableLayersFlagButContainsANonDistrib
 }
 
 func TestCopyAnImageFromATarToARepoThatDoesNotContainNonDistributableLayersButTheFlagWasIncluded(t *testing.T) {
-	env := BuildEnv(t)
-	imgpkg := Imgpkg{t, Logger{}, env.ImgpkgPath}
-	defer env.Cleanup()
+	t.Run("environment with internet", func(t *testing.T) {
```
cppforlife: im not sure i understand the distinction of having two tests where one uses stopNetworkAccessToSourceRegistry and the other one does not. what are we trying to verify?

test/e2e/copy_from_image_test.go
```diff
@@ -126,31 +126,63 @@ func TestCopyImageInputToTarWithNonDistributableLayersFlagButContainsANonDistrib
...
-	imgpkg.Run([]string{"copy", "-i", env.RelocationRepo, "--to-tar", tarFilePath})
+		// copy to tar skipping NDL
+		imgpkg.Run([]string{"copy", "-i", env.RelocationRepo, "--to-tar", tarFilePath})
 
-	imgpkg.RunWithOpts([]string{"copy", "--tar", tarFilePath, "--to-repo", repoToCopyName, "--include-non-distributable"}, RunOpts{
-		AllowError:   true,
-		StderrWriter: &stdOutWriter,
-		StdoutWriter: &stdOutWriter,
+		imgpkg.RunWithOpts([]string{"copy", "--tar", tarFilePath, "--to-repo", repoToCopyName, "--include-non-distributable"}, RunOpts{
+			AllowError:   true,
```
cppforlife: is Allowing Error here intentional? if so what error are we expecting? are there other assertions besides hint message we should be doing?

test/e2e/copy_from_image_test.go
```diff
@@ -186,43 +218,85 @@ func TestCopyAnImageFromARepoToATarThatDoesNotContainNonDistributableLayersButTh
...
+
+		// general setup
+		testDir := env.Assets.CreateTempFolder("image-to-tar")
+		tarFilePath := filepath.Join(testDir, "image.tar")
+
+		nonDistributableLayerDigest := env.ImageFactory.PushImageWithANonDistributableLayer(env.RepoForAirgappedTesting)
+		repoToCopyName := env.RelocationRepo + "include-non-distributable"
+
+		// copy to tar
+		imgpkg.Run([]string{"copy", "-i", env.RepoForAirgappedTesting, "--to-tar", tarFilePath, "--include-non-distributable"})
```
cppforlife: what behaviour are we trying to test by using RepoForAirgappedTesting vs RelocationRepo?

DennisDenuto: We want to have confidence that when copying from a tarball (that was copied from registry A) to a registry B, that registry A is not reached at all.

The above is something that we caught manually while working on a performance improvement story.

Having 2 separate tests for 'internet' and 'without internet' is useful for the `performance` story the team is working on. 

Which is that when copying a tarball (from registry A) to registry B,  and there *is* internet available, *and* the registries are the same, the layers will be mounted. (and if the `--include-non-distributable-layer` flag is provided,, but those layers are missing in the tarball, those layers are also mounted too! <-- we got this 'feature' for free).

The assertions will verify that when there is internet that the non distributable layers are mounted (even though the tarball didn't contain those layers)

When there isn't any internet we will verify that the layers were not copied and that the appropriate error message is emitted.

test/e2e/copy_from_image_test.go
DennisDenuto: It is intentional, the error we are expecting is on the next line which is that we expect a message explaining that the non-distributable layers could not be copied from the tarball since they were missing. There is a hint that the initial copy was missing the `--include-non-distributable` flag (hence why the non-distributable layers are missing).

test/e2e/copy_from_image_test.go
DennisDenuto: Ok. I'll make this change.

test/e2e/imgpkg.go
DennisDenuto: good catch. i'll remove this.

Issue #63, opened 2021-01-28
Allow writing a non-distributable layer upstream


gcheadle-vmware:
# Context

As part of #62 a 'workaround' solution was found to [ggcr's inability to write a non-distributable layer](https://github.com/google/go-containerregistry/blob/72597da50ce903895087828b5fc4b918636366f4/pkg/v1/remote/write.go#L76-L78).

The workaround solution was to write the 'foreign' layers upfront before trying to write the image manifest. 

We decided to proceed with the workaround solution, since there was some urgency behind getting window images working. See https://kubernetes.slack.com/archives/CH8KCCKA5/p1598013615011900

# Description
Create a PR or Issue to https://github.com/google/go-containerregistry to allow writing a non-distributable layer. 
Change imgpkg to use the above change to writing an image manifest that contains a 'foreign' image layer.

# Notes
- There's a [closed issue](https://github.com/google/go-containerregistry/issues/741) that has _some_ background.
- Matt Moore (mattmoor@vmware.com) worked on this library if it's difficult getting traction in the repo

---

## Tasks
- [x] Introduce Option to allow specifying "allow non distributable layers"
- [x] Wherever that option is used, support the allowance 
- - [x] Write (this is all we really need, but the options are also used...)
- - [x] MultiWrite
- [x] Add task to use upstream GGCR functionality _OR_ create a story to do so.

StevenLocke:
[Made the PR](https://github.com/google/go-containerregistry/pull/930) in upstream GGCR. Currently blocked on signing the CLA, unsure if that should be done as an individual or for the company.

After that gets merged, we should use it in imgpkg!

aaronshurley:
Looks like the PR was merged a couple of weeks back but has not been part of a release yet. Let's keep this issue open until we can consume the upstream changes.

DennisDenuto:
go-container registry released 0.4.1 that contains the PR to non-distributable layers

https://github.com/google/go-containerregistry/releases/tag/v0.4.1

aaronshurley:
This issue was resolved in https://github.com/vmware-tanzu/carvel-imgpkg/pull/110 and released in v0.6.0.

PR #95, opened 2021-02-24
Improve imgpkg copy performance

DennisDenuto:
- Uses blob mounting if copying an image between the same registry
- Uses [MultiWrite](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/multi_write.go)
-- Has current limitations of only allowing copying to the same repository which is true today for the copy command
-- Has current limitations of Images that contain stream.Layers will not be deduped (the performance improvement will not apply to these images)
-- Writing `non-distributable` layers are currently being uploaded serially. The idea here is that once [PR](https://github.com/google/go-containerregistry/pull/930) is released, we can leverage the new `non-distributable` option and let MultiWrite handle writing every layer (including non-distributable layers) in parallel.  The current implementation in this PR should result in slower copying if lots of images having mostly layers being non-distributable and not mountable. If this is a concern, this PR can be updated to also ensure writing non-distributable layers are done in parallel (knowing that this code will be removed once the google-container registry PR is released) 
- Added a perf test that runs a couple of benchmarks. imgpkg (w/out performance improvements) consistently fails the perf test (the initial push and copy take the same time). imgpkg (with performance improvement) consistently pass the perf test (the initial push takes around 30 seconds while the copy now takes 0.5 seconds)

Will leave in draft mode as some manual testing and a sanity check of the commits still needs to be conducted.
- [x] sanity check changes with Joaoa
- [x] test to see what happens if copying from a tarball that references layers in a private (not readable) repo (it should continue to work) <- it does continue to work.
- [x] test to make sure that copying a bundle from cf-for-k8s has at least the same or better performance characteristic. <- copying saw a 20% improvement


pkg/imgpkg/cmd/fake_registry_test.go
```diff
@@ -33,11 +34,11 @@ func NewFakeRegistry(t *testing.T) *FakeRegistry {
 
 func (r *FakeRegistry) Build() *imagesetfakes.FakeImagesReaderWriter {
 	fakeRegistry := &imagesetfakes.FakeImagesReaderWriter{}
-	fakeRegistry.GenericCalls(func(reference name.Reference) (descriptor v1.Descriptor, err error) {
+	getDescriptor := func(reference name.Reference, r *FakeRegistry) (v1.Descriptor, error) {
```
joaopapereira, line 37: Since this is used in multiple places would it make sense to be a FakeRegistry method?

pkg/imgpkg/image/registry.go
```diff
@@ -94,6 +111,48 @@ func (i Registry) Image(ref regname.Reference) (regv1.Image, error) {
 	return regremote.Image(overriddenRef, i.opts...)
 }
 
+func (i Registry) MultiWrite(imageOrIndexesToUpload map[regname.Reference]regremote.Taggable, concurrency int) error {
+	for ref, img := range imageOrIndexesToUpload {
```
joaopapereira, line 115: as we talked this have the serial upload of the non-distributable layers but will be removed as soon as go container registry releases a new version

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
+	case item.Index != nil:
+		lock.Lock()
+		defer lock.Unlock()
+		imageOrIndexesToWrite[uploadTagRef] = *item.Index
+	}
+
+	errCh <- nil
+}
+
+func verifyProcessedImages(item imagedesc.ImageOrIndex, errCh chan error, o *ImageSet, importRepo regname.Repository, registry ImagesReaderWriter, importedImages *ProcessedImages) {
```
joaopapereira: instead of passing `o` can we make `verifyProcessedImages` part of the struct?

test/e2e/copy_from_image_test.go
```diff
@@ -195,89 +236,47 @@ func TestCopyAnImageFromARepoToATarThatDoesNotContainNonDistributableLayersButTh
...
 func TestCopyToTarWithoutNonDistributableLayersFlagButContainsANonDistributableLayerShouldPrintAWarningMessage(t *testing.T) {
-	t.Run("environment with internet", func(t *testing.T) {
-		env := BuildEnv(t)
-		imgpkg := Imgpkg{t, Logger{}, env.ImgpkgPath}
-		defer env.Cleanup()
-
-		// general setup
-		testDir := env.Assets.CreateTempFolder("image-to-tar")
-		tarFilePath := filepath.Join(testDir, "image.tar")
+	env := BuildEnv(t)
```
joaopapereira: Instead of this maybe we can shorten the name of the func and have this `t.Run` with more description

test/perf/e2e.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package perf
```
joaopapereira: Is the idea here to isolate the perf package from the e2e?
Personally, I would prefer to have these common structs at the test package level and be consumed by both sets of tests.

test/perf/perf_test.go
```diff
@@ -0,0 +1,71 @@
...
+		env.ImageFactory.PushImage(perfTestingRepo, int64(GB))
+	})
+
+	benchmarkResultCopyInSameRegistry := testing.Benchmark(func(b *testing.B) {
+		imgpkg := Imgpkg{b, Logger{}, env.ImgpkgPath}
+
+		imgpkg.Run([]string{"copy", "-i", perfTestingRepo, "--to-repo", perfTestingRepo + strconv.Itoa(b.N)})
+	})
+
+	fmt.Println(fmt.Sprintf("initial push took: %v", benchmarkResultInitialPush.T))
```
joaopapereira: We should use our logger struct

joaopapereira (commented):
I like the general approach and think that the mountable images are going to be helpful timewise.
As we talked about yesterday the only thing that feels a bit awkward is for the imageset to decide if we should use a mountable image or not. Would it make sense for the registry to be the one to decide if we should mount or not an image? Or is it too late?

Added some comments on the PR

DennisDenuto:
Manual testing saw the following change to UX:

Because every image is uploaded via MultiWrite, the output no longer shows each image printed during the upload of the image. (i.e. "copy | importing index.docker.io/cloudfoundry/eirini-event-reporter-cf-for-k8s@sha256:1798ea550cfb7b1457698dc9a9766ad5333da05c1e0cd29e748b7db34e648e1b -> index.docker.io/k8slt/cf-for-k8s-from-tar-2@sha256:1798ea550cfb7b1457698dc9a9766ad5333da05c1e0cd29e748b7db34e648e1b...")

Instead the 'copy image' output is shown *after* MultiWrite has finished uploading the images, and is shown during the verify and tagging portion of the copy.

So... the contents of the output has stayed the same. However, during a copy of cf for k8s, it will print out
```
"copy | importing 45 images..."

< Then MultiWrite will upload all 45 images, which will take minutes >

"copy | importing index.docker.io/cloudfoundry/eirini-event-reporter-cf-for-k8s@sha256:1798ea550cfb7b1457698dc9a9766ad5333da05c1e0cd29e748b7db34e648e1b -> index.docker.io/k8slt/cf-for-k8s-from-tar-2@sha256:1798ea550cfb7b1457698dc9a9766ad5333da05c1e0cd29e748b7db34e648e1b..."

<the above will be printed as each image is verified and tagged>

"copy | imported 45 images"
"Succeeded"
```

I think having MultiWrite provide a way to get a 'status' channel on which images have been uploaded would be a useful PR upstream See issue https://github.com/google/go-containerregistry/issues/952. That channel could be used to restore our UX to how it used to behave.

Perhaps changing the log message from "copy | importing 45 images..." to "copy | importing 45 images, this may take some time..." is sufficient for now?

Thoughts? Concerns?

test/perf/perf_test.go
DennisDenuto: done

test/e2e/copy_from_image_test.go
DennisDenuto: This is a good call out. I collapsed 3 "long named tests" into a test called:
"TestCopyRepoToTarAndThenCopyFromTarToRepo"
and then those 3 tests from before are nested t.Runs.

pkg/imgpkg/imageset/image_set.go
DennisDenuto: Done

pkg/imgpkg/image/registry.go
DennisDenuto: I pinky promise this will go away as part of https://github.com/vmware-tanzu/carvel-imgpkg/issues/63

pkg/imgpkg/image/resolved.go
```diff
@@ -31,27 +29,15 @@ func (i ResolvedImage) URL() (string, error) {
...
-	if err != nil {
-		return "", err
-	}
-
-	if imgDescriptor.Digest.String() != imgDescriptor2.Digest.String() {
-		return "", fmt.Errorf("Expected digest resolution to be consistent over two separate requests")
-	}
-
-	digestURL, err := regname.NewDigest(tag.Repository.String() + "@" + imgDescriptor.Digest.String())
+	digest, err := regname.NewDigest(tag.Repository.String() + "@" + hash.String())
```
cppforlife: it seems like ResolvedImage is only used in one place. can we inline this simplified logic there?

pkg/imgpkg/image/registry.go
```diff
@@ -85,6 +85,23 @@ func (i Registry) Generic(ref regname.Reference) (regv1.Descriptor, error) {
...
+func (i Registry) Get(ref regname.Reference) (*regremote.Descriptor, error) {
+	return regremote.Get(ref, i.opts...)
+}
+
+func (i Registry) Digest(ref regname.Reference) (regv1.Hash, error) {
+	overriddenRef, err := regname.ParseReference(ref.String(), i.refOpts...)
+	if err != nil {
+		return regv1.Hash{}, err
+	}
+	desc, err := regremote.Head(overriddenRef, i.opts...)
```
cppforlife, line 97: should we return digest directly from the reference if reference is regv1.Digest?

pkg/imgpkg/imagedesc/types.go
```diff
@@ -133,3 +140,23 @@ func (t ImageOrIndex) Tag() string {
 		panic("Unknown item")
 	}
 }
+
+func (t ImageOrIndex) MountableImage(registry RegistryRemoteImage) (regv1.Image, error) {
```
cppforlife: i dont believe this pkg interacts with anything that talks to remote registries. let's move this method into ImageSet.

pkg/imgpkg/imagedesc/types.go
```diff
@@ -133,3 +140,23 @@ func (t ImageOrIndex) Tag() string {
...
+	if t.Image == nil {
+		return nil, fmt.Errorf("Unable to retrieve a mountable image on an imageindex")
+	}
+
+	reference, err := regname.ParseReference((*t.Image).Ref())
+	if err != nil {
+		return nil, err
+	}
+	descriptor, err := registry.Get(reference)
```
cppforlife: if we are going to grab images from a remote source, we need to have a gurantee that reference points to an immutable image (digest). otherwise it could be a tag and this could return a random/mutable image.

pkg/imgpkg/imagedesc/types.go
```diff
@@ -133,3 +140,23 @@ func (t ImageOrIndex) Tag() string {
...
+	if err != nil {
+		return nil, fmt.Errorf("Getting mountable image failed: %s: %s", reference, err)
```
cppforlife: out of curiosity, how do we know it's a mountable image? Get seems to be returning plain regv1.Image

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
 
-	return importedImages, nil
+func addImageOrImageIndexToTaggableMapForMultiWrite(item imagedesc.ImageOrIndex, errCh chan error, importRepo regname.Repository, registry ImagesReaderWriter, imageOrIndexesToWrite map[regname.Reference]regremote.Taggable, lock *sync.Mutex) {
+	itemDigest, err := item.Digest()
+	if err != nil {
+		errCh <- err
+		return
+	}
+
+	tag := fmt.Sprintf("imgpkg-%s-%s", itemDigest.Algorithm, itemDigest.Hex)
```
cppforlife: we already build this tag in one place. how do we avoid building it up in multiple places (not suggesting that we pull this out into a method).

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
-				DigestRef:           importDigestRef.Name(),
-				Image:               regImage,
-				ImageIndex:          regImageIndex,
-			})
-			errCh <- nil
+	errChVerifyImages := make(chan error, len(imgOrIndexes))
+	for _, item := range imgOrIndexes {
+		item := item // copy
+
+		go func() {
```
cppforlife, line 119: instead of propagating errChVerifyImages to function, let's have verifyProcessedImages return an error and use channel as close to go func() instantiation as possible. this way we are not leaking async nature of what we are doing into verifyProcessedImages and its children.

pkg/imgpkg/imagedesc/types.go
joaopapereira: This is interesting because this was not here in the first version of the PR and when I talked with @DennisDenuto we moved it here.
This package does talk with the registry to retrieve the descriptors for the images.

Current ImageSet knows how to Export and Import images but Mountable Images feel like more an implementation detail that ImageSet shouldn't care about because it just provides the images to go-container registry for upload. My initial thought was that the image should be able to determine if it is mountable or not. Unfortunately we can only mount images that are in the same registry and if you call `imgpkg` in a air gapped environment go-containerregistry will try to mount the image from a place that might not be accessible.

Another option that eventually we could think about is to push this concern to the Registry itself. Doing that could also be interesting but I am not 100% sure if we would have all the needed information when we are performing the writing. (Do not remember the code right now but I see no reason why we would not have all the information we need at that level)

pkg/imgpkg/imageset/image_set.go
cppforlife: same applies to verifyProcessedImages() taking errCh

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
+		UnprocessedImageRef: UnprocessedImageRef{existingRef.Name(), item.Tag()},
+		DigestRef:           importDigestRef.Name(),
+		Image:               regImage,
+		ImageIndex:          regImageIndex,
+	})
+	errCh <- nil
 }
 
-func (o *ImageSet) importImage(item imagedesc.ImageOrIndex,
+func (o *ImageSet) tagAndVerifyImage(item imagedesc.ImageOrIndex,
```
cppforlife: since MultiWrite is doing all the work, we no longer have "real time" logging of image imports (o.logger.Write...). this is going to be weird for the user as im guessing it looks like there is a big gap of "silence" in output while things get uploaded. and then we print importing... more or less at once.

cppforlife: i think we should split this function into two: one that verifies things and one that adds additional tag (at glance dont see anything being reused between two parts).

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
+		lock.Lock()
+		defer lock.Unlock()
+		imageOrIndexesToWrite[uploadTagRef] = *item.Index
+	}
+
+	errCh <- nil
+}
+
+func (o *ImageSet) verifyProcessedImages(item imagedesc.ImageOrIndex, errCh chan error, importRepo regname.Repository, registry ImagesReaderWriter, importedImages *ProcessedImages) {
+	existingRef, err := regname.NewDigest(item.Ref())
```
cppforlife, line 205: it seems that this line and tagAndVerifyImage is doing some repetitive work in getting the digest.

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
+	if err != nil {
+		return nil, err
+	}
 
-			importDigestRef, err := o.importImage(item, existingRef, importRepo, registry)
-			if err != nil {
-				errCh <- fmt.Errorf("Importing image %s: %s", existingRef.Name(), err)
-				return
-			}
+	err = registry.MultiWrite(imageOrIndexesToWrite, o.concurrency)
```
cppforlife, line 110: it seems that MultiWrite is only going to work against same repo [0] (is that going to screw us up if we decide to do copy with multiple repo destinations based on where user wants us to write images to @joaopapereira)

[0] https://github.com/google/go-containerregistry/blob/4d068fbcb51f224f9667497fe24b83506f979ede/pkg/v1/remote/multi_write.go#L34

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
+
+	tag := fmt.Sprintf("imgpkg-%s-%s", itemDigest.Algorithm, itemDigest.Hex)
+	uploadTagRef, err := regname.NewTag(fmt.Sprintf("%s:%s", importRepo.Name(), tag))
+	if err != nil {
+		errCh <- fmt.Errorf("Building upload tag image ref: %s", err)
+		return
+	}
+
+	switch {
+	case item.Image != nil:
```
cppforlife, line 155: we should panic in default switch case if it's not image or index.

pkg/imgpkg/imageset/image_set.go
```diff
@@ -77,57 +79,135 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
...
+		imageToWrite := regv1.Image(*item.Image)
+
+		if imageBlobsCanBeMounted(itemRef, uploadTagRef) {
+			imageToWrite, err = item.MountableImage(registry)
+			if err != nil {
+				imageToWrite = regv1.Image(*item.Image)
+			}
+		}
+
+		lock.Lock()
```
cppforlife: i think it would be better for this function to return uploadTagRef and taggable image/index, and have imageOrIndexesToWrite and associated lock be used in the same scope as it's defined. (this way you dont have to pass them around) to mutate.

cppforlife: (goes hand in hand with not passing down channel and dealing with them in the same scope as go routine).

pkg/imgpkg/imagedesc/types.go
cppforlife: also this function is only applicable to images so if it takes ImageWithRef we wouldnt need first conditional. i also think that we can use imageBlobsCanBeMounted inside here so that question of mountability is encasulated in this function.

test/e2e/copy_from_image_test.go
```diff
@@ -210,32 +174,35 @@ func TestCopyToTarWithoutNonDistributableLayersFlagButContainsANonDistributableL
...
 
 		_, err = layer.Compressed()
-		if err == nil {
-			t.Fatalf("Expected non-distributable layer to NOT be copied into registry, however it was")
+		if err != nil {
+			t.Fatalf("Expected to find a non-distributable layer however it wasn't found. Got response code: %v", err)
+		}
+
+		imageDigest := fmt.Sprintf("@%s", extractDigest(t, stderr.String()))
+		cmd := exec.Command("docker", "pull", repoToCopyName+imageDigest)
```
cppforlife: ideally we wouldnt have any dependence on docker. we can use imgpkg pull -i here.

test/e2e/copy_from_image_test.go
```diff
@@ -278,6 +245,40 @@ func TestCopyToTarWithoutNonDistributableLayersFlagButContainsANonDistributableL
 			t.Fatalf("Expected non-distributable layer to NOT be copied into registry, however it was")
 		}
 	})
+
+	t.Run("With --lock-output flag should generate a valid ImageLock file", func(t *testing.T) {
```
cppforlife, line 247: is this related to perf work?

test/perf/imgpkg.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package perf
```
cppforlife: seems like we are using same type of functionality. can we find a different method to section off the tests? (if that's what the intent was to do with a separate package).

test/perf/perf_test.go
```diff
@@ -0,0 +1,72 @@
...
+type ByteSize int64
+
+const (
+	_           = iota
+	KB ByteSize = 1 << (10 * iota)
+	MB
+	GB
+)
+
+func TestBenchmarkCopyingLargeImageWithinSameRegistryShouldBeFast(t *testing.T) {
```
cppforlife: can we add cf-for-k8s test from kbld as a "pressure" test for large number of images? (https://github.com/vmware-tanzu/carvel-kbld/blob/develop/test/e2e/packaging_test.go#L160)

pkg/imgpkg/imagedesc/types.go
cppforlife: >  Unfortunately we can only mount images that are in the same registry

that's what conceptually it means to be mountable so we model based on that. only place that has source and destination is ImageSet class we got going on. this function in current form does nothing related to mounting with the exception of being named Mountable. it could be named FromRegistry for example imho and it would reflect more accurately what it's doing. but does it really make sense to have a method that fetches image from a registry on an image itself (ImageWithRef is a regv1.Image)? imho not, hence proposal to move it to be on ImageSet as a prive method.

test/e2e/copy_from_image_test.go
DennisDenuto: I can make this change, but i'm confused to why it is ideal.

pkg/imgpkg/imageset/image_set.go
DennisDenuto: I can make this change, although it's not obvious to me why it is 'better'

pkg/imgpkg/imageset/image_set.go
DennisDenuto: I can make this change, although it's not obvious to me why. It sounds like this function is doing too much? funnily enough it's now doing half of what it used to do :-P 

I can halve this function's responsibility one more time.

DennisDenuto: Yeah this is true. This is what my second comment raised. I outline a couple of ideas on how to handle this (assuming we want to).

pkg/imgpkg/imageset/image_set.go
DennisDenuto: I'll go ahead and make this change. Although I don't fully understand the concern here.

pkg/imgpkg/imageset/image_set.go
DennisDenuto: Hmm i'm not sure! but if someone accidentally adds (or deletes) a character to that tag string, a few tests fail.

pkg/imgpkg/imagedesc/types.go
DennisDenuto: Get() returns a `regremote.Descriptor` which is metadata about an image. 
From the docs about remote Descriptors:
```
Descriptor provides access to metadata about remote artifact and accessors for efficiently converting it into a v1.Image or v1.ImageIndex.
```

We later use this remote Descriptor to get a v1.Image (which under the hood is a mountable image). 

Feels like you hinting that the word "mountable image" should be named something else?

pkg/imgpkg/image/registry.go
DennisDenuto: This questions seems to want to answer itself. Which is "yes".

But.... I think.... while it might have been a correct digest at some point of time, that artifact might have been deleted from the registry. In which case we would want to make the remote call to "Head" with a reference (which might be a digest) and see it fail with a `404`

test/e2e/copy_from_image_test.go
DennisDenuto: No, that was just a test being re-organized. running it in a t.Run allowed it to belong to a parent context and for the test name to be a sentence.

test/perf/perf_test.go
DennisDenuto: I did the manual test for this. Yeah i'm happy to add it to the perf test. What do you suggest this pressure test asserts?

pkg/imgpkg/imageset/image_set.go
DennisDenuto: I thought the saying went to stay calm and not panic? 

I'll add the panic to the case.

test/perf/e2e.go
DennisDenuto: Ok, i'll see if I can move `imgpkg` and the `logger` somewhere more re-usable between the two kind of test suites

pkg/imgpkg/image/resolved.go
DennisDenuto: Ok, i do enjoy deleting code.

test/perf/perf_test.go
DennisDenuto: On a call with @cppforlife and he suggested adding a time assertion. make sure copying a large cf-for-k8s bundle takes less than X time.

test/perf/imgpkg.go
DennisDenuto: Spoke to @cppforlife and sharing the imgpkg run and logger into a shared package between e2e and perf.

Make this a separate commit.

pkg/imgpkg/imageset/image_set.go
DennisDenuto: @cppforlife and I had a convo, we think that in the future, if setting custom destination renaming of images, we are not blocked technically by this change. We could for e.g. call MultiWrite, multiple times.

pkg/imgpkg/imageset/image_set.go
DennisDenuto: @cppforlife talked, we don't think this PR should be blocked by this concern. It is a valid concern. We also don't think changing the log message "copying.... this may take some time" is sufficient. 

- we think that this should be solved by contributing upstream to https://github.com/google/go-containerregistry/issues/952
-- This will change the UX to a 'progress bar' view. "We think this might be OK, for users" DK

pkg/imgpkg/imagedesc/types.go
DennisDenuto: @cppforlife and I moved MountableImage to ImageSet. A strong reason being that there isn't a concept of an ImageIndex being mountable.

pkg/imgpkg/imagedesc/types.go
DennisDenuto: @cppforlife done

pkg/imgpkg/image/registry.go
DennisDenuto: @cppforlife there is a question as to whether a separate function is introduced (to check if an image has been removed), or additional comments around the Digest needs to be made.

pkg/imgpkg/imageset/image_set.go
DennisDenuto: @cppforlife is going to re-read the code and provide an updated feedback on this code.

test/perf/imgpkg.go
DennisDenuto: I decided to move every utility file (even those that are not shared across the 2 packages) into a 'helpers' package.

It results in only test files living under e2e / perf, which I think is cleaner because it helps to see what is being tested at a glance.

test/perf/e2e.go
DennisDenuto: I decided to move every utility file (even those that are not shared across the 2 packages) into a 'helpers' package.

It results in only test files living under e2e / perf, which I think is cleaner because it helps to see what is being tested at a glance.

pkg/imgpkg/cmd/fake_registry_test.go
DennisDenuto: If i'm understanding this correctly, you are asking if the `getDescriptor` function should be a method on `FakeRegistry`. 

I think the short answer is 'no' since `fakeRegistry` is a counterfeiter object of the `ImagesReaderWriter` interface. (It is generated code, and to change it would involve changing the registry interface)

I'm going to resolve this conversation. If this is an incorrect interpretation, let me know and we can improve this part of the code together.

PR #97, opened 2021-02-26
Maintain annotations when copying from an ImageLock file

### LINKED ISSUE #77: imgpkg copy from lock file does not copy annotations
**What steps did you take:**

```
kbld -f examples/simple-app-build-and-push/ --imgpkg-lock-output foo.yml
imgpkg copy --lock foo.yml --lock-output foo2.yml --to-repo localhost:5000/foo
```

```foo.yml
---
apiVersion: imgpkg.carvel.dev/v1alpha1
images:
- annotations:
    kbld.carvel.dev/id: simple-app
  image: localhost:5000/cppforlife/kbld-simple-app2@sha256:09f45eea0d452749445d1d015745b2bc5cfcbfb49f952d482d14f37ae7680a38
kind: ImagesLock
```

```foo2.yml
---
apiVersion: imgpkg.carvel.dev/v1alpha1
images:
- image: localhost:5000/foo@sha256:09f45eea0d452749445d1d015745b2bc5cfcbfb49f952d482d14f37ae7680a38
kind: ImagesLock
```

**What happened:**
The `annotations` were dropped when I copied the image

**What did you expect:**
They'd be maintained, so that the mappings of app -> image would be maintained

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.3.0

joaopapereira:
This PR has 2 commits
1 - Maintain the annotations on the ImageLock file that is output by copy, when the copy is done from an ImageLock file
Fixes #77 
2 - Add testify to the code base
start using testify on the tests that were created/modified by the prior fix
rearrange some tests to start using t.Run and logger.Section for clearer test structure


cppforlife:
can we localize logic of retaining annotations within ImageLock class? e.g. new ImageLock object takes as input previous ImageLock and picks up annotations.

joaopapereira:
> can we localize logic of retaining annotations within ImageLock class? e.g. new ImageLock object takes as input previous ImageLock and picks up annotations.

The problem with this is that we do not have access to an ImageLock file. The read of this file is done inside `CopyRepoSrc` and it feels like returning it would be strange since the majority of the cases we will not have an ImageLock file.

I see a possibility here where the `CopyToRepo` could return an ImageLock and BundleLock struct, that way in the `Copy.Run` we would just write it to disk. The caveat I see here is that we are pushing the logic about which type of files we want to create into the `CopyToRepo` struct and that feels a little bit like the UX struct responsibility. 

Not sure what do you think @cppforlife ?

cppforlife:
hmm, what about retaining annotations on UnprocessedImage instead of trying to keep them separate?

joaopapereira:
I tried that but the problem is that we use the `UnprocessedImage` as the key for `ProcessedImages` which causes other types of problems.
- It has to be a pointer or else it will cause a compilation error
- We would have to change the way we are indexing because `ProcessedImage` by default will not have the annotations because that information does not get passed around

So after fighting this for a bit I ended up deciding to separate the annotation, not a fan but maybe in the future we can see if there is a better way to consolidate everything as we evolve the Recursive bundle track since we will have to do some changes in near areas

pkg/imgpkg/cmd/copy.go
```diff
@@ -196,10 +196,25 @@ func (o *CopyOptions) writeImagesLockOutput(processedImages *ctlimgset.Processed
...
-		})
+	if o.LockInputFlags.LockFilePath != "" {
+		var err error
+		imagesLock, err = lockconfig.NewImagesLockFromPath(o.LockInputFlags.LockFilePath)
+		if err != nil {
+			return err
+		}
+		for i, image := range imagesLock.Images {
+			for _, img := range processedImages.All() {
+				if img.UnprocessedImageRef.DigestRef == image.Image {
```
cppforlife: probably worth pulling this out as a method on processed images so it's easier to error out when there is match found (case isnt covered in this code).

pkg/imgpkg/cmd/copy.go
```diff
@@ -196,10 +196,25 @@ func (o *CopyOptions) writeImagesLockOutput(processedImages *ctlimgset.Processed
...
+	if o.LockInputFlags.LockFilePath != "" {
+		var err error
+		imagesLock, err = lockconfig.NewImagesLockFromPath(o.LockInputFlags.LockFilePath)
+		if err != nil {
+			return err
+		}
+		for i, image := range imagesLock.Images {
+			img, found := processedImages.FindByURL(ctlimgset.UnprocessedImageRef{DigestRef: image.Image})
+			if !found {
+				return fmt.Errorf("Expected image %s to have been relocate but was not", image.Image)
```
cppforlife: nit: let's wrap value in quotes. e.g. `Expected image "%s" to...`
nit: relocate -> copied. we dont use relocate term anywhere.

pkg/imgpkg/cmd/copy.go
```diff
@@ -196,10 +196,25 @@ func (o *CopyOptions) writeImagesLockOutput(processedImages *ctlimgset.Processed
...
-			Image: img.DigestRef,
-		})
+	if o.LockInputFlags.LockFilePath != "" {
+		var err error
+		imagesLock, err = lockconfig.NewImagesLockFromPath(o.LockInputFlags.LockFilePath)
+		if err != nil {
+			return err
+		}
+		for i, image := range imagesLock.Images {
+			img, found := processedImages.FindByURL(ctlimgset.UnprocessedImageRef{DigestRef: image.Image})
```
cppforlife, line 206: is this going to not work in cases when UnprocessedImageRef included a Tag? do we need to not index on the tag, or does it matter?

joaopapereira: If you provide an ImagesLock file there is no place to add a tag. So I believe it is safe to ignore the tag in this particular situation.

PR #100, opened 2021-03-05
Pulling a Recursive bundle writes to disk

DennisDenuto:
This allows pulling a bundle that contains a bundle that could contain a bundle...

Things left to do:
- [x] Run some manual testing
- [x] Sanity check the commits to make sure there isn't any left over TODO's 
- [x] Squash some of those commits

Things to point out:
-  As discussed with Joao, "Found 1 Bundle packaged" isn't printed in the output (to simplify the recursive bundle pull logic)
- Pulling bundles is currently done serially. A performance improvement can be made to run in parallel. If we want that, it will be part of a different story.


pkg/imgpkg/bundle/bundle.go
```diff
@@ -19,48 +21,116 @@ const (
 )
 
 type Bundle struct {
-	plainImg     *plainimg.PlainImage
-	imgRetriever ctlimg.ImagesMetadata
+	plainImg      *plainimg.PlainImage
+	imgRetriever  ctlimg.ImagesMetadata
+	numSubBundles int
```
cppforlife: numSubBundles variable seems to be only used in pull. let's localize it. if we need to use in recursion, it could be passed around.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -19,48 +21,116 @@ const (
...
+	return &Bundle{plainImg, imagesMetadata, 0}
 }
 
 func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
 func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
-func (o *Bundle) Pull(outputPath string, ui ui.UI) error {
+func (o *Bundle) Pull(outputPath string, ui goui.UI, recursive bool) error {
+	return o.pull(recursive, outputPath, "", map[string]bool{}, ui)
```
cppforlife: order of vars for Pull and pull could be made consistent to make it less surprising when reading.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -19,48 +21,116 @@ const (
...
 	img, err := o.checkedImage()
 	if err != nil {
 		return err
 	}
 
-	ui.BeginLinef("Pulling bundle '%s'\n", o.DigestRef())
+	if o.rootBundle(bundlePath) {
+		ui.BeginLinef("Pulling bundle '%s'\n", o.DigestRef())
+	} else {
+		ui.BeginLinef("Pulling Nested bundle '%s'\n", o.DigestRef())
```
cppforlife: nested shouldnt be capitlized.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -19,48 +21,116 @@ const (
...
 	}
 
-	imagesLock, err := lockconfig.NewImagesLockFromPath(filepath.Join(outputPath, ImgpkgDir, ImagesLockFile))
+	imagesLock, err := lockconfig.NewImagesLockFromPath(filepath.Join(baseOutputPath, bundlePath, ImgpkgDir, ImagesLockFile))
 	if err != nil {
 		return err
 	}
 
-	err = NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).WriteToPath(outputPath, ui)
+	for _, image := range imagesLock.Images {
```
cppforlife: it seems like we are looping over the entire images list even if user did not specify recursive. i see that we also have a message `To retrieve the configuration of all the nested bundles use the --recursive flag`. this is quite expensive especially for bundles that may references 1000s of images. let's not do any recursive processing if flag is not specified and drop the message.

DennisDenuto: Not iterating over images list (to check if the image is a bundle) affects the output of the recursive bundle command. Specifically https://github.com/vmware-tanzu/carvel-imgpkg/blob/24f13393a485b185f8fadfc65e6180b7e67f5a8d/pkg/imgpkg/bundle/bundle.go#L101

@joaopapereira Thoughts on the above potential change to the output message?

Also, Can you expand on why you want to remove the message `To retrieve the configuration of all the nested bundles use the --recursive flag`. Is it to make the recursive flag less likely for folk to use it even if they don't need to? 
Another approach would be to capture your concern into the message.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: attention. to. detail. i'll update this.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: lol I was actually going for a plot twist in this story. I'll update this.

pkg/imgpkg/bundle/bundle.go
cppforlife: > Not iterating over images list (to check if the image is a bundle) affects the output of the recursive bundle command

so yeah, im suggesting we drop all that output so that we dont iterate over images in non-recursive case. it's a pretty  expensive operation.

joaopapereira: I thought a little bit about the implications of having 1000s of images in a bundle, but is this a use case we know it will exist? I think that in terms of UX it is better if we inform the user that this is a bundle that contains subbundles and in order to get all the configuration, they need to use the `--recursive` flag.

I think there are ways we can optimize the current code to be more resilient
1. We can upfront fetch all the images in a threaded way and afterward just iterate over them
2. If we do not have a `--recursive` we can print just the warning message and not the bundle name allows us to break sooner from this loop

pkg/imgpkg/bundle/bundle.go
```diff
@@ -19,48 +21,116 @@ const (
...
+
+		if o.shouldPrintNestedBundlesHeader(bundlePath, o.numSubBundles) {
+			ui.BeginLinef("\nNested bundles\n")
+		}
+		bundleDigest, err := name.NewDigest(image.Image)
+		if err != nil {
+			return err
+		}
+		if recursive {
+			err = subBundle.pull(recursive, baseOutputPath, filepath.Join(ImgpkgDir, SubBundlesDir, digestFriendlyDirectoryPath(bundleDigest)), imagesProcessed, goui.NewIndentingUI(ui))
```
joaopapereira: Maybe the function `digestFriendlyDirectoryPath` can return already the full path instead of returning only a folder name. Maybe call it `subBundlePath(bundleDigest string) string`

pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,365 @@
...
+		assert.NoError(t, err)
+		defer os.Remove(outputPath)
+
+		// test subject
+		err = subject.Pull(outputPath, fakeUI, recursiveBundles)
+		assert.NoError(t, err)
+		assert.DirExists(t, outputPath)
+
+		// assert icecream bundle was recursively pulled onto disk
+		digest, err := fakeRegistry.state["index.docker.io/icecream/bundle:latest"].image.Digest()
```
joaopapereira: Can we not use the return of the `WithBundleFromPath` instead of going inside the fake? Looks like we are spilling the registry implementation into this test

pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,365 @@
...
+		outputDirConfigFile := filepath.Join(outputPath, "config.yml")
+		assert.FileExists(t, outputDirConfigFile)
+		actualConfigFile, err := os.ReadFile(outputDirConfigFile)
+		assert.NoError(t, err)
+		expectedConfigFile, err := os.ReadFile("test_assets/bundle/config.yml")
+		assert.NoError(t, err)
+		assert.Equal(t, string(actualConfigFile), string(expectedConfigFile))
+	})
+
+	t.Run("bundle referencing another bundle that references another bundle", func(t *testing.T) {
```
joaopapereira: This test felt a little bit unexpected, because I did not see any indication that the pull was not recursive, so the assertions looked incorrect.
I would prefer to have that information in the test name or/and change the call to `Pull` where we explicitly add the value `false` to it.

The fact that this variable, that is quite important being set at the beginning of the file fills too far away from where it is used. But if we want to keep it I would consider also renaming it to `doNotPullNestedBundles` or something similar

pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,365 @@
...
+		assert.NoError(t, err)
+
+		outputDirConfigFile = filepath.Join(outputPath, ".imgpkg", "bundles", strings.ReplaceAll(digest.String(), "sha256:", "sha256-"))
+		assert.NoDirExists(t, outputDirConfigFile)
+	})
+}
+
+func TestPullBundlesRecursivelyWritingContentsToDisk(t *testing.T) {
+	fakeUI := &bundlefakes.FakeUI{}
+	recursiveBundles := true
```
joaopapereira:
```suggestion
	pullNestedBundles := true
```

pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,365 @@
...
+		// assert apples bundle was recursively pulled onto disk
+		digest, err = fakeRegistry.state["index.docker.io/apples/bundle:latest"].image.Digest()
+		assert.NoError(t, err)
+
+		outputDirConfigFile = filepath.Join(outputPath, ".imgpkg", "bundles", strings.ReplaceAll(digest.String(), "sha256:", "sha256-"))
+		assert.NoDirExists(t, outputDirConfigFile)
+	})
+}
+
+func TestPullBundlesRecursivelyWritingContentsToDisk(t *testing.T) {
```
joaopapereira:
```suggestion
func TestPullAllNestedBundlesWritingContentsToDisk(t *testing.T) {
```

pkg/imgpkg/bundle/repo/some-bundle-name/.imgpkg/bundle.yml
```diff
@@ -0,0 +1,9 @@
+apiVersion: imgpkg.carvel.dev/v1alpha1
```
joaopapereira: Can we move these files to a `test_assets` folder so that we know what we are using these for?

pkg/imgpkg/bundle/test_assets/bundle/.imgpkg/bundle.yml
```diff
@@ -0,0 +1,9 @@
+apiVersion: imgpkg.carvel.dev/v1alpha1
```
joaopapereira, line 1: where do we use this bundle at?

pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,365 @@
...
+		outputDirConfigFile := filepath.Join(outputPath, ".imgpkg", "bundles", strings.ReplaceAll(digest.String(), "sha256:", "sha256-"), "config.yml")
+		assert.FileExists(t, outputDirConfigFile)
+		actualConfigFile, err := os.ReadFile(outputDirConfigFile)
+		assert.NoError(t, err)
+		expectedConfigFile, err := os.ReadFile("test_assets/bundle_with_mult_images/config.yml")
+		assert.NoError(t, err)
+		assert.Equal(t, string(actualConfigFile), string(expectedConfigFile))
+	})
+
+	t.Run("bundle referencing another bundle that references another bundle", func(t *testing.T) {
```
joaopapereira: Is this test the same as the previous one but with 1 more level of nesting? If so I am ok with having only the 3 level bundle test.
Also, I think it would be nice to have a test with
```
bundle1
   bundle2
      bundle3
   bundle3
```

pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,365 @@
...
+		actualConfigFile, err = os.ReadFile(outputDirConfigFile)
+		assert.NoError(t, err)
+		expectedConfigFile, err = os.ReadFile("test_assets/bundle_with_mult_images/config.yml")
+		assert.NoError(t, err)
+		assert.Equal(t, string(actualConfigFile), string(expectedConfigFile))
+	})
+}
+
+func TestPullBundlesOutputToUser(t *testing.T) {
+	recursive := false
```
joaopapereira:
```suggestion
	doNotPullNestedBundles := false
```

pkg/imgpkg/bundle/bundle.go
```diff
@@ -19,48 +21,116 @@ const (
...
-	return &Bundle{plainImg, imagesMetadata}
+	return &Bundle{plainImg, imagesMetadata, 0}
 }
 
 func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
 func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
-func (o *Bundle) Pull(outputPath string, ui ui.UI) error {
+func (o *Bundle) Pull(outputPath string, ui goui.UI, recursive bool) error {
```
joaopapereira:
```suggestion
func (o *Bundle) Pull(outputPath string, ui goui.UI, pullNestedBundles bool) error {
```

pkg/imgpkg/bundle/bundle.go
```diff
@@ -19,48 +21,116 @@ const (
...
+	return o.pull(recursive, outputPath, "", map[string]bool{}, ui)
+}
+
+func (o *Bundle) pull(recursive bool, baseOutputPath string, bundlePath string, imagesProcessed map[string]bool, ui goui.UI) error {
```
joaopapereira:
```suggestion
func (o *Bundle) pull(baseOutputPath string, bundlePath string, pullNestedBundles bool, ui goui.UI, imagesProcessed map[string]bool) error {
```

pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -0,0 +1,365 @@
...
+
+To retrieve the configuration of all the nested bundles use the --recursive flag
+
+Locating image lock file images...
+One or more images not found in bundle repo; skipping lock file update`, output.String())
+	})
+}
+
+func TestPullBundlesRecursivelyOutputToUser(t *testing.T) {
+	recursive := true
```
joaopapereira:
```suggestion
	pullNestedBundles := true
```

pkg/imgpkg/bundle/fake_images_metadata_builder_test.go
```diff
@@ -0,0 +1,259 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package bundle_test
```
joaopapereira, line 4: On one hand, I like the fact that we have a smaller implementation of the fake registry, but on the other handle it feels like we are duplicating the logic of the registry and in the future, if we change the Metadata Reader interface we would have to duplicate the logic in this fake as well as in the other fake registry implementation. Not sure if we will a lot but doing this.

If we are afraid of having a very big fake implementation for the registry we can try to find a way to compose smaller fakes to do the work we need. If we also create tests for our fake registry if we are concerned about the size or correctness of the functionality

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -28,12 +28,14 @@ type ImagesLock struct {
 
 func (o *ImagesLock) WriteToPath(outputPath string, ui ui.UI) error {
 	imagesLockPath := filepath.Join(outputPath, ImgpkgDir, ImagesLockFile)
-	ui.BeginLinef("Locating image lock file images...\n")
 
 	imagesLock, skipped, err := o.LocalizeImagesLock()
 	if err != nil {
 		return err
 	}
+
```
joaopapereira, line 65: Nice catch

pkg/imgpkg/image/dir_image.go
```diff
@@ -42,13 +42,18 @@ func (i *DirImage) AsDirectory() error {
 		return err
 	}
 
+	//TODO: use constants
+	if !strings.Contains(i.dirPath, ".imgpkg/bundles") {
```
joaopapereira: We talked about the package `image` not being aware of the presence of bundles, so I think we should move this print into the caller of this function.

joaopapereira (changes requested):
The implementation looks good. I left some comments on possible ways to improve the retrieval of image information and also so other suggestions around naming and test structure

pkg/imgpkg/bundle/bundle.go
cppforlife: > but is this a use case we know it will exist?

this is the primary package management in kapp-controller.

both of options you are proposing are still optimizing for the wrong case imho. which use case today requires to present that message where it would be helpful?

> We can upfront fetch all the images in a threaded way and afterward just iterate over them

1000s of images in parallel is still 1000s of images. especially for systems that will be calling to imgpkg pull continiously this is very expensive.

pkg/imgpkg/bundle/bundle_test.go
DennisDenuto: Ok, I have updated the test name to indicate that nested bundles shouldn't be pulled (in the context of pulling without -r) and that nested bundles *should* be pulled (in the context of pulling with -r)

pkg/imgpkg/bundle/repo/some-bundle-name/.imgpkg/bundle.yml
DennisDenuto: No. these files were added by accident. purging them asap!

pkg/imgpkg/bundle/test_assets/bundle/.imgpkg/bundle.yml
DennisDenuto: It is used in the bundles_test.go as the most leaf bundle node.

for e.g. `bundle referencing multiple of the same bundles`

pkg/imgpkg/bundle/bundle_test.go
DennisDenuto: hmmm, `doNotPullNestedBundles := false` is a double negative and will read as "pull nested bundles", which is the opposite of what this test is trying to do.

As part of a previous feedback you gave, this variable is now named: `pullNestedBundles := false`

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: Thx. I try.

pkg/imgpkg/bundle/fake_images_metadata_builder_test.go
DennisDenuto: Yeah the smaller implementation def helped with writing the tests, and I think also helps with reading the tests too. 

It also helped establish what the test API needed to write these tests should be. Which is `WithBundleFromPath` and `WithEveryImageFrom`.

Yeah, We might want to consider creating tests for our fake registries, if they start to get a bit too big. and then tests for those tests if they get too big... lol jk

I think another approach I have been meaning to try, is now that we have a better understanding on what the 'test API' should be, we do something similar to what google container registry does. Which is spin up an actual test registry with http endpoints `httptest.NewServer(registry.New())` <- This fake registry builder would just populate the state (bundles from disk) and use `registry.New()` handler to serve images. 

I do think some consolidation is in order, probably after we get some convergence around our tests.

Happy to chat more in person to strategize with you over this.

pkg/imgpkg/bundle/bundle_test.go
DennisDenuto: Ok, i'll add a digest to the return `BundleInfo`

pkg/imgpkg/bundle/bundle_test.go
DennisDenuto: Ok i'll delete that 1 level test.

pkg/imgpkg/image/dir_image.go
DennisDenuto: done!

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Ok, as discussed, 
Will drop logging
nested bundle output && the message `To retrieve the configuration of all the nested bundles use the --recursive flag`

Parallelism will be introduced as a separate piece of work and not part of this PR.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Ok. done

pkg/imgpkg/bundle/bundle.go
```diff
@@ -30,7 +32,7 @@ type Bundle struct {
 }
 
 func NewBundle(ref string, imagesMetadata ctlimg.ImagesMetadata) *Bundle {
-	return NewBundleWithReader(ref, imagesMetadata, &singleLayerReader{})
+	return &Bundle{plainimg.NewPlainImage(ref, imagesMetadata), imagesMetadata, &singleLayerReader{}}
```
joaopapereira: Any particular reason not to use the `NewBundleWithReader`
 constructor?

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func (o *Bundle) AllImagesLock() (*ImagesLock, error) {
-	return  o.buildAllImagesLock(map[string]struct{}{})
+	return o.buildAllImagesLock(map[string]struct{}{})
```
joaopapereira, line 19: good catch :D

joaopapereira (commented):
Added a question

pkg/imgpkg/plainimage/plain_image.go
```diff
@@ -129,7 +129,6 @@ func (i *PlainImage) Pull(outputPath string, ui ui.UI) error {
 	}
 
 	ui.BeginLinef("Pulling image '%s'\n", i.DigestRef())
-
```
cppforlife: nit: lets keep newline

pkg/imgpkg/image/dir_image.go
```diff
@@ -48,7 +48,7 @@ func (i *DirImage) AsDirectory() error {
 			return err
 		}
 
-		i.logger.BeginLinef("Extracting layer '%s' (%d/%d)\n", digest, idx+1, len(layers))
+		i.logger.BeginLinef("  Extracting layer '%s' (%d/%d)\n", digest, idx+1, len(layers))
```
cppforlife: this seems like a weird spot to add indent. this class has no idea what context it's in (it seems that you wanted to indent this lines because they were under some heading). instead we should inject NewDirImage with indenting ui so that knowledge of indentation is localized to bundle's Pull.

pkg/imgpkg/bundle/bundle_images_lock.go
cppforlife: why didnt go fmt in hack/build not pick this up?

DennisDenuto: Thx! I try.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Nope. Changed!

pkg/imgpkg/bundle/contents.go
```diff
@@ -19,6 +19,7 @@ import (
 
 const (
 	ImgpkgDir      = ".imgpkg"
+	SubBundlesDir  = "bundles"
```
cppforlife: SubBundlesDir -> BundlesDir

pkg/imgpkg/bundle/bundle.go
```diff
@@ -45,32 +47,95 @@ func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
...
+func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool, bundlePath string, imagesProcessed map[string]bool, numSubBundles int) error {
 	img, err := o.checkedImage()
 	if err != nil {
 		return err
 	}
 
-	ui.BeginLinef("Pulling bundle '%s'\n", o.DigestRef())
+	if o.rootBundle(bundlePath) {
+		ui.BeginLinef("Pulling bundle '%s'\n", o.DigestRef())
+		ui.BeginLinef("Bundle Layers\n")
```
cppforlife: is this heading only printed for the root bundle? if so, it's a little weird to print it for root bundle and not for layers of nested bundles.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -45,32 +47,95 @@ func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
...
+
+	err = NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).WriteToPath(filepath.Join(baseOutputPath, bundlePath), imagesLockUI)
 	if err != nil {
 		return fmt.Errorf("Rewriting image lock file: %s", err)
 	}
 
 	return nil
 }
 
+func subBundlePath(bundleDigest name.Digest) string {
```
cppforlife: let's put it on `func (o *Bundle)`

pkg/imgpkg/bundle/bundle.go
```diff
@@ -45,32 +47,95 @@ func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
...
+			numSubBundles++
+
+			if o.shouldPrintNestedBundlesHeader(bundlePath, numSubBundles) {
+				ui.BeginLinef("\nNested bundles\n")
+			}
+			bundleDigest, err := name.NewDigest(image.Image)
+			if err != nil {
+				return err
+			}
+			err = subBundle.pull(baseOutputPath, goui.NewIndentingUI(ui), pullNestedBundles, subBundlePath(bundleDigest), imagesProcessed, numSubBundles)
```
cppforlife: nit: format

```
err = subBundle.pull(baseOutputPath, goui.NewIndentingUI(ui),
    pullNestedBundles, subBundlePath(bundleDigest), imagesProcessed, numSubBundles)
```

pkg/imgpkg/bundle/bundle.go
```diff
@@ -45,32 +47,95 @@ func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
 func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
-func (o *Bundle) Pull(outputPath string, ui ui.UI) error {
+func (o *Bundle) Pull(outputPath string, ui goui.UI, pullNestedBundles bool) error {
+	return o.pull(outputPath, ui, pullNestedBundles, "", map[string]bool{}, 0)
+}
+
+func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool, bundlePath string, imagesProcessed map[string]bool, numSubBundles int) error {
```
cppforlife: nit: format

```
func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool, 
       bundlePath string, imagesProcessed map[string]bool, numSubBundles int) error {
```

pkg/imgpkg/bundle/bundle.go
```diff
@@ -86,3 +151,9 @@ func (o *Bundle) checkedImage() (regv1.Image, error) {
 	}
 	return img, err
 }
+
+type noopWriter struct{}
+
+func (noopWriter) Write(_ []byte) (n int, err error) {
+	return 0, nil
```
cppforlife: len(bs) should be returned instead of 0 to follow io.Writer interface.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -45,32 +47,95 @@ func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
...
+			err = subBundle.pull(baseOutputPath, goui.NewIndentingUI(ui), pullNestedBundles, subBundlePath(bundleDigest), imagesProcessed, numSubBundles)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	imagesLockUI := ui
+	if !o.rootBundle(bundlePath) {
+		imagesLockUI = goui.NewWriterUI(noopWriter{}, noopWriter{}, goui.NoopLogger{})
```
cppforlife: there is a NewNoopUI in go-cli-ui module.

pkg/imgpkg/plainimage/plain_image.go
DennisDenuto: ok. Done

pkg/imgpkg/image/dir_image.go
DennisDenuto: Ok Done.

pkg/imgpkg/bundle/contents.go
DennisDenuto: Ok. Done

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Yes, this was being only printed for the root bundle. Went ahead and dropped  printing "Bundle Layers" when a root bundle is being processed.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: ok done.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Ok.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Ok.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Ok, yeah I love deleting code!

pkg/imgpkg/bundle/bundle.go
DennisDenuto: This struct was deleted in favor of using the `goui.NewNoopUI`

joaopapereira (approved):
LGTM

PR #102, opened 2021-03-10
Refactor bundle_test.go fake registry

DennisDenuto:
- Bumps google-container-registry library
- Uses a [v2 compliant registry](https://github.com/vmware-tanzu/carvel-imgpkg/blob/508de4948bdd754ae0b0caca286e7dc58f1f63b0/vendor/github.com/google/go-containerregistry/pkg/registry/registry.go#L24) for tests to read/write to. This is a change from having to 'mock' out the `ImagesMetadata` functions.
  - This pattern can be (but hasn't in this PR) apply to [helpers.fake_registry](https://github.com/vmware-tanzu/carvel-imgpkg/blob/56c87b10541d207cf38fad9984e49caba78d4f91/test/helpers/fake_registry.go#L27-L26)
  - Should make understanding our test code simpler (since there is less logic around mocking our registry object)


pkg/imgpkg/bundle/bundle_test.go
```diff
@@ -82,7 +84,12 @@ func TestPullAllNestedBundlesWritingContentsToDisk(t *testing.T) {
 		defer fakeImagesMetadataBuilder.CleanUp()
 
 		fakeImagesMetadataBuilder.WithBundleFromPath("repo/some-bundle-name", "test_assets/bundle").WithEveryImageFrom("test_assets/image_with_config", map[string]string{})
-		subject := bundle.NewBundle("repo/some-bundle-name", fakeImagesMetadataBuilder.Build())
+
+		build := fakeImagesMetadataBuilder.Build()
+		u, err := url.Parse(fakeImagesMetadataBuilder.server.URL)
```
joaopapereira: Isn't this equivalent to
`fakeImagesMetadataBuilder.relocateToTestServer("repo/some-bundle-name")`

pkg/imgpkg/bundle/fake_images_metadata_builder_test.go
```diff
@@ -127,41 +86,37 @@ func (r *FakeImagesMetadataBuilder) WithImageFromPath(imageNameFromTest string,
...
+func (r *FakeTestRegistryBuilder) CleanUp() {
 	for _, tarPath := range r.state {
 		os.Remove(filepath.Join(tarPath.path, ".imgpkg", "images.yml"))
 	}
+	if r.server != nil {
+		r.server.Close()
+	}
+}
+
+func (r *FakeTestRegistryBuilder) relocateToTestServer(repo string) string {
```
joaopapereira:
```suggestion
func (r *FakeTestRegistryBuilder) referenceOnTestServer(repo string) string {
```

Since there are no relocations in `imgpkg` maybe it makes sense to call relocation to this method. The above is a suggestion.

pkg/imgpkg/bundle/fake_images_metadata_builder_test.go
```diff
@@ -8,94 +8,53 @@ import (
...
+	"github.com/k14s/imgpkg/pkg/imgpkg/imagelayers"
 	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
+	"github.com/stretchr/testify/assert"
 )
 
-type FakeImagesMetadataBuilder struct {
-	state map[string]*ImageWithTarPath
-	t     *testing.T
+type FakeTestRegistryBuilder struct {
+	state  map[string]*ImageWithTarPath
```
joaopapereira:
```suggestion
	images  map[string]*ImageWithTarPath
```

This is no longer the state but the list of images we are going to push to the registry.

pkg/imgpkg/bundle/fake_images_metadata_builder_test.go
joaopapereira: Looking at the rest of the functions, is there any reason for this function to be private?

joaopapereira (changes requested):
Nice to see we are keeping our dependencies up to date 😄. It looks like a great improvement since the implementation of our fake was starting to get pretty hard to follow.

Since we are simplifying this fake maybe it would make sense to do the same for the `helpers` package version. 
The major difference that I see is that the other fake represents multiple registries with different URLs, which might become challenging. Unless we spin up multiple registries via `NewFakeImagesMetadataBuilder`

pkg/imgpkg/bundle/bundle_test.go
DennisDenuto: Good catch. Thx I made the change

pkg/imgpkg/bundle/fake_images_metadata_builder_test.go
DennisDenuto: Good catch. Thx I made the change

pkg/imgpkg/bundle/fake_images_metadata_builder_test.go
DennisDenuto: Good catch. Thx I made the change

joaopapereira (approved):
LGTM

joaopapereira:
I talked with @DennisDenuto talked about the update of `helpers` package registry and agreed that we could do this in the next story that will touch the tests that will use this Fake.

Issue #86, opened 2021-02-19
Copy Recursive Bundles from Lock file to registry


joaopapereira:
## Use Case

**As A** Software Packager
**I want to** Create a bundle to distribute the multiple applications that are part of a deployment
**So that** Application consumers can retrieve all needed software from a single place

## Prerequirements
The following images exist in the registry

- `my.registry.io/bundle-1@sha256:aaaaaaaaaafdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0`
  The `.imgpkg/images.yml` looks like the following
  ```yaml
  apiVersion: imgpkg.carvel.dev/v1alpha1
  kind: ImagesLock
  images:
  - image: my.registry.io/img-1@sha256:bbbbbbbbbbfdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
  - image: my.registry.io/bundle-2@sha256:ddddddddddfdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
  ```

- `my.registry.io/bundle-2@sha256:ddddddddddfdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0`
  The `.imgpkg/images.yml` looks like the following
  ```yaml
  apiVersion: imgpkg.carvel.dev/v1alpha1
  kind: ImagesLock
  images:
  - image: my.registry.io/img-2@sha256:eeeeeeeeeefdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
  ```

- `my.registry.io/img-1@sha256:bbbbbbbbbbfdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0`
- `my.registry.io/img-2@sha256:eeeeeeeeeefdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0`

### 🟢 Scenario 1: User copies recursive bundles between repositories using `--experimental-recursive-bundle` using a bundle lock file
**Given** the following file exists in `r-bundle/.imgpkg/images.yml`
```yaml
apiVersion: imgpkg.carvel.dev/v1alpha1
kind: ImagesLock
images:
- image: my.registry.io/bundle-1@sha256:aaaaaaaaaafdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
[...]

PR #104, opened 2021-03-15
Copy bundles from Lock file

DennisDenuto:
This is a draft (so so draft it could be a beer 🍺) PR to increase visibility on changes being made. As well as keep track of any design decisions being made as work is iterated on.

- Fixes e2e test. (e2e were failing when run against dockerhub)
- Always copy recursively from lockfile (does not yet honor the recursive flag)
- Refactor helpers fake_registry.go <- spins up a v2 complaint registry. This results in a lot less mock code that has to be maintained.

EDIT: No longer a draft PR :-)


DennisDenuto:
Task List:

- [x] Maintain behavior without --experimental-recursive-bundle on simple bundles using a bundle lock file
- [x] Fail without --experimental-recursive-bundle on recursive bundle bundles using a bundle lock file
- [x]  Fail without --experimental-recursive-bundle on recursive bundle bundles using a bundle name (-b)
- [x] Fails when try to copy bundle using an image lock file
- [x] reduce verbosity of fake registry
- [x] replace 'duplicate' fake_images_metadata_builder_test for helpers.fake_registry

DennisDenuto:
While implementing Scenario 3 from issue https://github.com/vmware-tanzu/carvel-imgpkg/issues/86, I noticed that issue https://github.com/vmware-tanzu/carvel-imgpkg/issues/85 hadn't implemented the same error case scenario. So as part of this PR, in order to maintain consistency between a bundle lock file and a bundle name, I will also implement scenario 3 for issue #85 .

This has performance implications. In order to know whether a bundle contains a nested bundle, a request needs to be made to the registry.

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -175,6 +187,26 @@ func (o CopyRepoSrc) getSourceImages() (*ctlimgset.UnprocessedImageRefs, error)
...
+	}
+	for _, i := range imagesLock.Images {
+		isBundle, err := ctlbundle.NewBundle(i.Image, o.registry).IsBundle()
+		if err != nil {
+			return lockconfig.ImagesLock{}, err
+		}
+		if isBundle {
+			return lockconfig.ImagesLock{}, fmt.Errorf("This bundle contains bundles, in order to copy please use the --experimental-recursive-bundle flag to copy nested bundles")
+		}
+	}
```
cppforlife: no need to check this here since push is verifying this: https://github.com/vmware-tanzu/carvel-imgpkg/blob/f0aadec0ec3e58bf4927be4f7ad87bd112a585dc/pkg/imgpkg/bundle/contents.go#L89-L92 (was already prepped for recursive bundles with a allowInnerBundles bool.

DennisDenuto: Hmmm i'm not following / missing a nuance here...

This check happens during a `copy`. Specifically they are trying to copy a bundle that contains nested bundles, but they haven't provided the feature flag. In this case we want to provide a meaningful error message to the user, teaching them about this flag.

From the issue, it is implementing `Scenario 3: Fail without --experimental-recursive-bundle on recursive bundle bundles using a bundle lock file`

cppforlife: if we remove this function before release then this doesnt matter.

in case we keep this function in the next release then this affects performance of regular users that use imgpkg because we would be iterating over each image and fetching its config to check if something is a bundle. this flag was made to ensure that we do not "ship" recursive bundle code too early afaik, not with an intent that we would be encouraging ppl to use it (since it's not ready for consumption). (in other words this code favors experimental users instead of regular users).

(code i linked above prevents users from making bundles with nested bundles if flag is not used, hence we know that user knows about the flag).

DennisDenuto: > if we remove this function before release then this doesnt matter.

True.

I am going to assume we ship a release *with* the feature flag still in... Because, you know, an urgent security fix had to be applied...

In that case, the only risk I can think of is, if someone tries to copy a bundle (that contains a nested bundle), and they do *not* specify the feature flag (because to err is human), then the copy command will succeed (Even though the nested bundles haven't been copied)

Eh, The above concern is probably unlikely to happen in practice. So I'll remove the check and update the tests too to reflect this change in behavior.

Issue #96, opened 2021-02-26
Provide copy progress feedback


danielburrell:
**Describe the problem/challenge you have**
As a user, I am pushing images to a docker registry as part of an airgap deployment. Sometimes this takes a long time and I'm left wondering if progress is being made. I'm not calling imgpkg directly, it's being wrapped.

**Describe the solution you'd like**
Some sort of parsable output indicating the progress being made as if it were an uploading data file. "305k/3000k - 10% ETA 30 seconds" this information is considered useful because it shows how much has been uploaded, how much must be uploaded, expresses that as a percentage for simplicity, and then shows the expected time it will complete.

**Anything else you would like to add:**
[Additional information that will assist in solving the issue.]

pivotaljohn:
Complete agreement, @danielburrell.

It'll be a minute...

To do this, we're preparing a contribution to our upstream dependency (we use GGCR for all our direct registry interactions): https://github.com/google/go-containerregistry/issues/952

With progress reports coming in from GGCR, `imgpkg` can then provide that up-to-date status. 👍🏼 

Watch this space for updates.

DennisDenuto:
There has been a [PR submitted upstream](https://github.com/google/go-containerregistry/pull/967) to go-container-registry that should allow improving the status feedback from imgpkg.

Once that PR has been merged in, and a new version of go-container-registry has been released, we should be able to work on this.

aaronshurley:
Before starting this work, let's add some acceptance criteria here so that we consider what the UX will be.

StevenLocke:
### Current `imgpkg copy` output during a copy of cf for k8s:

```
"copy | importing 45 images..."

"copy | importing index.docker.io/cloudfoundry/eirini-event-reporter-cf-for-k8s@sha256:1798ea550cfb7b1457698dc9a9766ad5333da05c1e0cd29e748b7db34e648e1b -> index.docker.io/k8slt/cf-for-k8s-from-tar-2@sha256:1798ea550cfb7b1457698dc9a9766ad5333da05c1e0cd29e748b7db34e648e1b..."

<the above will be printed as each image is verified and tagged>

"copy | imported 45 images"
"Succeeded"
```

### Progress Bar

Showing bytes copied, total bytes to copy, a percentage of those two, and an estimate for time remaining.
```
"copy | importing 45 images"

[================>_______________________________] (45% ETA: 5m10s)
55555 Kb / 1234567 Kb (100 Kb/s)

"Succeeded"
```

[...]

PR #105, opened 2021-03-16
Improve UX for copy

DennisDenuto:
Because every image is uploaded via MultiWrite, the output no longer
shows each image printed during the upload of the image. This change
now indicates to the user which images are being imported, providing
some level of progress information.

This is a 'small improvement' that we are making, with the idea that will be changed during the implementation for https://github.com/vmware-tanzu/carvel-imgpkg/issues/96 (Leveraging the progress status changes made to the [go container registry](https://github.com/google/go-containerregistry/pull/967)

Co-authored-by: Dennis Leon <leonde@vmware.com>
Co-authored-by: Garrett Cheadle <gcheadle@vmware.com>

### DIFF
--- a/pkg/imgpkg/imageset/image_set.go
+++ b/pkg/imgpkg/imageset/image_set.go
@@ -97,6 +97,19 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
 			}
 			imageOrIndexesToWriteLock.Lock()
 			defer imageOrIndexesToWriteLock.Unlock()
+
+			itemDigest, err := item.Digest()
+			if err != nil {
+				errCh <- err
+				return
+			}
+			importDigestRef, err := regname.NewDigest(fmt.Sprintf("%s@%s", importRepo.Name(), itemDigest))
+			if err != nil {
+				errCh <- err
+				return
+			}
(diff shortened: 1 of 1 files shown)


pkg/imgpkg/imageset/image_set.go
```diff
@@ -97,6 +97,19 @@ func (o *ImageSet) Import(imgOrIndexes []imagedesc.ImageOrIndex,
 			}
 			imageOrIndexesToWriteLock.Lock()
 			defer imageOrIndexesToWriteLock.Unlock()
+
+			itemDigest, err := item.Digest()
+			if err != nil {
+				errCh <- err
+				return
+			}
+			importDigestRef, err := regname.NewDigest(fmt.Sprintf("%s@%s", importRepo.Name(), itemDigest))
```
cppforlife, line 106: the logic for determining the destination ref is elsewhere (one of the other private functions). having it be in two distinct places doesnt smell right to me.

gcheadle-vmware: @DennisDenuto and I took a look and we couldn't identify a private function that determined the destination digest reference of an image. We found one for returning a tag, but we want to print out the digest. The diff shows that we pulled the logging statement from tagAndVerifyItem(), so we copied the prerequisites from that function for the logging.

PR #106, opened 2021-03-18
Copy bundles from lockfile to tar

DennisDenuto:
This is a draft PR to increase visibility on changes being made. As well as keep track of any design decisions being made as work is iterated on.

- re-organize copy_src_test to try and group the different features together.
  - Reduce the number of top-level tests to try and capture the different 'features' of imgpkg `copy`. 
  - Now there are less top-level tests and more nested tests. 
  - The top-level 'happy case' test names are also less verbose. Instead test that deviate from the 'happy case' test will contain the information in the test name. i.e. `TestCopyToTar()` <- happy case `TestCopyToTarWithNonDistributableLayers()` <- tests foreign layers and hence *is* the test that is more verbose.
- backfills test to ensure that copying from a bundlelock to a tarfile works!
- This  PR is built on top of the branch: copy-bundles-from-lock-86 since those changes are needed to implement this story.

### DIFF
--- a/pkg/imgpkg/cmd/copy_repo_src_test.go
+++ b/pkg/imgpkg/cmd/copy_repo_src_test.go
@@ -41,38 +41,6 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-func TestCopyingBundleToRepoWithMultipleRegistries(t *testing.T) {
-	fakeDockerhubRegistry := helpers.NewFakeRegistry(t)
-	defer fakeDockerhubRegistry.CleanUp()
-	fakePrivateRegistry := helpers.NewFakeRegistry(t)
-	defer fakePrivateRegistry.CleanUp()
-
-	sourceBundleName := "library/bundle"
-	destinationBundleName := "library/copied-bundle"
-
-	randomImage1FromDockerhub := fakeDockerhubRegistry.WithRandomImage("random-image1")
-	fakePrivateRegistry.WithImage(sourceBundleName, randomImage1FromDockerhub.Image)
-
-	// test_assets/bundle contains images that live in dockerhub
-	bundleWithImageRefsToDockerhub := fakePrivateRegistry.WithBundleFromPath(sourceBundleName, "test_assets/bundle_with_dockerhub_images").WithImageRefs([]lockconfig.ImageRef{
-		{Image: randomImage1FromDockerhub.RefDigest},
-	})
-
-	subject := subject
-	subject.BundleFlags = BundleFlags{bundleWithImageRefsToDockerhub.RefDigest}
-	subject.registry = fakePrivateRegistry.Build()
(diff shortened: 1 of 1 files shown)


pivotaljohn (commented):
Broadly: all of this feedback is either reinforcing or are suggestions.

---

To determine how well I've revealed intention, I have a check I use:

> Based on this function's/test's name alone (in context), do I "know" what its behavior is without looking at its implementation?

When I do that with the portion of this test file that you've structured, I have been able to say (resoundingly), "why, yes!" 👍🏼 

...

Another lens I'm using on top of tests is to ask,

> Based on only the outline of this test suite, do I "know" the capabilities of this feature?

In this case, again, I feel that the essential elements are _all_ present and illustrated in relationship to each other, so far. I can read "the story" from the suite. 👍🏼 

Other observations:
- 👍🏼  That you've kept the hierarchy flat means that as I read the tests, there's fewer pieces of information I have to hold in my mental cache... making the whole thing easier to read.
- 👍🏼  When an aspect of a test is congruent to that same part from another test, you're using identical phrasing. That consistency allows me to pattern match across the outline, making the whole thing easier to read. It simultaneously surfaces "concepts" that help me develop my mental model of the software
- 🤔  I think you can drop `Copying` from the start of all the top-level test names; the name of this file means (and starts each test sentence with) "Copying from a Repository Source".

pivotaljohn (commented):
formatting:
- manually indent lines that are "too long"
  - otherwise, readers have to scroll or endure lousy indentation from soft-wraps...
  - arbitrarily, "too long" is > 120 chars

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -573,6 +566,39 @@ func TestCopyingToRepoImageContainingOnlyDistributableLayers(t *testing.T) {
...
+kind: ImagesLock
+images:
+- image: %s
+  annotations:
+    my-annotation: first-image
+- image: %s
+  annotations:
+    my-annotation: second-image
+`, image1.RefDigest, image2RefDigest)
+		lockFile, err := ioutil.TempFile("", "images.lock.yml")
```
cppforlife: are we deleting this file anywhere?

DennisDenuto:
@pivotaljohn 

> formatting:
> 
> * manually indent lines that are "too long"
>   
>   * otherwise, readers have to scroll or endure lousy indentation from soft-wraps...
>   * arbitrarily, "too long" is > 120 chars

Thanks for the feedback. done done!

I also dropped the word "Copying" from the test names. It is even less verbose and still communicates the intent of the test :-)

pkg/imgpkg/cmd/copy_repo_src_test.go
DennisDenuto: nope. nice catch. updated the test.

PR #109, opened 2021-03-23
Pull updates images lock files

DennisDenuto:
This is a draft (so so draft it could be a beer 🍺) PR to increase visibility on changes being made. As well as keep track of any design decisions being made as work is iterated on.

- 'first attempt' implementation on updating bundle / nested bundle image lock files. 
-- still requires refactoring on how it was implemented. but it makes the tests go green.
- Need to determine whether when determining if a nested bundle has been relocated, the nested bundle repo is used, or whether the root bundle repo is used (or both?)
- doesn't implement the UI portion of the story yet...

### DIFF
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -48,11 +48,16 @@ func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
 func (o *Bundle) Pull(outputPath string, ui goui.UI, pullNestedBundles bool) error {
-	return o.pull(outputPath, ui, pullNestedBundles, "", map[string]bool{}, 0)
+	_, err := o.checkedImage()
+	if err != nil {
+		return err
+	}
+
+	return o.pull(outputPath, ui, pullNestedBundles, "", map[string]bool{}, 0, o.Repo())
 }
 
-func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
-	bundlePath string, imagesProcessed map[string]bool, numSubBundles int) error {
+func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool, bundlePath string,
+	imagesProcessed map[string]bool, numSubBundles int, rootBundleRepo string) error {
 	img, err := o.checkedImage()
 	if err != nil {
 		return err
@@ -104,8 +109,8 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 			if err != nil {
 				return err
 			}
-			err = subBundle.pull(baseOutputPath, goui.NewIndentingUI(ui),
-				pullNestedBundles, o.subBundlePath(bundleDigest), imagesProcessed, numSubBundles)
+			err = subBundle.pull(baseOutputPath, goui.NewIndentingUI(ui), pullNestedBundles,
+				o.subBundlePath(bundleDigest), imagesProcessed, numSubBundles, rootBundleRepo)
 			if err != nil {
 				return err
 			}
@@ -117,7 +122,8 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 		imagesLockUI = goui.NewNoopUI()
 	}
 
-	err = NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).WriteToPath(filepath.Join(baseOutputPath, bundlePath), imagesLockUI)
+	err = NewImagesLock(imagesLock, o.imgRetriever, rootBundleRepo).WriteToPath(
+		filepath.Join(baseOutputPath, bundlePath), imagesLockUI)
 	if err != nil {
 		return fmt.Errorf("Rewriting image lock file: %s", err)
 	}
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -48,16 +48,11 @@ func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
 func (o *Bundle) Pull(outputPath string, ui goui.UI, pullNestedBundles bool) error {
-	_, err := o.checkedImage()
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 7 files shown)


DennisDenuto:
Task List:
- [x] assert on UI output on tests-
- [x]  ~~Implement the UI portion when re-writing images lock files~~
- [x] confirm whether a nested bundle is co-located based on the nested bundle repo, or the 'root bundle' repo.
- [x] re-organize the bundle_test.go. improve hierarchy and naming of tests / sub tests 
- [x] ~~refactor the bundle re-write images lock file implementation.~~
- [x] sanity check commits. we squashed a few commits and only expect most of the changes in bundle and bundle_test.go

DennisDenuto:
In an attempt to come up with a simplified implementation, we (@gcheadle-vmware and I) decided on the following:

- Re-writing a (bundle's and nested bundle's) ImagesLockFile only occurs  *if* every image (and nested bundle image) is *co-located* to the 'root bundles' repo.

# Scenario 1 
registry.io/repo/a
- Root Bundle A

registry.io/repo/b
- Bundle B
- Image C

*Root Bundle A (repo/a)* `- image-references -` *Bundle B (repo/b)*
*Bundle B (repo/b)* `- image-references-` *Image C (repo/b)*

results in:
No imageslock files being updated (even though Image C had been co-located to repo/b)

# Scenario 2 
registry.io/repo/a
- Root Bundle A
- Bundle C

registry.io/repo/b
- Bundle B
- Bundle C

*Root Bundle A (repo/a)* `- image-references -` *Bundle B (repo/b)*
*Bundle B (repo/b)* `- image-references-` *Bundle C (repo/b)* 

results in:
Bundle C imageslock file is updated because Bundle C has been co-located with the root Bundle A
Bundle A and Bundle B imageslock files is *not* updated

pkg/imgpkg/bundle/bundle.go
```diff
@@ -117,7 +122,8 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 		imagesLockUI = goui.NewNoopUI()
 	}
 
-	err = NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).WriteToPath(filepath.Join(baseOutputPath, bundlePath), imagesLockUI)
+	err = NewImagesLock(imagesLock, o.imgRetriever, rootBundleRepo).WriteToPath(
```
cppforlife: hmm using rootBundleRepo vs o.Repo() is leading me to think there is a problem here.

above we are doing 

```
subBundle := NewBundle(image.Image, o.imgRetriever)
isBundle, err := subBundle.IsBundle()
```

and im guessing that image.Image is not localized (is this true?) which then presents a problem since we would try to fetch bundle from original location. if thats true then we should be using a localized copy of the image since we want to be pulling stuff from local place and if we do initialize NewBundle with correct image, then we can continue using o.Repo() for ImagesLock().

DennisDenuto: Oh yeah, good catch. We changed the impl to localize the image reference to the root bundle repo, prior to recursing to the next bundle.

 As for using `o.Repo()` instead of using the `rootBundleRepo`.... When we 'localize' nested bundles that are 2+ levels nested deep, should it be localized relative to the root bundle Repo or the 'parent' bundle repo? 

for more details see the above comment: https://github.com/vmware-tanzu/carvel-imgpkg/pull/109#issuecomment-805336588

The main driver at the time was to simplify the impl, but I think changing it back to localizing to the 'parent' bundle instead of the 'root' bundle might actually be simpler. 

Happy to hear your thoughts on this one.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -48,11 +48,16 @@ func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
 func (o *Bundle) Pull(outputPath string, ui goui.UI, pullNestedBundles bool) error {
-	return o.pull(outputPath, ui, pullNestedBundles, "", map[string]bool{}, 0)
+	_, err := o.checkedImage()
+	if err != nil {
+		return err
+	}
```
cppforlife: not sure why we need this section here, first few lines of o.pull does the same thing.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -74,8 +79,13 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 		return err
 	}
 
+	localizedImagesLockToRootRepo, notLocalizedToRootBundle, err := NewImagesLock(imagesLock, o.imgRetriever, rootBundleRepo).LocalizeImagesLock()
```
cppforlife: picking conversation from our previous thread...

i think we should be localizing against the localized-repo location (not root location) since each bundle is "independent". (in our particular case both of them would be the same since copy cmd is localized everything into one repo, but that's really a decision of copy to make). so since we are now using `localizedImagesLockToRootRepo.Images` to loop over images and use that to make NewBundle with correct bundle location, we can use o.Repo() in NewImagesLock. this should also eliminate rootBundleRepo arg.

pkg/imgpkg/bundle/bundle.go
cppforlife: ah i see because we use o.Repo()...

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Ok. I've made the change to localize against the 'parent' bundle instead of the 'root' bundle.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -117,9 +121,16 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 		imagesLockUI = goui.NewNoopUI()
 	}
 
-	err = NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).WriteToPath(filepath.Join(baseOutputPath, bundlePath), imagesLockUI)
-	if err != nil {
-		return fmt.Errorf("Rewriting image lock file: %s", err)
+	imagesLockUI.BeginLinef("\nLocating image lock file images...\n")
```
cppforlife, line 124: since we added printing to here, do we need this method? https://github.com/vmware-tanzu/carvel-imgpkg/blob/400d8e97f430c1e800efaf21a018a636f22dbc99/pkg/imgpkg/bundle/images_lock.go#L58

DennisDenuto: Nice spot. deleted this unused function.

PR #110, opened 2021-03-25
Use ggcr library to write images with non-distributable layers

DennisDenuto:
There has been an upstream change made to ggcr that allows configuring `Write` and `MultiWrite` to write layers that have a media type of 'non-distributable'.

This PR removes the previous approach of handling writing non-dist layers, and instead delegates to the library.

Another nice benefits (apart from less code) is potential performance improvement due to de-duping layer logic now also applied to non-dist layers

### DIFF
--- a/pkg/imgpkg/cmd/copy.go
+++ b/pkg/imgpkg/cmd/copy.go
@@ -18,17 +18,17 @@ import (
 )
 
 type CopyOptions struct {
-	ImageFlags                  ImageFlags
-	BundleFlags                 BundleFlags
-	LockInputFlags              LockInputFlags
-	LockOutputFlags             LockOutputFlags
-	TarFlags                    TarFlags
-	RegistryFlags               RegistryFlags
-	IncludeNonDistributableFlag IncludeNonDistributableFlag
-	ExperimentalFlags           ExperimentalFlags
+	ImageFlags        ImageFlags
+	BundleFlags       BundleFlags
+	LockInputFlags    LockInputFlags
+	LockOutputFlags   LockOutputFlags
+	TarFlags          TarFlags
+	RegistryFlags     RegistryFlags
+	ExperimentalFlags ExperimentalFlags
 
-	RepoDst     string
-	Concurrency int
+	RepoDst                 string
+	Concurrency             int
+	IncludeNonDistributable bool
 }
 
 func NewCopyOptions() *CopyOptions {
@@ -57,10 +57,11 @@ func NewCopyCmd(o *CopyOptions) *cobra.Command {
 	o.LockOutputFlags.Set(cmd)
 	o.TarFlags.Set(cmd)
 	o.RegistryFlags.Set(cmd)
-	o.IncludeNonDistributableFlag.Set(cmd)
 	o.ExperimentalFlags.Set(cmd)
 	cmd.Flags().StringVar(&o.RepoDst, "to-repo", "", "Location to upload assets")
 	cmd.Flags().IntVar(&o.Concurrency, "concurrency", 5, "Concurrency")
+	cmd.Flags().BoolVar(&o.IncludeNonDistributable, "include-non-distributable", false,
+		"Include non-distributable layers when copying an image/bundle")
 	return cmd
 }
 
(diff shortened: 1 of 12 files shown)


pkg/imgpkg/cmd/copy.go
```diff
@@ -77,9 +75,12 @@ func (o *CopyOptions) Run() error {
 	logger := ctlimg.NewLogger(os.Stderr)
 	prefixedLogger := logger.NewPrefixedWriter("copy | ")
 
-	registry, err := ctlimg.NewRegistry(o.RegistryFlags.AsRegistryOpts(), imagelayers.NewImageLayerWriterCheck(o.IncludeNonDistributableFlag.IncludeNonDistributable))
+	opts := o.RegistryFlags.AsRegistryOpts()
```
cppforlife: opts -> registryOpts to make it more obvious what this is in context of copy cmd.

pkg/imgpkg/cmd/copy.go
```diff
@@ -77,9 +75,12 @@ func (o *CopyOptions) Run() error {
...
+	opts.IncludeNonDistributableLayers = o.IncludeNonDistributableFlag.IncludeNonDistributable
```
cppforlife: it would be nice if we remove https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/pkg/imgpkg/cmd/non_distributable_flags.go dedicated struct and just inline field on CopyOptions since it's just a single flag used in one command. this would make `o.IncludeNonDistributableFlag.IncludeNonDistributable` into `o.IncludeNonDistributable`  which i think reads a bit better

cppforlife: any thoughts on renaming flag from `--include-non-distributable` to `--include-non-distributable-layers`?

pkg/imgpkg/cmd/copy.go
```diff
@@ -7,10 +7,8 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/google/go-containerregistry/pkg/v1/types"
-	"github.com/k14s/imgpkg/pkg/imgpkg/imagelayers"
```
cppforlife, line 11: sounds like we can actually delete imagelayers package as well?

pkg/imgpkg/image/registry.go
```diff
@@ -35,9 +33,10 @@ const (
 )
 
 type RegistryOpts struct {
-	CACertPaths []string
-	VerifyCerts bool
-	Insecure    bool
+	CACertPaths                   []string
+	VerifyCerts                   bool
+	Insecure                      bool
+	IncludeNonDistributableLayers bool
```
cppforlife, line 40: i would throw in a newline before this since first 3 fields are related to connectivity and this one isnt.

pkg/imgpkg/cmd/copy.go
cppforlife: scratch that i remembered that it was being used by TarWriter. i think given its single use location, we should move contents of imagelayers into tar writer.

pkg/imgpkg/cmd/copy.go
DennisDenuto: ok done.

pkg/imgpkg/cmd/copy.go
DennisDenuto: My thoughts are:
- this is a slightly better name for this flag.
- We have already shipped with this flag. It would be a breaking change. I would want to capture this change in its own issue / separate PR as that brings concerns that are orthogonal to this 'refactor' PR. 
-- I don't have enough concern for this flag rename to create a GH issue for it. My take is if we get feedback from users that this flag is confusing I would prioritize it accordingly. 

I will go ahead and remove the `IncludeNonDistributableFlag` struct and add it to the `CopyOptions` instead.

pkg/imgpkg/cmd/copy.go
DennisDenuto: I'll move the files to the tarwriter package. I do think there is value in keeping the logic of whether a layer should be written in its own struct and not be inlined into the tarwriter itself.

pkg/imgpkg/image/registry.go
DennisDenuto: ok.

pkg/imgpkg/cmd/copy.go
cppforlife: added https://github.com/vmware-tanzu/carvel-imgpkg/issues/112

PR #117, opened 2021-04-02
Automate creating a GH release

DennisDenuto:
Use [goreleaser](https://goreleaser.com/) to create a GH release

Use [github-release-notes](github.com/buchanae/github-release-notes) to generate release notes

TODO:

- [x] add goreleaser github action (trigger a release when a tag is pushed)
- [x] Upload binaries (instead of uploading the tar.gz files)
- [x] Checksum file should include checksums for binaries (not the tar.gz files)
- [x] ensure that binary files are 'reproducible builds' (goreleaser builds the same sha as ./hack/build-binaries.sh)
- [x] Use ld flags to update the version variable (from a tag) when publishing a release
~~- [x] make a carvel member generate a one-time token to provide to the goreleaser github action~~ <- No need. read: https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret
- [x] Release notes should contain checksum (not just as an attached file) <- leave this as a manual step for now
- [x] add a post release step in the GH action, that verifies the SHA of the binaries uploaded via gorelease match up with the SHAs built by a native go binary
- [x] change `hack/build-binary.sh` to generate the version based on the tag the commit is on. since version is now a build flag.
- [x] check if we can restrict network of the github action (only allow requests to github.com)??
- [x] Change github owner (in the .goreleaser.yml file) from DennisDenuto -> vmware-tanzu
~~- [ ] hack/build scripts can use goreleaser to generate binary files~~
- [ ] automate homebrew update https://goreleaser.com/customization/homebrew/
- [ ] automate creating a snapshot of OSM latest
~~- [ ] copy the gorelease GH action into our repo. and in our gh.yml point to the digest of the relocated image~~
- [ ] Update release documentation
  - [ ] review draft release notes and add the checksum to the release notes (goreleaser attaches the file but doesn't add it to the release notes section)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,6 @@
 name: goreleaser
 
 on:
-  pull_request:
   push:
     tags:
       - 'v*'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,8 @@ name: goreleaser
 on:
   pull_request:
   push:
+    tags:
+      - 'v*'
 
 jobs:
   goreleaser:
@@ -17,13 +19,80 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.15
+          go-version: 1.16.0
       -
         name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         if: startsWith(github.ref, 'refs/tags/')
         with:
-          version: ${{ github.event.release.tag_name }}
-          args: release --rm-dist
+          version: 0.162.0
+          args: release --rm-dist --debug
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      -
+        uses: actions/github-script@v4
+        id: get-checksums-from-draft-release
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          result-encoding: string
+          script: |
+            var crypto = require('crypto');
+            const { owner, repo } = context.repo;
+
+            // https://docs.github.com/en/rest/reference/repos#list-releases
+            // https://octokit.github.io/rest.js/v18#repos-list-releases
+            var releases = await github.repos.listReleases({
+              owner: owner,
+              repo: repo
+            });
+
+            var crypto = require('crypto')
+            var fs = require('fs')
+            const url = require('url');
+            const https = require('https');
+
+            checksums = {}
+            for (const r of releases["data"]) {
+              if (r.draft && `refs/tags/${r.tag_name}` == "${{ github.ref }}") {
+                for (const asset of r.assets) {
+                  var release_asset = await github.repos.getReleaseAsset({ headers: {accept: `application/octet-stream`}, accept: `application/octet-stream`, owner: owner, repo: repo, asset_id: asset.id });
+                  const hash = crypto.createHash('sha256');
+
+                  let http_promise = new Promise((resolve, reject) => {
+                    https.get(release_asset.url, (stream) => {
+                      stream.on('data', function (data) {
+                        hash.update(data);
+                      });
+                      stream.on('end', function () {
+                        checksums[asset.name]= hash.digest('hex');
+                        resolve(`${asset.name}`);
+                      });
+                    });
+                  });
+                  await http_promise;
+                }
+              }
+            }
+            console.log(checksums)
+
+            return `${checksums['imgpkg-darwin-amd64']}  ./imgpkg-darwin-amd64
+            ${checksums['imgpkg-linux-amd64']}  ./imgpkg-linux-amd64
+            ${checksums['imgpkg-windows-amd64.exe']}  ./imgpkg-windows-amd64.exe`
+
+      -
+        name: verify uploaded artifacts
+        if: startsWith(github.ref, 'refs/tags/')
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: |
+          set -e -x
+          VERSION=`echo ${{ github.ref }}  | grep -Eo '[0-9].*'`
+
+          ./hack/build-binaries.sh "$VERSION" > ./go-checksums
+          cat ./go-checksums
+          diff ./go-checksums <(cat <<EOF
+          ${{steps.get-checksums-from-draft-release.outputs.result}}
+          EOF
+          )
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,29 @@
+name: goreleaser
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          version: ${{ github.event.release.tag_name }}
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       -
         name: Checkout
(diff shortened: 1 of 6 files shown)


danielhelfand (commented):
The goreleaser.yml looks great. 

Was thinking a bit more on triggering the release. Was wondering if changing the version change commit step to something like `go build -ldflags "-X github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/cmd/version.Version=$RELEASE_VERSION"` and invoking whatever script is used [manually via a GitHub Action](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow). The user could just specify the version to the action and run it.

DennisDenuto:
@danielhelfand I like this idea. I'll add it to the goreleaser.yml file. I'll be working on this prob this Friday.

DennisDenuto:
Adding some concerns around security raised by @jtigger 

1. tons of stuff around GH (including getting into our GitHub Action container)
1. goreleaser itself injecting malicious code
1. modifying the goreleaser/goreleaser image, injecting malicious code


A few ideas towards making this more 'secure':
1. Use the native go binary to build imgpkg and only use goreleaser to update the gh release notes / attach the files built by the native go binary
2. Add a post release step in the GH action, that verifies the SHA of the binaries uploaded via gorelease match up with the SHAs built by a native go binary 
3. relocate/fork the goreleaser github action, into our dockerhub repo to ensure we have more control over it. (At time of writing, to ease with development I am using the public goreleaser github action)

DennisDenuto:
Spoke to @jtigger 

We decided on:
- [x] make a carvel member generate a one-time token to provide to the goreleaser github action
- [x] add a post release step in the GH action, that verifies the SHA of the binaries uploaded via gorelease match up with the SHAs built by a native go binary
- [x] check if we can restrict network of the github action (only allow requests to github.com)??


Couldn't find a way to restrict network access on GH actions. However further reduced the scope of our token to 'write' on content. Allowing it enough permission to create a release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,98 @@
...
+              repo: repo
+            });
+
+            var crypto = require('crypto')
+            var fs = require('fs')
+            const url = require('url');
+            const https = require('https');
+
+            checksums = {}
+            for (const r of releases["data"]) {
```
cppforlife, line 58: this seems a bit backwards to me -- instead of being an authority for checksums for built binaries, we are asking the system to give us binaries back to generate the checksums. we are trusting that upload succeeded, nothing was tampered in any way, and we have successfully downloaded everything.

mya be another angle to think about is how does signing play into this (since we would want to sign a set of checksums)? doesnt goreleaser generate checksums.txt?

.github/workflows/release.yml
```diff
@@ -0,0 +1,98 @@
...
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.0
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
```
cppforlife: should we be locking down to a particular digest? (this would help us to prevent a threat where goreleaser repo is compramised)

.goreleaser.yml
```diff
@@ -0,0 +1,84 @@
...
+    main: ./cmd/imgpkg
+    binary: imgpkg-{{ .Os }}-{{ .Arch }}
+
+    # not sure if the ldflags buildid is needed anymore. check https://github.com/golang/go/issues/16860
+    flags:
+      - -trimpath
+
+    ldflags:
+      - -buildid=
+      - -X github.com/k14s/imgpkg/pkg/imgpkg/cmd.Version={{ .Version }}
```
cppforlife, line 23: what are thoughts behind injecting version here vs in code?

.github/workflows/release.yml
DennisDenuto: hmmm, this part of the release GH action is to verify that the binaries uploaded by goreleaser match the same sha as binaries generated via the `hack/build-binaries.sh`

The concern is that `goreleaser` *could* have been compromised, and thus, uploaded a compromised binary. 

So, compare the sha's of the uploaded binary built with `goreleaser`, with the sha's generated using native `go build` via `hack/build-binaries.sh`

> upload succeeded , nothing was tampered

Perhaps I should just assume nothing was tampered. Although leaving this 'verification' step is hopefully one less attack vector.


FWIW, This decision was largely driven by the above comment: https://github.com/vmware-tanzu/carvel-imgpkg/pull/117#issuecomment-819869853

.goreleaser.yml
DennisDenuto: It results in 1 less manual step done during a release.

Instead of making a code change and then pushing a tag. It changes to: 'push a tag'

If someone wanted to checkout the code and build our project, they can rebuild the *same* binary uploaded via `goreleaser` or via the `hack/build-binary.sh`.

.github/workflows/release.yml
DennisDenuto: this is a good idea. The plan was actually to fork and host our own copy of `goreleaser-action`

mainly driven by: https://github.com/vmware-tanzu/carvel-imgpkg/pull/117#issuecomment-819869853

Although I think for now, a sha is more then sufficient.

.github/workflows/release.yml
cppforlife: > So, compare the sha's of the uploaded binary built with goreleaser, with the sha's generated using native go build via hack/build-binaries.sh

oh very interesting. i think i missed that part. are we able to reliably guarantee that our script is equivalent to whatever goreleaser is producing?

DennisDenuto: > our script is equivalent to whatever goreleaser is producing

Yes, locally and when run as a GH action it produces the same `sha` as our internal `hack/build.sh` script.

also, from the debug logs goreleaser provides, it runs the equivalent `go build` command...

However, `goreleaser` is a binary. And while today it generates the same `sha` it might not tomorrow (assuming it is compromised / malicious)


> i think i missed that part

This script `get-checksums-from-draft-release` downloads the binaries attached to the github draft release and calculates its `sha`. The next step compares the `sha` generated by `hack/build` with the `sha` calculated in this step of the GH action.

By the time this step is running, the token given to `goreleaser` (previous step) should be revocated. i.e. `goreleaser` can't send that token to another server to upload a different binary to our GH release in some future date (during and after this verification step runs)

.github/workflows/release.yml
DennisDenuto: Done

DennisDenuto:
Will move the below remaining TODO items in a separate (future PR)

```
- automate homebrew update https://goreleaser.com/customization/homebrew/
- automate creating a snapshot of OSM latest
- Update release documentation
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
+name: goreleaser
+
+on:
+  pull_request:
```
cppforlife: do we need this for PRs?

.gitignore
```diff
@@ -5,3 +5,4 @@
 /tmp
 .DS_Store
 /test/e2e/assets/simple-app/empty-dir
+dist/
```
cppforlife, line 8: should we move this into tmp/dist?

hack/build.sh
```diff
@@ -2,16 +2,18 @@
 
 set -e -x -u
 
+VERSION="$(git describe --tags | grep -Eo 'v[0-9]+\.[0-9]+\.[0-9]+' || echo 'develop')"
```
cppforlife: any reason to bother to include this in hack/build.sh? it's only used for development.

.github/workflows/release.yml
DennisDenuto: Had a call with dk. Marking this conversation as resolved.

pkg/imgpkg/cmd/version.go
```diff
@@ -10,9 +10,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const (
-	Version = "0.8.0"
-)
+var Version = ""
```
cppforlife: lets set this to `develop`

DennisDenuto: done

.github/workflows/release.yml
DennisDenuto: nope. removed.

hack/build.sh
DennisDenuto: nope. changed.

.gitignore
DennisDenuto: hmm... I don't see a config for goreleaser to make this change. 
but also, goreleaser will practically only ever be run by an ephemeral env (github action). So the `dist` folder should rarely (if ever) make it on a dev workstation.

hack/build-binaries.sh
```diff
@@ -2,14 +2,20 @@
 
 set -e -x -u
 
-./hack/build.sh
+LATEST_GIT_TAG=$(git describe --tags | grep -Eo 'v[0-9]+\.[0-9]+\.[0-9]+')
+VERSION="${1:-$LATEST_GIT_TAG}"
+
+go fmt ./cmd/... ./pkg/... ./test/...
+go mod vendor
+go mod tidy
```
cppforlife, line 10: just noticed this, do we need to do this here?

PR #121, opened 2021-04-07
Support multiple registry auth via environment variables

DennisDenuto:
This is a draft PR to increase visibility on changes being made. As well as keep track of any design decisions being made as work is iterated on.


DennisDenuto:
TODO:

- [x] Rebase commits
- [x] Do the TODO's in the codebase commits 
- [x] Update documentation

DennisDenuto:
Created a [PR](https://github.com/vmware-tanzu/carvel.dev/pull/85) for doc changes

pkg/imgpkg/image/keychain.go
```diff
@@ -0,0 +1,208 @@
...
+	errCredentialsNotFoundMessage = "credentials not found in native keychain"
+	// ErrCredentialsMissingServerURL and ErrCredentialsMissingUsername standardize
+	// invalid credentials or credentials management operations
+	errCredentialsMissingServerURLMessage = "no credentials server URL"
+	errCredentialsMissingUsernameMessage  = "no credentials username"
+
+	globalPrefix = "IMGPKG_REGISTRY"
+)
+
+func Keychain(username string, password string, token string, anon bool) regauthn.Keychain {
```
cppforlife: - it might be good to take a struct with named fields since there are 3 fields that have the same type (could be opts struct that customRegistryKeychain uses). i typically recommend this when types by themselves arent distinguishing enough.
- might be good to take a `environFunc func() []string` argument that would be set to os.Environ and some fake function in tests so that tests do not have to fiddle with real environment of the process (aka global state).

pkg/imgpkg/image/keychain.go
```diff
@@ -0,0 +1,208 @@
...
+	case k.Anon:
+		return regauthn.Anonymous, nil
+	default:
+		return retryDefaultKeychain(func() (regauthn.Authenticator, error) {
+			return regauthn.DefaultKeychain.Resolve(res)
+		})
+	}
+}
+
+func retryDefaultKeychain(doFunc func() (regauthn.Authenticator, error)) (regauthn.Authenticator, error) {
```
cppforlife: lets put retryDefaultKeychain on customRegistryKeychain since it's only used by this class to indicate that its only used there.

pkg/imgpkg/image/keychain.go
```diff
@@ -0,0 +1,208 @@
...
+
+// constants copied from https://github.com/vmware-tanzu/carvel-imgpkg/blob/c8b1bc196e5f1af82e6df8c36c290940169aa896/vendor/github.com/docker/docker-credential-helpers/credentials/error.go#L4-L11
+const (
+	// ErrCredentialsNotFound standardizes the not found error, so every helper returns
+	// the same message and docker can handle it properly.
+	errCredentialsNotFoundMessage = "credentials not found in native keychain"
+	// ErrCredentialsMissingServerURL and ErrCredentialsMissingUsername standardize
+	// invalid credentials or credentials management operations
+	errCredentialsMissingServerURLMessage = "no credentials server URL"
+	errCredentialsMissingUsernameMessage  = "no credentials username"
```
cppforlife: we can probably localize this as consts in the retry function since that's where it's only being used.

pkg/imgpkg/image/keychain.go
```diff
@@ -0,0 +1,208 @@
...
+
+	globalPrefix = "IMGPKG_REGISTRY"
```
cppforlife: i think this can also be inlined since it's only used in one place.

pkg/imgpkg/image/keychain.go
```diff
@@ -0,0 +1,208 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package image
```
cppforlife: it might be a good time to move registry and keychain stuff into their own package since they are both not that related to other classes in image package. i can even see that package being shared between imgpkg and kbld.

pkg/imgpkg/image/keychain_test.go
```diff
@@ -0,0 +1,136 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package image
```
cppforlife: should be in its own package image_test to avoid accidental private state use from image package.

pkg/imgpkg/image/keychain.go
```diff
@@ -0,0 +1,208 @@
...
+	return regauthn.Anonymous, nil
+}
+
+func (k *envKeychain) collect() ([]envKeychainInfo, error) {
+	k.collectLock.Lock()
+	defer k.collectLock.Unlock()
+
+	if k.collected || k.collectErr != nil {
+		return append([]envKeychainInfo{}, k.infos...), k.collectErr
+	}
```
joaopapereira:
```suggestion
	if k.collected {
		return append([]envKeychainInfo{}, k.infos...), nil
	}

	if k.collectErr != nil {
		return nil, k.collectErr
	}
```
I am suggesting this change because it is clearer, at least in my head these 2 possible states. If not you will have to look through the code to see if there is a case where an error has occurred and we collected the keychain information.

pkg/imgpkg/image/keychain_test.go
```diff
@@ -0,0 +1,136 @@
...
+	assert.NoError(t, err)
+
+	assert.Equal(t, &authn.Basic{
+		Username: "flagUser",
+		Password: "flagPass",
+	}, auth)
+}
+
+func setTestEnv(t *testing.T, key string, value string) {
+	assert.NoError(t, os.Setenv(globalPrefix+"_"+key, value))
```
joaopapereira:
```suggestion
	assert.NoError(t, os.Setenv("IMGPKG_REGISTRY_"+key, value))
```

For testing purposes, I prefer to have hardcoded the constant strings. This allows us to catch errors if for some reason these constants change or are incorrect.

pkg/imgpkg/image/keychain_test.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func unsetTestEnv(t *testing.T) {
+	for _, env := range os.Environ() {
+		if strings.HasPrefix(env, globalPrefix) {
```
joaopapereira:
```suggestion
		if strings.HasPrefix(env, "IMGPKG_REGISTRY_") {
```
For testing purposes, I prefer to have hardcoded the constant strings. This allows us to catch errors if for some reason these constants change or are incorrect.

pkg/imgpkg/image/keychain.go
```diff
@@ -0,0 +1,208 @@
...
+
+	for _, env := range os.Environ() {
+		pieces := strings.SplitN(env, "=", 2)
+		if len(pieces) != 2 {
+			continue
+		}
+
+		var matched bool
+
+		for key, updateFunc := range funcsMap {
```
joaopapereira: Since `funcsMap` is a Map I think that there should be a more efficient way to do this

pkg/imgpkg/image/keychain.go
joaopapereira: I like the second suggestion because this function can already filter out the environment variables that we care about so that we do not have to go through all of them.

joaopapereira (changes requested):
I agree with @cppforlife I think we should move the keychain and the repository from the image package since it feels out of place in that package.
Also left some change suggestions

pkg/imgpkg/image/keychain.go
cppforlife: imho filtering still belongs in this class since it's related to interpreting various env vars names. i was trying to highlight that we can do dependency injection (for reading env vars) via a function arg.

pkg/imgpkg/registry/keychain.go
```diff
@@ -0,0 +1,217 @@
...
+
+			if updateFunc, ok := funcsMap[authOpt]; ok {
+				matched = true
+				err := updateFunc(&defaultInfo, pieces[1])
+				if err != nil {
+					k.collectErr = err
+					return nil, k.collectErr
+				}
+			} else {
+				splits := strings.SplitN(authOpt, "_", 2)
```
StevenLocke: This isn't quite right. It would work if our "types" didn't include `_` themselves, but some do. Whoops! Back to the drawing board.

cppforlife:
@joaopapereira @StevenLocke my 2c: this (parsing env vars) is not perf sensitive code so readability here should be prime objective.

joaopapereira:
> @joaopapereira @StevenLocke my 2c: this (parsing env vars) is not perf sensitive code so readability here should be prime objective.

After looking at the implementation there was a point that I overlooked. So maybe we can revert back to the code we had there and when we do that I think we are ok to merge this PR

pkg/imgpkg/registry/keychain.go
```diff
@@ -0,0 +1,214 @@
...
+		result = append(result, info)
+	}
+
+	k.infos = result
+	k.collected = true
+
+	return append([]envKeychainInfo{}, k.infos...), nil
+}
+
+type envKeychainInfo struct {
```
joaopapereira: Lets move this definition near the place it is used, the prior struct

pkg/imgpkg/imagetar/tar_writer.go
```diff
@@ -202,22 +202,22 @@ func (w *TarWriter) writeLayers() error {
 }
 
 func (w *TarWriter) fillInLayers(writtenLayers map[string]writtenLayer) error {
-	var sortedWritterLayers []writtenLayer
+	var sortedWriterLayers []writtenLayer
```
cppforlife: this was meant to be `sortedWrittenLayers`

pkg/imgpkg/registry/keychain.go
```diff
@@ -0,0 +1,214 @@
...
+
+type KeychainOpts struct {
+	Username string
+	Password string
+	Token    string
+	Anon     bool
+}
+
+func Keychain(keychainOpts KeychainOpts, environFunc func() []string) regauthn.Keychain {
+	return regauthn.NewMultiKeychain(customRegistryKeychain{opts: keychainOpts},
```
cppforlife: it seems that customRegistryKeychain falls back to retrying default keychain. this would happen before trying to access envKeychain based on my understanding of NewMultiKeychain. are the tests falling to envKeychain because regauthn.DefaultKeychain is set to anon? this seems like a pretty nasty global variable.

pkg/imgpkg/registry/keychain_test.go
```diff
@@ -0,0 +1,121 @@
...
+	resource, err := name.NewRepository("some.fake.registry/imgpkg_test")
+	assert.NoError(t, err)
+
+	auth, err := keychain.Resolve(resource)
+	assert.NoError(t, err)
+
+	assert.Equal(t, authn.Anonymous, auth)
+}
+
+func TestEnvAuthWhenEnvVarsProvided(t *testing.T) {
```
joaopapereira: Can we add in one of these tests an environment variable that looks like "SOMETHING_USERNAME" to ensure we are only considering the ones that start with `IMGPKG_REGISTRY`?

pkg/imgpkg/registry/keychain_test.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/k14s/imgpkg/pkg/imgpkg/registry"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAnonAuthWhenNoEnvVarsProvided(t *testing.T) {
```
joaopapereira: Please refactor these tests to use the `t.Run` format that we have been using in `imgpkg`

pkg/imgpkg/registry/keychain.go
joaopapereira: Also if you set Anon to true, it will try the next keychain in the list

test/e2e/auth_test.go
```diff
@@ -0,0 +1,77 @@
...
+package e2e
+
+import (
+	"testing"
+
+	"github.com/k14s/imgpkg/test/helpers"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBasicAuth(t *testing.T) {
```
joaopapereira: I would change these tests to also follow the `t.Run` layout we are using in other tests.

joaopapereira (changes requested):
Please consolidate the commits, so that we do not have the extra commit for the change of the map, and the subsequent rollback commit.

pkg/imgpkg/registry/keychain.go
DennisDenuto: Ok i'll add some logic around handling cases when different overlapping credentials are provided to imgpkg. Namely, cli flags, env variables and config.json.

I'll work towards meeting the following cases.

## case 1) username + password is provided via cli flags
MultiKeychain uses customKeyChain

## case 2) anon is provided via cli flags
An "Anonymous" keychain is used (notice that MultiKeychain isn't used)

## case 3) env credentials are provided with the anon cli flag
e.g.
```
REGISTRY_HOSTNAME=some.reg
REGISTRY_USERNAME=foo
imgpkg pull --anon
```
An "Anonymous" keychain is used (notice that MultiKeychain isn't used)

## case 4) env credentials are provided, *and* credentials are also contained in `config.json`
MultiKeychain traverses the keychain (custom keychain -> env keychain -> default keychain) until it finds a keychain to resolve the resource




Note: An "Anonymous keychain" is a yet-to-be created keychain that resolves every resource to "Anonymous"

pkg/imgpkg/registry/keychain.go
DennisDenuto: Done

pkg/imgpkg/registry/keychain_test.go
DennisDenuto: Good idea. Also added a parent test context to test the rules around ordering around using multiple keychains

pkg/imgpkg/registry/keychain.go
cppforlife: > An "Anonymous" keychain is used (notice that MultiKeychain isn't used)

im not sure we would want that. e.g. you might be configuring anonymous access for everything except for providing credentials for a specific registry. it seems at high level that MultiKeyChain does not capture "i found anon case and i want to use it" and instead treats anon case as fallthrough to the next option. how can we continue using MultiKeyChain and support our "explicit" anon case?

DennisDenuto: Ok.

I'll make the ordering the following:

```
1. registry matching feature
  a. IMGPKG_USERNAME_0
2. generic auth
  a. IMGPKG_USERNAME
  b. --username <-- would take precedence over (a)
3. anon flag
4. default
```

pkg/imgpkg/registry/keychain_test.go
DennisDenuto: Ok done

test/e2e/auth_test.go
DennisDenuto: done.

joaopapereira (approved):
LGTM

PR #126, opened 2021-04-13
Bug Fix - stop crashing when copying a tar with an ImageIndex

DennisDenuto:
Opening as a draft as there is still work to do. Namely

- [x] backfill test copying a tarfile with an ImageIndex

### DIFF
--- a/pkg/imgpkg/cmd/copy.go
+++ b/pkg/imgpkg/cmd/copy.go
@@ -8,6 +8,7 @@ import (
 	"os"
 
 	regname "github.com/google/go-containerregistry/pkg/name"
+	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/k14s/imgpkg/pkg/imgpkg/bundle"
 	ctlimg "github.com/k14s/imgpkg/pkg/imgpkg/image"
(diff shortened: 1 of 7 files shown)


pkg/imgpkg/cmd/copy.go
```diff
@@ -242,6 +242,10 @@ func (o *CopyOptions) writeBundleLockOutput(bundle *bundle.Bundle) error {
 func processedImagesMediaType(processedImages *ctlimgset.ProcessedImages) []string {
 	everyMediaType := []string{}
 	for _, image := range processedImages.All() {
+		if image.Image == nil {
```
cppforlife: dont we need to recurse into image index as well?

DennisDenuto: We do! we discovered this while TDD'ing. See updated impl.

pkg/imgpkg/cmd/copy.go
```diff
@@ -242,17 +243,42 @@ func (o *CopyOptions) writeBundleLockOutput(bundle *bundle.Bundle) error {
...
-		for _, layer := range layers {
-			mediaType, err := layer.MediaType()
+		if image.ImageIndex != nil {
+			indexManifest, err := image.ImageIndex.IndexManifest()
 			if err != nil {
 				continue
 			}
-			everyMediaType = append(everyMediaType, string(mediaType))
+			for _, descriptor := range indexManifest.Manifests {
+				image, err := image.ImageIndex.Image(descriptor.Digest)
```
cppforlife: i might be wrong, i thought that indexes could contain other indexes as well.

DennisDenuto: Added logic to recurse through image indexes

joaopapereira (approved):
LGTM

Issue #132, opened 2021-04-22
Improve imgpkg copy UX when copying big bundles 


iamhsk:
**Describe the problem/challenge you have**
When copying a big bundle with lots of images and nested bundles, it takes a long time and users do not get any feedback on what is happening after they run the `imgpkg copy` command. This is happening for both copying `--to-tar` and `--to-repo`.

**Describe the solution you'd like**
As a user, I would like to have some visibility into what is going on behind the scenes and also would like the performance of the actual copying to be faster. 

**Anything else you would like to add:**
Faster! ⚡️

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

joaopapereira:
The main improvements were shipped on version v0.7.0.
The only missing part is the inclusion or not of the spinner that is asked for in this story. Keeping this issue open until we solve this question

aaronshurley:
I believe that the only remaining task for this issue is whether we include a spinner or not. #163 seems to encapsulate that question so I'm going to close this out. Please feel free to reopen if my logic is off.

PR #133, opened 2021-04-22
Improve UX and performance of copy command 

joaopapereira:
Related to #132 

- [x] Remove unused code
- [x] Add more information to the UX about what is happening in the first step of the copy
- [x] Improve performance of the copy command

### DIFF
--- a/pkg/imgpkg/bundle/bundle_images_lock.go
+++ b/pkg/imgpkg/bundle/bundle_images_lock.go
@@ -65,23 +65,6 @@ func (o *Bundle) buildAllImagesLock(processedImgs map[string]struct{}) (*ImagesL
 	return allImagesLock, nil
 }
 
-// ImagesLockLocalized returns possibly modified images lock
-// with image URLs relative to bundle location
-func (o *Bundle) ImagesLockLocalized() (lockconfig.ImagesLock, error) {
-	img, err := o.checkedImage()
-	if err != nil {
-		return lockconfig.ImagesLock{}, err
-	}
-
-	imagesLock, err := o.imagesLockReader.Read(img)
-	if err != nil {
-		return lockconfig.ImagesLock{}, err
-	}
-
-	imagesLock, _, err = NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).LocalizeImagesLock()
-	return imagesLock, err
-}
-
 type singleLayerReader struct{}
 
 func (o *singleLayerReader) Read(img regv1.Image) (lockconfig.ImagesLock, error) {
--- a/pkg/imgpkg/bundle/bundle_images_lock.go
+++ b/pkg/imgpkg/bundle/bundle_images_lock.go
@@ -9,17 +9,20 @@ import (
 	"io"
 	"io/ioutil"
 	"path/filepath"
+	"sync"
 
 	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
+	"github.com/k14s/imgpkg/pkg/imgpkg/util"
 )
 
-func (o *Bundle) AllImagesLock() (*ImagesLock, error) {
-	return o.buildAllImagesLock(map[string]struct{}{})
+func (o *Bundle) AllImagesLock(concurrency int) (*ImagesLock, error) {
+	throttleReq := util.NewThrottle(concurrency)
+	return o.buildAllImagesLock(&throttleReq, &processedImages{processedImgs: map[string]struct{}{}})
 }
 
-func (o *Bundle) buildAllImagesLock(processedImgs map[string]struct{}) (*ImagesLock, error) {
+func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *processedImages) (*ImagesLock, error) {
 	img, err := o.checkedImage()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


joaopapereira:
spinner

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -32,28 +34,38 @@ func (o *Bundle) buildAllImagesLock(processedImgs map[string]struct{}) (*ImagesL
 
 	allImagesLock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo())
 
+	errChan := make(chan error, len(imagesLock.Images))
+	throttle := util.NewThrottle(concurrency)
```
cppforlife: nit: lets add a newline after this

cppforlife: throttle is contextualized to single invocation of buildAllImagesLocks so it does not throttle total number of requests at once. looks like we should pull it out.

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -65,6 +77,36 @@ func (o *Bundle) buildAllImagesLock(processedImgs map[string]struct{}) (*ImagesL
...
+	}
+	return imgLock, nil
+}
+
+type processedImages struct {
+	lock          sync.Mutex
+	processedImgs map[string]struct{}
+}
+
+func (p *processedImages) checkAndAddImage(ref string) bool {
```
cppforlife: should make this method public. it's a method we expect users of processedImages to use. processedImages being private already keeps it private to the package.

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -65,6 +77,36 @@ func (o *Bundle) buildAllImagesLock(processedImgs map[string]struct{}) (*ImagesL
 	return allImagesLock, nil
 }
 
+func (o *Bundle) imagesLockIfIsBundle(image lockconfig.ImageRef, processedImgs *processedImages, concurrency int) (*ImagesLock, error) {
+	bundle := NewBundleWithReader(image.Image, o.imgRetriever, o.imagesLockReader)
+	isBundle, err := bundle.IsBundle()
```
cppforlife, line 85: nit: newline before this

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -22,13 +23,18 @@ type ImagesLock struct {
 	imagesLock     lockconfig.ImagesLock
 	imgRetriever   ctlimg.ImagesMetadata
 	relativeToRepo string
+	mutex          sync.Mutex
```
cppforlife: this lock feels way to far from whats happening in Bundle. have you tried doing the lock around Merge in buildAllImagesLock?

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -94,18 +95,37 @@ func (o *ImagesLock) LocalizeImagesLock() (lockconfig.ImagesLock, bool, error) {
 	return imagesLock, false, nil
 }
 
-func (o *ImagesLock) LocationPrunedImageRefs() ([]lockconfig.ImageRef, error) {
+func (o *ImagesLock) LocationPrunedImageRefs(concurrency int) ([]lockconfig.ImageRef, error) {
 	var imageRefs []lockconfig.ImageRef
+	errChan := make(chan error, len(o.ImageRefs()))
+	throttle := util.NewThrottle(concurrency)
```
cppforlife, line 97: nit: newline after

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -94,18 +95,37 @@ func (o *ImagesLock) LocalizeImagesLock() (lockconfig.ImagesLock, bool, error) {
 	return imagesLock, false, nil
 }
 
-func (o *ImagesLock) LocationPrunedImageRefs() ([]lockconfig.ImageRef, error) {
+func (o *ImagesLock) LocationPrunedImageRefs(concurrency int) ([]lockconfig.ImageRef, error) {
 	var imageRefs []lockconfig.ImageRef
+	errChan := make(chan error, len(o.ImageRefs()))
```
cppforlife, line 96: nit: newline before

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -94,18 +95,37 @@ func (o *ImagesLock) LocalizeImagesLock() (lockconfig.ImagesLock, bool, error) {
...
+	throttle := util.NewThrottle(concurrency)
 	for _, imgRef := range o.ImageRefs() {
-		newImgRef := imgRef.DeepCopy()
+		imgRef := imgRef
```
cppforlife: might as well do DeepCopy here instead of inside go func.

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -94,18 +95,37 @@ func (o *ImagesLock) LocalizeImagesLock() (lockconfig.ImagesLock, bool, error) {
...
+			if err != nil {
+				errChan <- err
+				return
+			}
+
+			newImgRef.DiscardLocationsExcept(foundImg)
+
+			o.mutex.Lock()
+			defer o.mutex.Unlock()
+			imageRefs = append(imageRefs, newImgRef)
```
cppforlife, line 117: nit: newline before

cppforlife:
lets split off any code related to spinner into a separate PR. im still not on the same page with yall about spinners.

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: Yes, I thought about it, but the problem is that if you select the concurrency to 1 and have nested bundles the worker from the outer bundle will be blocked until the inner ones are resolved. Given that we only have 1 worker this will be blocked forever.
So I decided to incur this penalty and ignore the amount of concurrency you asked for in the command line.  Basically, you can specify how many concurrent workers you want running for all Images in a particular Bundle from the command line. So if you have 5 levels of nested bundles you will get 5 x ConcurrencyValueProvided

pkg/imgpkg/bundle/bundle_images_lock.go

pkg/imgpkg/bundle/images_lock.go
joaopapereira: done

pkg/imgpkg/bundle/images_lock.go

pkg/imgpkg/bundle/images_lock.go
joaopapereira: done

pkg/imgpkg/bundle/images_lock.go
joaopapereira: done

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: done

joaopapereira:
Moved the spinner to #134

pkg/imgpkg/bundle/bundle_images_lock.go
cppforlife: good point about blocking. i still think that concurrency should be against total requests in progress. feels problematic if the input (nested bundles) can affect amount of stuff being downloaded at once.

can we not throttle.Take/Done around the actual fetching and not just against the entire go routine?

joaopapereira: I pushed that down near the request now we should be able to throttle the request only

PR #134, opened 2021-04-28
Add spinner to copy

joaopapereira:
This PR is a spinoff of #133 with only the spinner functionality added


cppforlife:
decided not to pull in. can revisit later if necessary.

DennisDenuto:
@cppforlife Can you share your reasoning behind closing / not merging this PR

cppforlife:
PR has been sitting there for awhile. we briefly started talking about spinners and what exactly it helps with (and doesnt). discussion wasnt continued im guessing due to other priorities so i've closed it for now. we can start a discussion again if ppl have time.

pivotaljohn:
I found this bit helpful (from a non-public Slack conversation):

> from what i recall we never got to an agreement regrading this UX paradigm in carvel (spinners arent anywhere in carvel) ... blockage here is on UX.

DennisDenuto:
> from what i recall we never got to an agreement regrading this UX paradigm in carvel (spinners arent anywhere in carvel) ... blockage here is on UX.

Ok... so the original problem (users don't know what is happening in imgpkg because they see a blank screen for a moment) is deemed as a valid problem to solve... but it was the decision to use a spinner to solve that problem that is the issue.

What is the UX paradigm in carvel? Was the option to introduce 'spinners' into this paradigm discussed? or tackled another way, what is the carvel way to show that the program is 'loading' / how do we prompt the user to just chill for a moment?

pivotaljohn:
This about to be a very fruitful conversation... better had in an **open** "discussion" _issue_ rather than a **closed** _pull request_.

Let's take this convo here: #163

PR #141, opened 2021-05-05
Add bundle dir validations

### LINKED ISSUE #115: confusing error message when directory is not a bundle directory
```
$ imgpkg push -b k14stest/dk-test-repo-bundle -f examples/packaging-demo/repo
Error: Expected one '.imgpkg' dir, got 0:
```

seems like we can improve error message to indicate that this directory is not a bundle and needs to have a .imgpkg if it wants to be one. also what's up with trailing `:`?

carlisia:
Fixes #115.

Testing: successfully ran `./hack/test-all-local-registry.sh 5000`.

Notes:
👋   - first time contributor!

I made the assumption that for a bundle to be considered a bundle, it needs to have the `.imgpkg` directory. And for it to be a valid bundle, it needs to also include the `images.yml` file in its directory. I wasn't sure if to qualify as a bundle actually the entire `.imgpkg/images.yml` must exist. If this is so, I would change the validations a bit.

There might be an opportunity to converge more of the related existing tests in the  `push_test.go` file into the table test added inside the `TestBundleDirectoryErrors` test. If this PR is going in the right direction, in regards to converging the other tests, I can:
- do nothing
- include that refactoring in this PR
- open a separate PR just for that.

Let me know!

### DIFF
--- a/pkg/imgpkg/bundle/contents.go
+++ b/pkg/imgpkg/bundle/contents.go
@@ -114,7 +114,7 @@ func (b Contents) validateImgpkgDirs(imgpkgDirs []string) error {
 
 		msg := fmt.Sprintf("This directory is not a bundle. It it is missing %s", imgpkgPath)
 		if len(imgpkgDirs) > 0 {
-			msg = fmt.Sprintf("This directory is not a bundle. It it is missing a single instance of %s and instead has %s", imgpkgPath, strings.Join(imgpkgDirs, ", "))
+			msg = fmt.Sprintf("This directory constains multiple bundle definitions. Only a single instance of %s can be provided and instead these were provided %s", imgpkgPath, strings.Join(imgpkgDirs, ", "))
 		}
 
 		return bundleValidationError{msg}
@@ -131,7 +131,7 @@ func (b Contents) validateImgpkgDirs(imgpkgDirs []string) error {
 		if filepath.Dir(path) == flagPath {
 			imgpkgPath := filepath.Join(path, ImagesLockFile)
 			if _, err := os.Stat(imgpkgPath); os.IsNotExist(err) {
-				msg := fmt.Sprintf("The bundle expected '%s' to exist, but it wasn't found", imgpkgPath)
+				msg := fmt.Sprintf("The bundle expected .imgpkg/images.yml to exist, but it wasn't found in the path %s", imgpkgPath)
 
 				return bundleValidationError{msg}
 			}
--- a/pkg/imgpkg/bundle/contents.go
+++ b/pkg/imgpkg/bundle/contents.go
@@ -110,14 +110,18 @@ func (b *Contents) findImgpkgDirs() ([]string, error) {
 
 func (b Contents) validateImgpkgDirs(imgpkgDirs []string) error {
 	if len(imgpkgDirs) != 1 {
-		return bundleValidationError{
-			fmt.Sprintf("Expected one '%s' dir, got %d: %s",
-				ImgpkgDir, len(imgpkgDirs), strings.Join(imgpkgDirs, ", "))}
+		imgpkgPath := filepath.Join(ImgpkgDir, ImagesLockFile)
+
+		msg := fmt.Sprintf("This directory is not a bundle. It it is missing %s", imgpkgPath)
+		if len(imgpkgDirs) > 0 {
+			msg = fmt.Sprintf("This directory is not a bundle. It it is missing a single instance of %s and instead has %s", imgpkgPath, strings.Join(imgpkgDirs, ", "))
+		}
+
+		return bundleValidationError{msg}
 	}
 
-	path := imgpkgDirs[0]
-
 	// make sure it is a child of one input dir
+	path := imgpkgDirs[0]
 	for _, flagPath := range b.paths {
 		flagPath, err := filepath.Abs(flagPath)
 		if err != nil {
@@ -125,13 +129,21 @@ func (b Contents) validateImgpkgDirs(imgpkgDirs []string) error {
 		}
 
 		if filepath.Dir(path) == flagPath {
+			imgpkgPath := filepath.Join(path, ImagesLockFile)
+			if _, err := os.Stat(imgpkgPath); os.IsNotExist(err) {
+				msg := fmt.Sprintf("The bundle expected '%s' to exist, but it wasn't found", imgpkgPath)
+
+				return bundleValidationError{msg}
+			}
+
 			return nil
 		}
 	}
 
-	return bundleValidationError{
-		fmt.Sprintf("Expected '%s' directory, to be a direct child of one of: %s; was %s",
-			ImgpkgDir, strings.Join(b.paths, ", "), path)}
+	msg := fmt.Sprintf("Expected '%s' directory, to be a direct child of one of: %s; was %s",
+		ImgpkgDir, strings.Join(b.paths, ", "), path)
+
+	return bundleValidationError{msg}
 }
 
 type bundleValidationError struct {
--- a/pkg/imgpkg/cmd/push_test.go
+++ b/pkg/imgpkg/cmd/push_test.go
@@ -11,7 +11,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/k14s/imgpkg/pkg/imgpkg/bundle"
+	"github.com/stretchr/testify/require"
 )
 
 const emptyImagesYaml = `apiVersion: imgpkg.carvel.dev/v1alpha1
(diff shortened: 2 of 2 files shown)


pkg/imgpkg/bundle/contents.go
```diff
@@ -110,28 +110,40 @@ func (b *Contents) findImgpkgDirs() ([]string, error) {
...
-		return bundleValidationError{
-			fmt.Sprintf("Expected one '%s' dir, got %d: %s",
-				ImgpkgDir, len(imgpkgDirs), strings.Join(imgpkgDirs, ", "))}
-	}
+		imgpkgPath := filepath.Join(ImgpkgDir, ImagesLockFile)
 
-	path := imgpkgDirs[0]
+		msg := fmt.Sprintf("This directory is not a bundle. It it is missing %s", imgpkgPath)
+		if len(imgpkgDirs) > 0 {
+			msg = fmt.Sprintf("This directory is not a bundle. It it is missing a single instance of %s and instead has %s", imgpkgPath, strings.Join(imgpkgDirs, ", "))
```
joaopapereira:
```suggestion
			msg = fmt.Sprintf("This directory constains multiple bundle definitions. Only a single instance of %s can be provided and instead these were provided %s", imgpkgPath, strings.Join(imgpkgDirs, ", "))
```

pkg/imgpkg/bundle/contents.go
```diff
@@ -110,28 +110,40 @@ func (b *Contents) findImgpkgDirs() ([]string, error) {
...
 	// make sure it is a child of one input dir
+	path := imgpkgDirs[0]
 	for _, flagPath := range b.paths {
 		flagPath, err := filepath.Abs(flagPath)
 		if err != nil {
 			return err
 		}
 
 		if filepath.Dir(path) == flagPath {
+			imgpkgPath := filepath.Join(path, ImagesLockFile)
```
joaopapereira, line 132: This is a good catch, I think we were missing this check 👍

pkg/imgpkg/cmd/push_test.go
```diff
@@ -218,15 +255,19 @@ func Cleanup(dirs ...string) {
 }
 
 func createBundleDir(loc, imagesYaml string) error {
-	if imagesYaml == "" {
-		imagesYaml = emptyImagesYaml
-	}
-
-	bundleDir := filepath.Join(loc, ".imgpkg")
+	bundleDir := filepath.Join(loc, bundle.ImgpkgDir)
```
joaopapereira:
```suggestion
	bundleDir := filepath.Join(loc, ".imgpkg")
```
we have this as a string on purpose to ensure that if we change, unintentionally, the const we have tests failing.

pkg/imgpkg/cmd/push_test.go
```diff
@@ -218,15 +255,19 @@ func Cleanup(dirs ...string) {
...
 	}
 
+	if imagesYaml == "" {
+		imagesYaml = emptyImagesYaml
+	}
 	return ioutil.WriteFile(filepath.Join(bundleDir, "images.yml"), []byte(imagesYaml), 0600)
 }
+
+func createEmptyBundleDir(loc string) error {
+	bundleDir := filepath.Join(loc, bundle.ImgpkgDir)
```
joaopapereira:
```suggestion
	bundleDir := filepath.Join(loc, ".imgpkg")
```
Same as the above comment

joaopapereira: Maybe we can call this function from the `createBundleDir` function

pkg/imgpkg/cmd/push_test.go
```diff
@@ -116,26 +119,60 @@ func TestNestedImgpkgDirError(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("Failed to setup test: %s", err)
-	}
-
-	push := PushOptions{FileFlags: FileFlags{Files: []string{pushDir}}, BundleFlags: BundleFlags{Bundle: "foo"}}
-	err = push.Run()
-	if err == nil {
-		t.Fatalf("Expected validations to err, but did not")
+	pushDir := filepath.Join(tempDir, "imgpkg-push-dir")
+	imgpkgPath := filepath.Join(pushDir, bundle.ImgpkgDir, bundle.ImagesLockFile)
```
joaopapereira:
```suggestion
	imgpkgPath := filepath.Join(pushDir, ".imgpkg", "images.yml")
```

pkg/imgpkg/cmd/push_test.go
```diff
@@ -116,26 +119,60 @@ func TestNestedImgpkgDirError(t *testing.T) {
 	}
 }
 
-func TestBundleFlagWithoutDirectoryError(t *testing.T) {
+func TestBundleDirectoryErrors(t *testing.T) {
 	tempDir := os.TempDir()
```
joaopapereira: I think we need to move this `os.TempDir` inside the `t.Run`, because if we parallelize the tests both tests would be running in the same folder at the same time, which could cause an intermittent test failure.  Or am I overlooking something? 
This will complicate a little bit the error message check 😢

joaopapereira (changes requested):
Your assumption is correct, a bundle can only be considered a bundle if it contains the file `.imgpkg/images.yml` at the root of the bundle.

```
$ tree -a this-is-a-bundle
this-is-a-bundle
├── .imgpkg
│   └── images.yml
└── config.yml


$ tree -a not-a-bundle
not-a-bundle/
└── basic-step-2
    ├── .imgpkg
    │   └── images.yml
    └── config.yml

$ tree -a not-a-bundle
not-a-bundle
├── .imgpkg
│   └── bundle.yml
└── config.yml
```

I made some comments, great catch on the missing check.

The table tests look good.

If you are feeling like it, we are converting assertions to `testify`, like this piece you added https://github.com/vmware-tanzu/carvel-imgpkg/pull/141/files#diff-0a47862530dac27041a8ca97493f887bb8b9d816ff5b5bfba379bc29dc76edb0R148 could be just
```go
require.NoError(t, err)
```
If you want to do it is cool, if not that is cool as well. We will eventually get to it later 😄 

Thanks for your contribution 😄

pkg/imgpkg/cmd/push_test.go
carlisia: Ah yes, very good point.

pkg/imgpkg/cmd/push_test.go
carlisia: Hm. Good thinking. But then... the `createBundleDir` would also have to return the `bundleDir` needed in LN 265. Because  `createBundleDir` is called in the context of a test that doesn't need this value, I think it's best not to do this refactoring.

pkg/imgpkg/cmd/push_test.go
carlisia: Argh. Check mate. You're right though. I'll figure something out wrt the error message.

carlisia: Maybe we don't need to include the value of `pushDir` as part of the path. After all, we mostly want to point out that at this juncture `images.yml` is missing, so I think outputting `.imgpkg/images.yml` is plenty good. I'll push a change with that and we can see how it looks.

pkg/imgpkg/cmd/push_test.go
joaopapereira: You are right, let's keep it as is

carlisia:
I believe I have addressed everything.

pkg/imgpkg/cmd/push_test.go
```diff
@@ -116,26 +119,56 @@ func TestNestedImgpkgDirError(t *testing.T) {
...
+			tempDir := os.TempDir()
+			pushDir := filepath.Join(tempDir, "imgpkg-push-dir")
+
+			err := os.Mkdir(pushDir, 0700)
+			require.NoError(t, err)
+
+			if tc.createBundleDir {
+				err = createEmptyBundleDir(pushDir)
+				if err != nil {
+					Cleanup(pushDir)
```
joaopapereira: Do you think that instead of having this here we could do a `defer Cleanup(pushDir)`, around line 146?
That way we would be able to use the assertions from testify?

carlisia: Yes, every time I have this type of code I start out this way then remember that defers are not supposed to be used inside loops. This is because it only executes when the function returns, not the loop.

joaopapereira: In this case, we have a function that will be run in the `t.Run`. Does that make a difference?

carlisia: Great catch! Looks so much better now.

joaopapereira (approved):
LGTM

Thanks for the contribution

Issue #146, opened 2021-05-10
When copying a bundle that have images in a private registry copy fails


joaopapereira:
**What steps did you take:**
Scenario: When a bundle has images that are in a private repository, copy the bundle to a different registry. When trying to copy the bundle again, imgpkg reaches the original location instead of the location the bundle is being copied from.

Private repository:
repo.io/my-img

1. Push bundle to public.repo.io/my-bundle that includes `repo.io/my-img`
2. Copy bundle from `public.repo.io/my-bundle` to `different.repo.io/my-bundle`
3. Try to copy `different.repo.io/my-bundle` to `yet-another.repo.io/my-bundle`

**What happened:**
imgpkg tries to check if the images are bundles, based on the original location instead of using the `PrimaryLocation()` function.

**What did you expect:**
imgpkg should try to check if an image is a bundle, based on the primary location, which first is the same repository as the bundle, and only after should check it from the original location.

**Anything else you would like to add:**
The problem starts in https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/pkg/imgpkg/bundle/bundle_images_lock.go#L41

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.7.0
- Docker registry used (e.g. `Docker HUB`): gcr.io
- OS (e.g. from `/etc/os-release`): macOS Big Sur

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

joaopapereira:
Released in version 0.8.0

PR #148, opened 2021-05-12
Collect images from bundle repo

joaopapereira:
Change the way the images are collected from all the bundles.
Previously we were relying on `ImageRef.Image` to check if the Image was a Bundle and to retrieve the Images from it. This is not correct since that location might be behind a private repository, `imgpkg` should try to check if the bundle was copied to the current location or not and should use that information to try to read the needed information.

Relates to #146

### DIFF
--- a/pkg/imgpkg/bundle/bundle_images_lock.go
+++ b/pkg/imgpkg/bundle/bundle_images_lock.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"sync"
 
+	regname "github.com/google/go-containerregistry/pkg/name"
 	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
@@ -33,12 +34,22 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 		return nil, err
 	}
 
-	allImagesLock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo())
+	resultImagesLock := NewImagesLock(lockconfig.ImagesLock{}, o.imgRetriever, o.Repo())
 
 	errChan := make(chan error, len(imagesLock.Images))
 	mutex := &sync.Mutex{}
 
-	for _, image := range imagesLock.Images {
+	bImagesLock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo())
+	// We generate the locations at this point.
+	// This is done to ensure that the first place we look for each image is in
+	// the bundle repository we are currently processing, only after will try to
+	// check the original location
+	err = bImagesLock.GenerateImagesLocations()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, image := range bImagesLock.ImageRefs() {
 		if skip := processedImgs.CheckAndAddImage(image.Image); skip {
 			errChan <- nil
 			continue
@@ -46,15 +57,17 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 
 		image := image.DeepCopy()
 		go func() {
-			imgsLock, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs)
+			imgsLock, imgRef, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs)
 			if err != nil {
 				errChan <- err
 				return
 			}
+
+			mutex.Lock()
+			defer mutex.Unlock()
+			resultImagesLock.AddImageRef(imgRef)
 			if imgsLock != nil {
-				mutex.Lock()
-				defer mutex.Unlock()
-				err = allImagesLock.Merge(imgsLock)
+				err = resultImagesLock.Merge(imgsLock)
 				if err != nil {
 					errChan <- fmt.Errorf("Merging images for bundle '%s': %s", image.Image, err)
 					return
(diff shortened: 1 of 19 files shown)


pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -33,28 +33,31 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 		return nil, err
 	}
 
-	allImagesLock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo())
+	resultImagesLock := NewImagesLock(lockconfig.ImagesLock{}, o.imgRetriever, o.Repo())
+	bImagesLock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo())
```
cppforlife: bImagesLock -> currentImagesLock?

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -33,28 +33,31 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
...
 		image := image.DeepCopy()
 		go func() {
-			imgsLock, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs)
+			imgsLock, imgRef, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs)
 			if err != nil {
 				errChan <- err
 				return
 			}
+
+			resultImagesLock.AddImageRef(imgRef)
```
cppforlife, line 58: this needs to be done within a lock as well.

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -6,18 +6,17 @@ package bundle
...
 	ctlimg "github.com/k14s/imgpkg/pkg/imgpkg/image"
 	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
-	"github.com/k14s/imgpkg/pkg/imgpkg/util"
 )
 
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 github.com/cppforlife/go-cli-ui/ui.UI
 
 func NewImagesLock(imagesLock lockconfig.ImagesLock, imgRetriever ctlimg.ImagesMetadata, relativeToRepo string) *ImagesLock {
-	return &ImagesLock{imagesLock: imagesLock, imgRetriever: imgRetriever, relativeToRepo: relativeToRepo}
+	imgsLock := &ImagesLock{imagesLock: imagesLock, imgRetriever: imgRetriever, relativeToRepo: relativeToRepo}
```
cppforlife: it seems that if we are generating images locations in the constructor we actually dont need to save off relativeToRepo. can just pass into generateImagesLocations().

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -87,67 +82,13 @@ func (o *ImagesLock) LocalizeImagesLock() (lockconfig.ImagesLock, bool, error) {
...
-	return "", fmt.Errorf("Checking image existence: %s", err)
+	return imagesLock, skippedLocalization, nil
 }
 
-func (o *ImagesLock) imageRelativeToBundle(img string) (string, error) {
+func (o *ImagesLock) imageRelativeToBundle(img string) string {
 	imgParts := strings.Split(img, "@")
 	if len(imgParts) != 2 {
-		return "", fmt.Errorf("Parsing image URL: %s", img)
+		panic(fmt.Sprintf("The provided image URL '%s' does not contain a digest", img))
```
cppforlife: throwing in a prefix `Internal inconsistency: ...` will tell user that this is an internal error and not somethign they can fix (aka bug)

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -26,6 +25,13 @@ type ImagesLock struct {
 	relativeToRepo string
 }
 
+func (o *ImagesLock) generateImagesLocations() {
+	for i, imgRef := range o.ImageRefs() {
```
cppforlife: lets do `range o.imagesLock.Images` so that it's clear it's related to what we are mutating two lines down (`o.imagesLock.Images[i]...`).

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -39,45 +45,34 @@ func (o *ImagesLock) Merge(imgLock *ImagesLock) error {
...
 		// We assume that the bundle was not copied to the bundle location,
 		// so there we cannot localize any image
-		if foundImg != imageInBundleRepo {
-			return o.imagesLock, true, nil
+		if foundImg != imgRef.PrimaryLocation() {
+			imageRefs = []lockconfig.ImageRef{}
+			// Remove the bundle location on all the Images
+			for _, image := range o.imagesLock.Images {
+				imageRefs = append(imageRefs, image.DiscardLocationsExcept(image.Image))
+			}
```
cppforlife: nit: might be more readable if done outside of the loop below e.g.

```
if skippedLocalization {
   ...
}
```

just my 2c, not a blocking change.

pkg/imgpkg/bundle/images_lock.go

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: done

pkg/imgpkg/bundle/images_lock.go

pkg/imgpkg/bundle/images_lock.go

pkg/imgpkg/bundle/bundle_images_lock.go

Issue #131, opened 2021-04-19
Imgpkg fails to pull a bundle if a registry does not support the HEAD call.


ewrenn8:
**What steps did you take:**

We have discussed the steps to reproduce internally. Essentially, if a repo does not support the HEAD call on the manifests endpoint, imgpkg will error when checking for image existence, even if the image does exist.

**What happened:**
Bundle failed to pull

**What did you expect:**
Bundle to successfully be pulled since references image does exist and is publicly available.

**Anything else you would like to add:**

There were conversations around adding some fallback behavior which will issue a GET request in the case HEAD fails. @DennisDenuto pointed out crane did something similar in [this pr](https://github.com/google/go-containerregistry/pull/971)

DennisDenuto:
wrt the 'fallback' behavior that crane uses, after some more thought there are concerns around performance and hitting rate limiting with registries such as dockerhub. (Dockerhub rate-limiting counts `GET` requests but not `HEAD`)

`imgpkg` uses `HEAD` `/manifest` to determine if a bundle has been relocated (imgpkg will short-circuit the loop upon encountering the first image in the bundle that doesn't exist, so I don't think we will run into any perf issues here), *and* to verify that every image in the bundle has been copied correctly (this is a concern).

We believe it won't be uncommon for bundles to contain 1000's of images in the future, hence the performance  and rate-limiting concern.

I think we should attempt at 'fixing' the registries that are not conforming to the [dist spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints), but assuming we do that, this also leaves the question as to how imgpkg behaves with 'older' registries that don't conform.

Here is a straw-person idea:
What if we use `GET` requests when targeting a registry that doesn't play nice with `HEAD`, and for every other type of registry we use `HEAD`? (fyi I don't think there is a programmatic way to determine the provider or version of a registry. Perhaps a feature toggle 
 or new flag might be needed) 

Note: 
As a data-point, when `HEAD` in the code is changed to `GET`, the performance test for simulating a large "cf-for-k8s" type bundle goes from `55 seconds to 1 min 10 seconds`

cppforlife:
> I don't think there is a programmatic way to determine the provider or version of a registry. Perhaps a feature toggle or new flag might be needed

[...]

PR #149, opened 2021-05-14
Fallback to Get when Head Fails

joaopapereira:
Related to #131


pkg/imgpkg/registry/registry.go
```diff
@@ -97,7 +97,11 @@ func (r Registry) Digest(ref regname.Reference) (regv1.Hash, error) {
 	}
 	desc, err := regremote.Head(overriddenRef, r.opts...)
 	if err != nil {
-		return regv1.Hash{}, err
+		getDesc, err := r.Get(overriddenRef)
```
cppforlife: may be use `regremote.Get` here given ambiguity of r.Get vs r.Generic?

Issue #147, opened 2021-05-10
Copy Cosign Signature with The Rest of the Images


iamhsk:
**Describe the problem/challenge you have**
Currently we don't have a clear path to copy signatures with images. 

**Describe the solution you'd like**
Test out copying signatures using cosign

**Anything else you would like to add:**
N/A

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

cppforlife:
available in https://github.com/vmware-tanzu/carvel-imgpkg/releases/tag/v0.9.0 via --cosign-signatures flag on copy cmd.

PR #151, opened 2021-05-19
Enable transport of cosign signatures when copying

joaopapereira:
Find for each image being copied if they have a cosign signature and if so, copy it to the destination

Related to #147


pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return regname.Tag{}, fmt.Errorf("calculating the hash: %s", err)
+	}
+
+	tag, err := regname.NewTag(reference.Repository.Name())
+	if err != nil {
+		return regname.Tag{}, fmt.Errorf("calculating the tag: %s", err)
+	}
+
+	return tag.Tag(fmt.Sprintf("%s-%s.sig", digest.Algorithm, digest.Hex)), nil
```
cppforlife: i take it this (`.sig`) isnt right? we should actually use cosign binary in our e2e tests to make sure that sign, copy, verify actually works.

pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,64 @@
...
+func (c Cosign) signatureTag(reference regname.Digest) (regname.Tag, error) {
+	digest, err := v1.NewHash(reference.DigestStr())
+	if err != nil {
+		return regname.Tag{}, fmt.Errorf("calculating the hash: %s", err)
+	}
+
+	tag, err := regname.NewTag(reference.Repository.Name())
+	if err != nil {
+		return regname.Tag{}, fmt.Errorf("calculating the tag: %s", err)
+	}
```
cppforlife: this seems like a lot of work for what https://github.com/sigstore/cosign/blob/76fde761a4e016c57293113ee1362f1c24a647cb/pkg/cosign/fetch.go#L51-L56 is doing. we should either copy the code and just construct our own v1.Descriptor with a digest and use Munge.

pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,64 @@
...
+func (c Cosign) Signature(reference regname.Digest) (imageset.UnprocessedImageRef, error) {
+	tag, err := c.signatureTag(reference)
+	if err != nil {
+		return imageset.UnprocessedImageRef{}, err
+	}
+
+	signature, err := c.registry.Digest(tag)
+	if err != nil {
+		if transportErr, ok := err.(*transport.Error); ok {
+			if transportErr.StatusCode == http.StatusNotFound || transportErr.StatusCode == http.StatusForbidden || transportErr.StatusCode == http.StatusUnauthorized {
```
cppforlife: i think we should only be checking if status is 404.

pkg/imgpkg/signature/retrieve_signatures.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func NewRetriever(finder Finder, concurrency int) *Retriever {
+	return &Retriever{
+		signatureFinder: finder,
+		concurrency:     concurrency,
+	}
+}
+
+func (s *Retriever) Signatures(images *imageset.UnprocessedImageRefs) (*imageset.UnprocessedImageRefs, error) {
```
cppforlife: this isnt happening in parallel. im also thinking we may want to introduce --cosign-signatures (bool; default=false) to the copy operation for now since most folks arent using signatures today.

pkg/imgpkg/signature/retrieve_signatures.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, ref := range images.All() {
+		imgDigest, err := name.NewDigest(ref.DigestRef)
+		if err != nil {
+			return nil, fmt.Errorf("signatures for: %s: %s", ref.DigestRef, err)
+		}
+
+		signature, err := s.signatureFinder.Signature(imgDigest)
+		if err != nil {
+			if _, ok := err.(NotFound); !ok {
+				return nil, fmt.Errorf("unable to get signature for image '%s': %s", imgDigest.Name(), err)
```
cppforlife: nit: we have been doing error messages with first letter capitalized across all carvel tools.

pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,64 @@
...
+				return imageset.UnprocessedImageRef{}, NotFound{}
+			}
+		}
+		return imageset.UnprocessedImageRef{}, err
+	}
+
+	signDigest, err := regname.NewDigest(reference.Name())
```
cppforlife: reference is already a regname.Digest, dont think we need to run through this again.

pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+
+	regname "github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+	"github.com/k14s/imgpkg/pkg/imgpkg/imageset"
+	"github.com/k14s/imgpkg/pkg/imgpkg/registry"
+)
+
+func NewCosignFinder(reg registry.Registry) *Cosign {
```
cppforlife: NewCosignFinder -> NewCosign

pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,64 @@
...
+	return &Cosign{registry: reg}
+}
+
+type Cosign struct {
+	registry registry.Registry
+}
+
+func (c Cosign) Signature(reference regname.Digest) (imageset.UnprocessedImageRef, error) {
+	tag, err := c.signatureTag(reference)
```
cppforlife: var name is confusing. it's a "tag" it's a tagRef?

pkg/imgpkg/signature/cosign.go
joaopapereira: done

pkg/imgpkg/signature/cosign.go
joaopapereira: Now using cosign to sign the images in e2e tests

pkg/imgpkg/signature/cosign.go
joaopapereira: decided to copy the code because the amount of libraries that cosign was adding to our code did not made sense since we are just using 1 two-line function from their code

pkg/imgpkg/signature/retrieve_signatures.go
joaopapereira: done

pkg/imgpkg/signature/cosign.go
joaopapereira: done

pkg/imgpkg/signature/cosign.go
joaopapereira: done

pkg/imgpkg/signature/cosign.go
joaopapereira: done

pkg/imgpkg/signature/retrieve_signatures.go
joaopapereira: done

.gitignore
```diff
@@ -3,5 +3,6 @@
 /imgpkg-linux-amd64
 /imgpkg-windows-amd64.exe
 /tmp
+bin/
```
cppforlife: should we use tmp/bin for this?

pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,52 @@
...
+	registry registry.Registry
+}
+
+func (c Cosign) Signature(imageRef regname.Digest) (imageset.UnprocessedImageRef, error) {
+	sigTagRef, err := c.signatureTag(imageRef)
+	if err != nil {
+		return imageset.UnprocessedImageRef{}, err
+	}
+
+	signature, err := c.registry.Digest(sigTagRef)
```
cppforlife: may be better: signature -> sigDigestRef?

pkg/imgpkg/signature/cosign.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		if transportErr, ok := err.(*transport.Error); ok {
+			if transportErr.StatusCode == http.StatusNotFound {
+				return imageset.UnprocessedImageRef{}, NotFound{}
+			}
+		}
+		return imageset.UnprocessedImageRef{}, err
+	}
+
+	signDigest := imageRef.Digest(signature.String())
```
cppforlife: i dont think we need this line. sigTagRef is already relative to imageRef and above registry.Digest call will get the digest ref.

joaopapereira: We use this call to see if the image exist or not, maybe we should have a function that wrap the Digest call

pkg/imgpkg/signature/cosign.go
joaopapereira: done

.gitignore
joaopapereira: done

PR #152, opened 2021-05-25
Generate Locations OCI Image

joaopapereira:
A couple of things had to change in order for this change to be done:
- Change the way the progress bar is being used, please refer to the first commit message for more information
- Add Logger with the level set, this will allow us to display messages with different level, not essential for this work, but helpful for the future
- Work to create the Locations OCI Image

What is missing:
- [x] Add more test coverage
- [x] Check if we can move the locations code into the Bundle struct
- [x] Check if it makes sense to refactor the singleLayerReader, we duplicated it to start
- [ ] General refactoring
- [x] The UI is being passed around, not sure if we want that...

### DIFF
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -29,6 +29,7 @@ type Bundle struct {
 	plainImg         *plainimg.PlainImage
 	imgRetriever     ctlimg.ImagesMetadata
 	imagesLockReader ImagesLockReader
+	imagesRef        map[string]ImageRef
 }
 
 func NewBundle(ref string, imagesMetadata ctlimg.ImagesMetadata) *Bundle {
@@ -36,16 +37,35 @@ func NewBundle(ref string, imagesMetadata ctlimg.ImagesMetadata) *Bundle {
 }
 
 func NewBundleFromPlainImage(plainImg *plainimg.PlainImage, imagesMetadata ctlimg.ImagesMetadata) *Bundle {
-	return &Bundle{plainImg, imagesMetadata, &singleLayerReader{}}
+	return &Bundle{plainImg: plainImg, imgRetriever: imagesMetadata, imagesLockReader: &singleLayerReader{}, imagesRef: map[string]ImageRef{}}
 }
 
 func NewBundleWithReader(ref string, imagesMetadata ctlimg.ImagesMetadata, imagesLockReader ImagesLockReader) *Bundle {
-	return &Bundle{plainimg.NewPlainImage(ref, imagesMetadata), imagesMetadata, imagesLockReader}
+	return &Bundle{plainImg: plainimg.NewPlainImage(ref, imagesMetadata), imgRetriever: imagesMetadata, imagesLockReader: imagesLockReader, imagesRef: map[string]ImageRef{}}
 }
 
 func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
 func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
+
+func (o *Bundle) AddImagesRef(refs ...ImageRef) {
+	for _, imageRef := range refs {
+		o.imagesRef[imageRef.Image] = imageRef
+	}
+}
+
+func (o *Bundle) ImageRef(imageDigest string) (ImageRef, bool) {
+	ref, found := o.imagesRef[imageDigest]
+	return ref, found
+}
+
+func (o *Bundle) ImagesRef() []ImageRef {
+	var imgsRef []ImageRef
+	for _, ref := range o.imagesRef {
+		imgsRef = append(imgsRef, ref)
+	}
+	return imgsRef
+}
 
 func (o *Bundle) Pull(outputPath string, ui goui.UI, pullNestedBundles bool) error {
 	return o.pull(outputPath, ui, pullNestedBundles, "", map[string]bool{}, 0)
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -50,18 +50,18 @@ func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
 func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
-func (o *Bundle) AddImageRefs(refs ...ImageRef) {
+func (o *Bundle) addImageRefs(refs ...ImageRef) {
 	for _, imageRef := range refs {
 		o.imagesRef[imageRef.Image] = imageRef
 	}
 }
 
-func (o *Bundle) ImageRef(imageDigest string) (ImageRef, bool) {
+func (o *Bundle) imageRef(imageDigest string) (ImageRef, bool) {
 	ref, found := o.imagesRef[imageDigest]
 	return ref, found
 }
 
-func (o *Bundle) ImageRefs() []ImageRef {
+func (o *Bundle) imageRefs() []ImageRef {
 	var imgsRef []ImageRef
 	for _, ref := range o.imagesRef {
 		imgsRef = append(imgsRef, ref)
@@ -76,7 +76,7 @@ func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesM
 	}
 	var bundleProcessedImage imageset.ProcessedImage
 	for _, image := range processedImages.All() {
-		ref, found := o.ImageRef(image.UnprocessedImageRef.DigestRef)
+		ref, found := o.imageRef(image.UnprocessedImageRef.DigestRef)
 		if found {
 			locationsCfg.Images = append(locationsCfg.Images, ImageLocation{
 				Image:    ref.Image,
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -72,7 +72,7 @@ func (o *Bundle) imageRefs() []ImageRef {
 func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesMetadataWriter, logger util.LoggerWithLevels) error {
 	locationsCfg := ImageLocationsConfig{
 		APIVersion: LocationAPIVersion,
-		Kind:       LocationKind,
+		Kind:       ImageLocationsKind,
 	}
 	var bundleProcessedImage imageset.ProcessedImage
 	for _, image := range processedImages.All() {
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -50,7 +50,7 @@ func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
 func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
-func (o *Bundle) AddImagesRef(refs ...ImageRef) {
+func (o *Bundle) AddImageRefs(refs ...ImageRef) {
 	for _, imageRef := range refs {
 		o.imagesRef[imageRef.Image] = imageRef
 	}
@@ -61,7 +61,7 @@ func (o *Bundle) ImageRef(imageDigest string) (ImageRef, bool) {
 	return ref, found
 }
 
-func (o *Bundle) ImagesRef() []ImageRef {
+func (o *Bundle) ImageRefs() []ImageRef {
 	var imgsRef []ImageRef
 	for _, ref := range o.imagesRef {
 		imgsRef = append(imgsRef, ref)
@@ -85,7 +85,6 @@ func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesM
 		}
 		if image.UnprocessedImageRef.DigestRef == o.DigestRef() {
 			bundleProcessedImage = image
-			break
 		}
 	}
 
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -80,7 +80,7 @@ func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesM
 		if found {
 			locationsCfg.Images = append(locationsCfg.Images, ImageLocation{
 				Image:    ref.Image,
-				IsBundle: ref.IsBundle,
+				IsBundle: *ref.IsBundle,
 			})
 		}
 		if image.UnprocessedImageRef.DigestRef == o.DigestRef() {
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -9,11 +9,13 @@ import (
 	"strings"
 
 	goui "github.com/cppforlife/go-cli-ui/ui"
-	"github.com/google/go-containerregistry/pkg/name"
+	regname "github.com/google/go-containerregistry/pkg/name"
 	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	ctlimg "github.com/k14s/imgpkg/pkg/imgpkg/image"
+	"github.com/k14s/imgpkg/pkg/imgpkg/imageset"
 	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
 	plainimg "github.com/k14s/imgpkg/pkg/imgpkg/plainimage"
+	"github.com/k14s/imgpkg/pkg/imgpkg/util"
 )
 
 const (
@@ -65,6 +67,40 @@ func (o *Bundle) ImagesRef() []ImageRef {
 		imgsRef = append(imgsRef, ref)
 	}
 	return imgsRef
+}
+
+func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesMetadataWriter, logger util.LoggerWithLevels) error {
+	locationsCfg := ImageLocationsConfig{
+		APIVersion: LocationAPIVersion,
+		Kind:       LocationKind,
+	}
+	var bundleProcessedImage imageset.ProcessedImage
+	for _, image := range processedImages.All() {
+		ref, found := o.ImageRef(image.UnprocessedImageRef.DigestRef)
+		if found {
+			locationsCfg.Images = append(locationsCfg.Images, ImageLocation{
+				Image:    ref.Image,
+				IsBundle: ref.IsBundle,
+			})
+		}
+		if image.UnprocessedImageRef.DigestRef == o.DigestRef() {
+			bundleProcessedImage = image
+			break
+		}
+	}
+
+	destinationRef, err := regname.NewDigest(bundleProcessedImage.DigestRef)
+	if err != nil {
+		panic(fmt.Sprintf("Internal inconsistency: '%s' have to be a digest", bundleProcessedImage.DigestRef))
+	}
+
+	logger.Debugf("creating Locations OCI Image\n")
+	// Using NewNoopUI because we do not want to have output from this push
+	err = NewLocations(logger).Save(reg, destinationRef, locationsCfg, goui.NewNoopUI())
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 func (o *Bundle) Pull(outputPath string, ui goui.UI, pullNestedBundles bool) error {
@@ -125,7 +161,7 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 			if o.shouldPrintNestedBundlesHeader(bundlePath, numSubBundles) {
 				ui.BeginLinef("\nNested bundles\n")
 			}
-			bundleDigest, err := name.NewDigest(image.Image)
+			bundleDigest, err := regname.NewDigest(image.Image)
 			if err != nil {
 				return err
 			}
@@ -156,7 +192,7 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 	return nil
 }
 
-func (*Bundle) subBundlePath(bundleDigest name.Digest) string {
+func (*Bundle) subBundlePath(bundleDigest regname.Digest) string {
 	return filepath.Join(ImgpkgDir, BundlesDir, strings.ReplaceAll(bundleDigest.DigestStr(), "sha256:", "sha256-"))
 }
 
--- a/pkg/imgpkg/bundle/bundle_images_lock.go
+++ b/pkg/imgpkg/bundle/bundle_images_lock.go
@@ -55,7 +55,7 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 
 			mutex.Lock()
 			defer mutex.Unlock()
-			resultImagesLock.AddImageRef(imgRef)
+			resultImagesLock.AddImageRef(imgRef, imgsLock != nil) // imgsLock will be != nil when the image is a bundle
 			if imgsLock != nil {
 				err = resultImagesLock.Merge(imgsLock)
 				if err != nil {
--- a/pkg/imgpkg/bundle/bundle_images_lock.go
+++ b/pkg/imgpkg/bundle/bundle_images_lock.go
@@ -9,37 +9,62 @@ import (
 	"io"
 	"io/ioutil"
 	"path/filepath"
+	"strings"
 	"sync"
 
+	regname "github.com/google/go-containerregistry/pkg/name"
 	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
 	"github.com/k14s/imgpkg/pkg/imgpkg/util"
 )
 
-func (o *Bundle) AllImagesLock(concurrency int) (*ImagesLock, error) {
+func (o *Bundle) AllImagesRefs(concurrency int, logger util.LoggerWithLevels) ([]*Bundle, ImagesRef, error) {
 	throttleReq := util.NewThrottle(concurrency)
-	return o.buildAllImagesLock(&throttleReq, &processedImages{processedImgs: map[string]struct{}{}})
-}
-
-func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *processedImages) (*ImagesLock, error) {
-	img, err := o.checkedImage()
+	bundles, imagesLock, err := o.buildAllImagesLock(&throttleReq, &processedImages{processedImgs: map[string]struct{}{}}, logger)
 	if err != nil {
-		return nil, err
+		return nil, ImagesRef{}, err
 	}
 
-	imagesLock, err := o.imagesLockReader.Read(img)
+	imagesRef := imagesLock.ImageRefs()
+	// Ensure that the correct IsBundle flag is provided.
+	// This loop needs to happen because we skipped some images for some bundle, and only at this point we have
+	// the full list of ImagesRef created and can fill the gaps inside each bundle
+	for _, bundle := range bundles {
+		for _, ref := range bundle.ImagesRef() {
+			imgRef, found := imagesRef.ImageRef(ref.Image)
+			if !found {
+				panic(fmt.Sprintf("Internal inconsistency: The Image '%s' cannot be found in the total list of images", ref.Image))
+			}
+
+			bundle.AddImagesRef(imgRef)
+		}
+	}
+
+	return bundles, imagesRef, err
+}
+
+func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *processedImages, logger util.LoggerWithLevels) ([]*Bundle, *ImagesLock, error) {
+	img, err := o.checkedImage()
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	resultImagesLock := NewImagesLock(lockconfig.ImagesLock{}, o.imgRetriever, o.Repo())
-	currentImagesLock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo())
 
-	errChan := make(chan error, len(imagesLock.Images))
+	imageRefs, err := o.fetchImagesRef(img, resultImagesLock, logger)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	bundles := []*Bundle{o}
+
+	errChan := make(chan error, len(imageRefs))
 	mutex := &sync.Mutex{}
 
-	for _, image := range currentImagesLock.ImageRefs() {
+	for _, image := range imageRefs {
+		o.AddImagesRef(image)
+
 		if skip := processedImgs.CheckAndAddImage(image.Image); skip {
 			errChan <- nil
 			continue
@@ -47,7 +72,7 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 
 		image := image.DeepCopy()
 		go func() {
-			imgsLock, imgRef, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs)
+			nestedBundles, imgsLock, imgRef, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs, logger)
 			if err != nil {
 				errChan <- err
 				return
@@ -55,7 +80,12 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 
 			mutex.Lock()
 			defer mutex.Unlock()
-			resultImagesLock.AddImageRef(imgRef, imgsLock != nil) // imgsLock will be != nil when the image is a bundle
+			if nestedBundles != nil {
+				bundles = append(bundles, nestedBundles...)
+			}
+
+			// Adds Image to the resulting ImagesLock
+			resultImagesLock.AddImageRef(imgRef, nestedBundles != nil) // nestedBundles will be != nil when the image is a bundle
 			if imgsLock != nil {
 				err = resultImagesLock.Merge(imgsLock)
 				if err != nil {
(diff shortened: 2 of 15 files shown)


pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,65 @@ import (
...
 	}
 
 	return nil
 }
 
-func (o *ImagesLock) AddImageRef(ref lockconfig.ImageRef) {
-	o.imagesLock.AddImageRef(ref)
+func (o *ImagesLock) AddImageRef(ref lockconfig.ImageRef, bundle bool) {
+	for _, image := range o.imagesRef {
+		if image.Image == ref.Image {
```
cppforlife: may have to override in case when isbundle==false before, and new bundle is true.

pkg/imgpkg/registry/with_progress.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (w WithProgress) Image(reference regname.Reference) (regv1.Image, error) {
+	return w.delegate.Image(reference)
+}
+
+func (w WithProgress) FirstImageExists(digests []string) (string, error) {
+	return w.delegate.FirstImageExists(digests)
+}
+
+func (w *WithProgress) MultiWrite(imageOrIndexesToUpload map[regname.Reference]remote.Taggable, concurrency int) error {
```
cppforlife: instead of accessing private opts, lets change sig of MultiWrite to include optional remote options.

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 func (o *Bundle) AllImagesLock(concurrency int) (*ImagesLock, error) {
+	// TODO: See if we can add the locations here
```
cppforlife: per our discussion, this method should probably start returning Bundles that were found. we can still keep returning ImagesLock but we should probably rename it to ImageRefs since that's what it is at this point.

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -77,17 +84,63 @@ func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error)
 		return nil, fmt.Errorf("Building import repository ref: %s", err)
 	}
 
+	c.logger.Debugf("copy the fetched images\n")
 	processedImages, ids, err := c.imageSet.Relocate(unprocessedImageRefs, importRepo, c.registry)
 	if err != nil {
 		return nil, err
 	}
 
+	if c.originBundle != nil {
```
cppforlife: per our discussion, consolidating this functionality on the Bundle would nicely keep consumers of bundles oblivious to image locations.

pkg/imgpkg/registry/with_progress.go
```diff
@@ -0,0 +1,60 @@
...
+func (w WithProgress) FirstImageExists(digests []string) (string, error) {
+	return w.delegate.FirstImageExists(digests)
+}
+
+func (w *WithProgress) MultiWrite(imageOrIndexesToUpload map[regname.Reference]remote.Taggable, concurrency int, opts ...remote.Option) error {
+	uploadProgress := make(chan regv1.Update)
+	w.logger.Start(uploadProgress)
+	defer w.logger.End()
+
+	return w.delegate.MultiWrite(imageOrIndexesToUpload, concurrency, append(opts, remote.WithProgress(uploadProgress))...)
```
cppforlife: i dunno if opts is being used at the call site, but since append is possibly mutating i would recommend `append(append([]remote.Option{}, opts...),  remote.WithProgress(uploadProgress))` to make sure we are copying it.

pkg/imgpkg/registry/registry.go
```diff
@@ -107,9 +107,10 @@ func (r Registry) Image(ref regname.Reference) (regv1.Image, error) {
 	return regremote.Image(overriddenRef, r.opts...)
 }
 
-func (r Registry) MultiWrite(imageOrIndexesToUpload map[regname.Reference]regremote.Taggable, concurrency int) error {
+func (r Registry) MultiWrite(imageOrIndexesToUpload map[regname.Reference]regremote.Taggable, concurrency int, opts ...regremote.Option) error {
 	return util.Retry(func() error {
-		return regremote.MultiWrite(imageOrIndexesToUpload, append(r.opts, regremote.WithJobs(concurrency))...)
+		lOpts := append(r.opts, opts...)
```
cppforlife: same as comment below about mutating opts

test/helpers/logger.go
```diff
@@ -25,6 +25,10 @@ func (l Logger) Section(msg string, f func()) {
 	f()
 }
 
+func (l Logger) Warnf(msg string, args ...interface{}) {
```
cppforlife: do we need warnings in tests?

pkg/imgpkg/cmd/copy.go
```diff
@@ -124,14 +123,15 @@ func (c *CopyOptions) Run() error {
 			signatureRetriever = signature.NewNoop()
 		}
 
+		levelLogger.LogLevel = util.LogTrace
```
cppforlife: no longer needed?

pkg/imgpkg/cmd/copy.go
```diff
@@ -327,8 +325,8 @@ func informUserToUseTheNonDistributableFlagWithDescriptors(logger Logger, includ
 	}
 
 	if includeNonDistributableFlag && noNonDistributableLayers {
-		logger.WriteStr("Warning: '--include-non-distributable-layers' flag provided, but no images contained a non-distributable layer.")
+		logger.Warnf("Warning: '--include-non-distributable-layers' flag provided, but no images contained a non-distributable layer.")
```
cppforlife: should Warnf add `Warning: `?

pkg/imgpkg/bundle/locations_config.go
```diff
@@ -0,0 +1,72 @@
...
+type LocationConfig struct {
+	APIVersion string          `json:"apiVersion"` // This generated yaml, but due to lib we need to use `json`
+	Kind       string          `json:"kind"`       // This generated yaml, but due to lib we need to use `json`
+	Images     []ImageLocation `json:"images"`     // This generated yaml, but due to lib we need to use `json`
+}
+
+type ImageLocation struct {
+	Image string `json:"image"` // This generated yaml, but due to lib we need to use `json`
+	// Repository will not be used right now because all images will be in the bundle repository
+	Repository string `json:"repository,omitempty"` // This generated yaml, but due to lib we need to use `json`
```
cppforlife: should we remove Repository for now?

pkg/imgpkg/bundle/locations_config.go
```diff
@@ -0,0 +1,72 @@
...
+	Bundle     bool   `json:"bundle"`               // This generated yaml, but due to lib we need to use `json`
```
cppforlife: wdyt about calling this `IsBundle  bool  json:"isBundle"`. imho is makes it more understandable that this is a t/f kinda a thing vs some bundle configuration.

pkg/imgpkg/bundle/locations_config.go
```diff
@@ -0,0 +1,72 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	LocationFilepath   = "images-locations.yml"
+	LocationKind       = "Locations"
+	LocationAPIVersion = "imgpkg.carvel.dev/v1alpha1"
+)
+
+type LocationConfig struct {
```
cppforlife: LocationConfig -> ImageLocationsConfig

pkg/imgpkg/bundle/locations_config.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return NewLocationConfigFromBytes(bs)
+}
+
+func NewLocationConfigFromBytes(data []byte) (LocationConfig, error) {
+	var lock LocationConfig
+
+	err := yaml.UnmarshalStrict(data, &lock)
+	if err != nil {
+		return lock, fmt.Errorf("Unmarshaling bundle lock: %s", err)
```
cppforlife: not a bundle lock

pkg/imgpkg/bundle/locations_config.go
```diff
@@ -0,0 +1,72 @@
...
+		return LocationConfig{}, fmt.Errorf("Reading path %s: %s", path, err)
+	}
+
+	return NewLocationConfigFromBytes(bs)
+}
+
+func NewLocationConfigFromBytes(data []byte) (LocationConfig, error) {
+	var lock LocationConfig
+
+	err := yaml.UnmarshalStrict(data, &lock)
```
cppforlife, line 42: hmm, whenever we add new fields, is "strict" going to screw us up?

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,175 @@
...
+
+func NewLocations(logger util.LoggerWithLevels) *Locations {
+	return NewLocationsWithReader(&locationsSingleLayerReader{}, logger)
+}
+
+func NewLocationsWithReader(reader LocationImageReader, logger util.LoggerWithLevels) *Locations {
+	return &Locations{reader: reader, logger: logger}
+}
+
+func (r Locations) Fetch(registry image.ImagesMetadata, bundleRef name.Digest) (LocationConfig, error) {
```
cppforlife: should we move registry into a constructor?

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,175 @@
...
+	r.logger.Tracef("pushing image\n")
+	_, err = plainimage.NewContents([]string{tmpDir}, nil).Push(tag, nil, reg, ui)
+	if err != nil {
+		return fmt.Errorf("pushing locations image to '%s': %s", tag.Name(), err)
+	}
+
+	return nil
+}
+
+func (r Locations) tagFromBundleRef(bundleRef name.Digest) (name.Tag, error) {
```
cppforlife: tagFromBundleRef -> locationsRefFromBundleRef

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,175 @@
...
+	return NewLocationsWithReader(&locationsSingleLayerReader{}, logger)
+}
+
+func NewLocationsWithReader(reader LocationImageReader, logger util.LoggerWithLevels) *Locations {
+	return &Locations{reader: reader, logger: logger}
+}
+
+func (r Locations) Fetch(registry image.ImagesMetadata, bundleRef name.Digest) (LocationConfig, error) {
+	r.logger.Tracef("fetching Locations OCI Images for bundle: %s\n", bundleRef)
+	locTag, err := r.tagFromBundleRef(bundleRef)
```
cppforlife: locTag -> locRef (tag is more of an impl detail, but it's definitely a ref)

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		if terr, ok := err.(*transport.Error); ok {
+			if terr.StatusCode == http.StatusNotFound {
+				return LocationConfig{}, &LocationsNotFound{image: locTag.Name()}
+			}
+		}
+		return LocationConfig{}, fmt.Errorf("fetching location image: %s", err)
+	}
+
+	r.logger.Tracef("reading the Locations configuration file\n")
```
cppforlife, line 72: should Tracef require to add newline everywhere?

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,175 @@
...
+	tag, err := r.tagFromBundleRef(bundleRef)
+	if err != nil {
+		return fmt.Errorf("calculating locations image tag: %s", err)
+	}
+
+	tmpDir, err := os.MkdirTemp("", "locations")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpDir)
```
cppforlife: im pretty sure it's erroring because directory not empty. os.RemoveAll?

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (r Locations) Save(reg ImagesMetadataWriter, bundleRef name.Digest, config LocationConfig, ui ui.UI) error {
+	r.logger.Tracef("saving Locations OCI Image for bundle: %s\n", bundleRef.Name())
+
+	tag, err := r.tagFromBundleRef(bundleRef)
+	if err != nil {
+		return fmt.Errorf("calculating locations image tag: %s", err)
+	}
+
+	tmpDir, err := os.MkdirTemp("", "locations")
```
cppforlife: lets name it imgpkg-bundle-locations to make it easier for folks to know who creates temp dirs if some get left around.

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,175 @@
...
+		return LocationConfig{}, fmt.Errorf("reading fetched location image: %s", err)
+	}
+
+	return cfg, err
+}
+
+func (r Locations) Save(reg ImagesMetadataWriter, bundleRef name.Digest, config LocationConfig, ui ui.UI) error {
+	r.logger.Tracef("saving Locations OCI Image for bundle: %s\n", bundleRef.Name())
+
+	tag, err := r.tagFromBundleRef(bundleRef)
```
cppforlife: tag -> locRef

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,94 @@ import (
 
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 github.com/cppforlife/go-cli-ui/ui.UI
 
+type ImageRef struct {
+	lockconfig.ImageRef
+	IsBundle bool
+}
+type ImagesRef []ImageRef
```
cppforlife: ImagesRef -> ImageRefs (plural of ImageRef, a single item in a collection). in other words, it's not a single Ref to bunch of images, it's multiple ImageRefs.

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,94 @@ import (
...
+		for j, imageRef := range result {
+			if imageRef.Image == ref.Image {
+				found = true
+				result[j] = ref
+				break
+			}
+		}
+
+		if !found {
+			result = append(result, ref)
```
cppforlife: this is getting a bit funky with reassigning i pointer. to make it a bit less fragile, lets switch to

```
type ImageRefs struct {
  refs []ImageRef
}
```

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,94 @@ import (
...
+		}
+	}
+	*i = result
+}
+func (i *ImagesRef) ImageRef(ref string) (ImageRef, bool) {
```
cppforlife: ImageRef -> Find

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,94 @@ import (
...
+	for i, imgRef := range o.imagesRef {
 		imageInBundleRepo := o.imageRelativeToBundle(imgRef.Image, relativeToRepo)
-		o.imagesLock.Images[i].AddLocation(imageInBundleRepo)
+		o.imagesRef[i].AddLocation(imageInBundleRepo)
 	}
 }
 
-func (o ImagesLock) ImageRefs() []lockconfig.ImageRef {
-	return o.imagesLock.Images
+func (o ImagesLock) ImagesRefs() ImagesRef {
```
cppforlife: ImagesRefs -> ImageRefs per pluralization above.

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,94 @@ import (
...
+type ImagesRef []ImageRef
+
+func (i *ImagesRef) AddImagesRef(refs ...ImageRef) {
+	result := *i
+	for _, ref := range refs {
+		found := false
+		for j, imageRef := range result {
+			if imageRef.Image == ref.Image {
+				found = true
+				result[j] = ref
```
cppforlife: should we do a deep copy?

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,94 @@ import (
...
+	}
+
+	imgsLock := &ImagesLock{imagesRef: imagesRef, imgRetriever: imgRetriever}
 	imgsLock.generateImagesLocations(relativeToRepo)
 	return imgsLock
 }
 
 type ImagesLock struct {
-	imagesLock   lockconfig.ImagesLock
+	imagesRef    []ImageRef
```
cppforlife: should we use ImageRefs type here?

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -67,24 +97,78 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
...
+		return nil, err
+	} else {
+		imagesLock, err := o.imagesLockReader.Read(img)
+		if err != nil {
 			return nil, err
 		}
+
+		currentImagesLock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo())
+		imageRefs = currentImagesLock.ImagesRefs()
+	}
```
cppforlife: i think we should always read ImagesLock (it's still source of truth for referenced images by the bundle), but if we also find locations then we should AddLocation to the image ref.

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -9,53 +9,83 @@ import (
...
+
 		if skip := processedImgs.CheckAndAddImage(image.Image); skip {
 			errChan <- nil
 			continue
 		}
 
 		image := image.DeepCopy()
 		go func() {
-			imgsLock, imgRef, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs)
+			lNestedBundles, imgsLock, imgRef, err := o.imagesLockIfIsBundle(throttleReq, image, processedImgs, logger)
```
cppforlife: what's `l` in `lNestedBundles`? should we just call it nestedBundles?

pkg/imgpkg/bundle/bundle.go
```diff
@@ -29,24 +31,83 @@ type Bundle struct {
...
 func NewBundleWithReader(ref string, imagesMetadata ctlimg.ImagesMetadata, imagesLockReader ImagesLockReader) *Bundle {
-	return &Bundle{plainimg.NewPlainImage(ref, imagesMetadata), imagesMetadata, imagesLockReader}
+	return &Bundle{plainImg: plainimg.NewPlainImage(ref, imagesMetadata), imgRetriever: imagesMetadata, imagesLockReader: imagesLockReader, imagesRef: map[string]ImageRef{}}
 }
 
 func (o *Bundle) DigestRef() string { return o.plainImg.DigestRef() }
 func (o *Bundle) Repo() string      { return o.plainImg.Repo() }
 func (o *Bundle) Tag() string       { return o.plainImg.Tag() }
 
+func (o *Bundle) AddImagesRef(refs ...ImageRef) {
```
cppforlife: given all the work in this PR, im not sure i see the purpose of imageRefs on bundle being public. from what i see it's only really used by the NoteCopy.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -29,24 +31,83 @@ type Bundle struct {
...
+	if err != nil {
+		return err
+	}
+	destinationRef, err := regname.ParseReference(bundleProcessedImage.DigestRef)
+	if err != nil {
+		return err
+	}
+
+	destDigestRef := destinationRef.Context().Digest(originBundleDigestRef.DigestStr())
+
```
cppforlife: isnt ref that we care about here the bundleProcessedImage.DigestRef? why do we need to use digest from o.DigestRef? (two should be the same).

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -113,70 +125,72 @@ func (c CopyRepoSrc) getSourceImages() (*ctlimgset.UnprocessedImageRefs, error)
...
+		return unprocessedImageRefs, bundles, nil
 	}
 }
 
-func (c CopyRepoSrc) getBundleImageRefs(bundleRef string) (*ctlbundle.Bundle, []lockconfig.ImageRef, error) {
-	bundle := ctlbundle.NewBundle(bundleRef, c.registry)
-
-	imgLock, err := bundle.AllImagesLock(c.Concurrency)
+func (c *CopyRepoSrc) getBundleImageRefs(bundleRef string) ([]*ctlbundle.Bundle, ctlbundle.ImagesRef, error) {
+	c.originBundle = ctlbundle.NewBundle(bundleRef, c.registry)
```
cppforlife: instead of saving this off, we should just return rootBundle. looks like `...) (*Bundle, []*ctlbundle.Bundle, ctlbundle.ImagesRef, error)`. otherwise it's just a hidden contract.

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -77,34 +82,41 @@ func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error)
 		return nil, fmt.Errorf("Building import repository ref: %s", err)
 	}
 
+	c.logger.Debugf("copy the fetched images\n")
 	processedImages, ids, err := c.imageSet.Relocate(unprocessedImageRefs, importRepo, c.registry)
 	if err != nil {
 		return nil, err
 	}
 
+	for _, bundle := range bundles {
```
cppforlife, line 89: shouldnt bundles also contain `originBundle` so that it can get a NoteCopy call?

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -9,53 +9,83 @@ import (
...
+	for _, bundle := range bundles {
+		for _, ref := range bundle.ImagesRef() {
+			imgRef, found := imagesRef.ImageRef(ref.Image)
+			if !found {
+				panic(fmt.Sprintf("Internal inconsistency: The Image '%s' cannot be found in the total list of images", ref.Image))
+			}
+
+			bundle.AddImagesRef(imgRef)
+		}
+	}
```
cppforlife: i think we need to figure out a better way to do this. trying to think what are we actually storing the state for...

test/helpers/logger.go
joaopapereira: no, but we need to comply with `util.LoggerWithLevels` interface

pkg/imgpkg/bundle/locations_config.go
joaopapereira: This function does not allow unknown fields. This will only cause problems if you are using an older version of `imgpkg` with a newer version of the Image Locations configuration. To be fair I think we want that because when we reintroduce the Repository field we need to respect it.

pkg/imgpkg/bundle/locations.go
joaopapereira: Unfortunately, Fetch and Save need 2 different types of registry struct. One needs the Reader the other needs the Writer. If we moved the registry to the constructor we would have to provide a reader/writer to accommodate both cases.
The problem with this is that the Bundle struct currently only has Read access to the registry when collecting the Image References aka `Bundle.AllImagesRefs`.

I decided to move the registry to the function itself because the only situation we need to write on this struct is in the NoteCopy function, so we would be giving write access to the majority of the cases where we do not need it.

pkg/imgpkg/bundle/locations.go
joaopapereira: Not a fan of it, nevertheless I was trying to keep some consistency with the `Printf` function. But I am open to change all the logger functions to provide the newline on them.

pkg/imgpkg/cmd/copy.go
joaopapereira: done

pkg/imgpkg/cmd/copy.go
joaopapereira: done

pkg/imgpkg/bundle/locations_config.go
joaopapereira: Done

pkg/imgpkg/bundle/locations_config.go
joaopapereira: done

pkg/imgpkg/bundle/locations_config.go
joaopapereira: done

pkg/imgpkg/bundle/locations_config.go
joaopapereira: done

pkg/imgpkg/bundle/locations.go
joaopapereira: done

pkg/imgpkg/bundle/locations.go
joaopapereira: done

pkg/imgpkg/bundle/locations.go
joaopapereira: fixed

pkg/imgpkg/bundle/locations.go
joaopapereira: done

pkg/imgpkg/bundle/locations.go
joaopapereira: done

pkg/imgpkg/bundle/images_lock.go
joaopapereira: Done

pkg/imgpkg/bundle/images_lock.go
joaopapereira: I did this, the only funky thing is that now you need to call `ImageRefs.ImageRefs` to get the refs instead of just iterating over the `ImagesRefs` Slice

pkg/imgpkg/bundle/images_lock.go
joaopapereira: Done

pkg/imgpkg/bundle/images_lock.go
joaopapereira: Done

pkg/imgpkg/bundle/images_lock.go

pkg/imgpkg/bundle/images_lock.go

pkg/imgpkg/bundle/bundle_images_lock.go

pkg/imgpkg/bundle/bundle.go
joaopapereira: We are using it for testing to ensure that the found references are the ones we are expecting

pkg/imgpkg/bundle/bundle.go
joaopapereira: yes you are right

pkg/imgpkg/bundle/images_lock.go
joaopapereira: fixed

pkg/imgpkg/registry/with_progress.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: fixed

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: fixed

pkg/imgpkg/registry/with_progress.go
joaopapereira: fixed

pkg/imgpkg/registry/registry.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: the originBundle is added as part of the call to `Bundle.AllImagesRefs`

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: We are storing the state for each bundle of the Images that are part of it. This later is used when we are trying to generate each individual Locations OCI Image to find from the list of the ProcessedImages the ones that each bundle needs.

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: Done

pkg/imgpkg/bundle/locations.go
cppforlife: sounds good. makes sense.

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -13,52 +13,103 @@ import (
...
+func (i *ImageRefs) Find(ref string) (ImageRef, bool) {
+	for _, imageRef := range i.refs {
+		if imageRef.Image == ref {
+			return imageRef, true
+		}
+	}
+
+	return ImageRef{}, false
+}
+func (i ImageRefs) ImageRefs() []ImageRef {
```
cppforlife, line 60: if you do not like ImageRefs{}.ImageRefs(), it could be called ImageRefs{}.All()

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -9,82 +9,165 @@ import (
...
 			mutex.Lock()
 			defer mutex.Unlock()
-			resultImagesLock.AddImageRef(imgRef)
-			if imgsLock != nil {
-				err = resultImagesLock.Merge(imgsLock)
-				if err != nil {
-					errChan <- fmt.Errorf("Merging images for bundle '%s': %s", image.Image, err)
-					return
-				}
+			if nestedBundles != nil {
```
cppforlife: dont think we need to check on nil here. it's typed so ... will take care of it.

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -9,82 +9,165 @@ import (
...
+				bundles = append(bundles, nestedBundles...)
 			}
+
+			// Adds Image to the resulting ImagesLock
+			processedImageRefs.AddImagesRef(ImageRef{
+				ImageRef: imgRef,
+				IsBundle: nestedBundles != nil, // nestedBundles will be != nil when the image is a bundle
```
cppforlife: we should check `len(nestedBundles) > 0` instead since nil and empty are not same.

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -59,12 +59,14 @@ func (c CopyRepoSrc) CopyToTar(dstPath string) error {
 	return nil
 }
 
-func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error) {
-	unprocessedImageRefs, err := c.getSourceImages()
+func (c *CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error) {
```
cppforlife: do we need to make CopyRepoSrc a pointer here after removal of originBundle?

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -59,12 +59,14 @@ func (c CopyRepoSrc) CopyToTar(dstPath string) error {
...
+	c.logger.Tracef(fmt.Sprintf("CopyToRepo(%s)\n", repo))
```
cppforlife: fmt.Sprintf is not necessary given f in Tracef

pkg/imgpkg/bundle/locations.go
```diff
@@ -0,0 +1,176 @@
...
+	regv1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/k14s/imgpkg/pkg/imgpkg/image"
+	"github.com/k14s/imgpkg/pkg/imgpkg/plainimage"
+	"github.com/k14s/imgpkg/pkg/imgpkg/util"
+)
+
+const (
+	locationsTag string = "%s-%s.locations.imgpkg"
```
cppforlife: lets changing this to `.image-locations.imgpkg`

pkg/imgpkg/bundle/locations_config.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	LocationFilepath   = "images-locations.yml"
+	LocationKind       = "Locations"
```
cppforlife: `LocationKind       = "Locations"`
to
`ImageLocationsKind       = "ImageLocations"`

pkg/imgpkg/bundle/locations_config.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	LocationFilepath   = "images-locations.yml"
```
cppforlife: "images-locations.yml" -> "image-locations.yml"

pkg/imgpkg/bundle/locations_config.go
cppforlife: > This function does not allow unknown fields. This will only cause problems if you are using an older version of imgpkg with a newer version of the Image Locations configuration.

this may happen for if kapp-controller's impgkg is on version A, and bundle was created with imgpkg version A+1. i think if meaning of ImageLocations changes significantly we should rev apiVersion. if meaning doesnt change significantly then we should be ok with new fields being optional. wdyt?

pkg/imgpkg/bundle/locations.go
cppforlife: locationsTag -> locationsTagFmt

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -9,82 +9,165 @@ import (
...
+		return ImageRefs{}, ImageRefs{}, err
+	}
+	return imageRefsToProcess, processedImageRefs, nil
+}
+
+func (o *Bundle) processLocations(imageRefs ImageRefs, locationsConfig ImageLocationsConfig) (ImageRefs, ImageRefs) {
+	resultImageRefs := imageRefs.DeepCopy()
+	processedImageRefs := ImageRefs{}
+	for _, imgRef := range imageRefs.ImageRefs() {
+		for _, image := range locationsConfig.Images {
```
cppforlife: image -> imgLoc?

pkg/imgpkg/bundle/bundle_images_lock.go

pkg/imgpkg/bundle/bundle_images_lock.go

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: removed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: done

pkg/imgpkg/bundle/locations.go
joaopapereira: done

pkg/imgpkg/bundle/locations_config.go
joaopapereira: Done

pkg/imgpkg/bundle/locations_config.go
joaopapereira: done

pkg/imgpkg/bundle/bundle_images_lock.go

joaopapereira:
> > This function does not allow unknown fields. This will only cause problems if you are using an older version of imgpkg with a newer version of the Image Locations configuration.
>
> this may happen for if kapp-controller's impgkg is on version A, and bundle was created with imgpkg version A+1. i think if meaning of ImageLocations changes significantly we should rev apiVersion. if meaning doesnt change significantly then we should be ok with new fields being optional. wdyt?

This is exactly the scenario that we want to guard against. In your example, `imgpkg` assumes that the images will be present in the bundle repository, which might not be true. This will cause an error because kapp-controller will not be able to get the Images that are part of the bundle

PR #160, opened 2021-06-08
improve pull performance by using OCI locations

DennisDenuto:
When pulling a bundle, in order to speed up the images re-writing, imgpkg will consider a bundle relocated if it can find its corresponding Location OCI.

- if it cannot get the Location OCI for the bundle, either it doesn't exist on the registry, or there was an error when attempting to get the location OCI, then it will fallback to the 'old' (slower) way of determining if a bundle has been relocated.


pkg/imgpkg/bundle/bundle.go
```diff
@@ -132,6 +133,21 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 		ui.BeginLinef("Pulling nested bundle '%s'\n", o.DigestRef())
 	}
 
+	bundleDigestRef, err := regname.NewDigest(o.plainImg.DigestRef())
+	if err != nil {
+		return err
+	}
+
+	loggerBuilder := util.NewLogger(uiBlockWriter{ui})
+	locationFetcher := cachedLocationFetcher{
```
joaopapereira: Can we name the parameters of the struct? it will help with the `nil` case

pkg/imgpkg/bundle/bundle.go
```diff
@@ -132,6 +133,21 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
...
+		LocationFetcher{
+			logger:          loggerBuilder.NewLevelLogger(util.LogWarn, loggerBuilder.NewPrefixedWriter("")),
```
joaopapereira: (Maybe different PR) Not liking this constructor, NewLevelLogger should not need a prefix writer on it. Maybe if we want a prefix we should have a constructor like `NewPrefixedLevelLogger(loglevel util.LogLevel, prefix string)` that would create a prefixed or use a different mechanism

pkg/imgpkg/bundle/bundle.go
```diff
@@ -142,12 +158,21 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
...
-	localizedImagesLockToRepo, notLocalizedToBundle, err := NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).LocalizeImagesLock()
+	lock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo(), locationFetcher)
+	localizedImagesLockToRepo, notLocalizedToBundle, err := lock.LocalizeImagesLock()
 	if err != nil {
 		return err
 	}
 
 	if pullNestedBundles {
+		imageBundles := map[string]bool{}
+		fetch, err := locationFetcher.Fetch()
```
joaopapereira: Let's rename `fetch` to `locationsConfig`

joaopapereira: even if we have this information cached, don't you feel like it is redundant to call the fetch here again since we have already done that on LocalizeImagesLock?

pkg/imgpkg/bundle/bundle.go
```diff
@@ -142,12 +158,21 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 		return err
 	}
 
-	localizedImagesLockToRepo, notLocalizedToBundle, err := NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).LocalizeImagesLock()
+	lock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo(), locationFetcher)
```
joaopapereira: I would like to remove the ImagesLock concept from the bundle package and remove the struct associated with it.
At this point, this struct is a wrapper around the ImageRefs which feels like a better abstraction. Eventually, this type could be used to generate an ImagesLock struct.

Maybe we can sit down and discuss this a little bit

test/perf/perf_test.go
```diff
@@ -75,7 +75,7 @@ func TestBenchmarkCopyingLargeBundleThatContainsImagesMostlyOnDockerHub(t *testi
 		maxTimeCopyShouldTake = 8 * time.Minute.Nanoseconds()
 	}
 
-	require.Greaterf(t, maxTimeCopyShouldTake, actualTimeTaken, "copying a large bundle took too long. Expected it to take maximum [%v] but it took [%v]", time.Duration(maxTimeCopyShouldTake), time.Duration(actualTimeTaken))
+	require.Greaterf(t, maxTimeCopyShouldTake, actualTimeTaken, fmt.Sprintf("copying a large bundle took too long. Expected it to take maximum [%v] but it took [%v]", time.Duration(maxTimeCopyShouldTake), time.Duration(actualTimeTaken)))
```
joaopapereira, line 78: Check if this is needed, if it is please have a separated commit for it

pkg/imgpkg/bundle/bundle.go
```diff
@@ -231,3 +262,34 @@ func (o *Bundle) checkedImage() (regv1.Image, error) {
...
+		ImageLocationsConfig
+		error
+	}
+}
+
+func (c cachedLocationFetcher) Fetch() (ImageLocationsConfig, error) {
+	if c.cachedResult != nil {
+		return c.cachedResult.ImageLocationsConfig, c.cachedResult.error
+	}
+	fetch, err := c.LocationFetcher.Fetch()
```
joaopapereira: Let's rename fetch to locationsConfig

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -66,20 +66,50 @@ func (i ImageRefs) DeepCopy() ImageRefs {
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	for _, imgRef := range o.imageRefs.ImageRefs() {
+		for _, imgLoc := range locationsConfig.Images {
+			if imgLoc.Image == imgRef.Image {
+				// We need to keep all the ImagesLock information and the only added pieces is if this image is a bundle or not
```
joaopapereira: No longer makes sense the comment, please remove

joaopapereira (changes requested):
Let's maybe keep the big refactor separated that will in some way remove the LocalizeImagesLock function and remove the ImagesLock struct from the bundle package.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: yep! done

pkg/imgpkg/bundle/bundle.go
DennisDenuto: renamed to `locationsConfig`

> even if we have this information cached, don't you feel like it is redundant to call the fetch here again since we have already done that on LocalizeImagesLock?

Will address this in a follow up refactor. With the goal of removing ImagesLock.Localized and instead getting ImageRefs with Bundle information (removing the need of calling Fetch again)

pkg/imgpkg/bundle/bundle.go
DennisDenuto: we sat down and talked. I like the general idea. Will work on this larger refactor in a follow up PR.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: done.

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: done. removed the comment.

test/perf/perf_test.go
DennisDenuto: Yeah we need it. The GreaterF function has a bug around printing a msg with args. It ends up looking like this otherwise:
```
Messages:   	[copying a large bundle took too long. Expected it to take maximum [%v] but it took [%v] 123ns 321ns]
```

I'll create a separate commit for this.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: Yeah good call out. I agree. I'll leave it to a different PR since it effects copy.

Also, I wonder if we shouldn't use a logger, and instead use `ui`. This would involve changing all of the copy cmd from using `logger` to `ui`.

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -108,38 +137,54 @@ func (o *ImagesLock) LocalizeImagesLock() (lockconfig.ImagesLock, bool, error) {
 	var imageRefs []lockconfig.ImageRef
 	imagesLock := lockconfig.NewEmptyImagesLock()
 
-	skippedLocalization := false
-	for _, imgRef := range o.imageRefs.ImageRefs() {
-		foundImg, err := o.imgRetriever.FirstImageExists(imgRef.Locations())
-		if err != nil {
-			return lockconfig.ImagesLock{}, false, err
+	_, err := o.imagesLockLocationFetcher.Fetch()
```
cppforlife: im not sure i understand why this call is doing here. we only look at returned err. what happens when err==nil?

pkg/imgpkg/bundle/bundle.go
```diff
@@ -231,3 +262,34 @@ func (o *Bundle) checkedImage() (regv1.Image, error) {
 	}
 	return img, err
 }
+
+type cachedLocationFetcher struct {
```
cppforlife: instead of introducing do-er object here, is there a reason why we cannot be using Locations object? if we need to cache it, can we introduce CachedLocations object?

pkg/imgpkg/bundle/bundle.go
```diff
@@ -142,12 +158,21 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
...
 	if err != nil {
 		return err
 	}
 
 	if pullNestedBundles {
+		imageBundles := map[string]bool{}
+		locationsConfig, err := locationFetcher.Fetch()
+		if err == nil {
+			for _, image := range locationsConfig.Images {
+				imageBundles[lock.imageRelativeToBundle(image.Image, o.Repo())] = image.IsBundle
```
cppforlife: private method access outside of the class.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -142,12 +158,21 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
...
-	localizedImagesLockToRepo, notLocalizedToBundle, err := NewImagesLock(imagesLock, o.imgRetriever, o.Repo()).LocalizeImagesLock()
+	lock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo(), locationFetcher)
+	localizedImagesLockToRepo, notLocalizedToBundle, err := lock.LocalizeImagesLock()
 	if err != nil {
 		return err
 	}
 
 	if pullNestedBundles {
+		imageBundles := map[string]bool{}
+		locationsConfig, err := locationFetcher.Fetch()
```
cppforlife: it seems that we are dealing with the guts of ImagesLock class, but outside of it? it would be better we encapsulate it into ImagesLock.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -157,11 +182,17 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
 				continue
 			}
 
-			subBundle := NewBundle(image.Image, o.imgRetriever)
-			isBundle, err := subBundle.IsBundle()
-			if err != nil {
-				return err
+			var isBundle, ok bool
+			var subBundle *Bundle
```
cppforlife: no need for this var, could just use := below

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -66,20 +66,49 @@ func (i ImageRefs) DeepCopy() ImageRefs {
...
 	}
 
-	imgsLock := &ImagesLock{imageRefs: imageRefs, imgRetriever: imgRetriever}
+	imgsLock := &ImagesLock{imageRefs: imageRefs, imgRetriever: imgRetriever, imagesLockLocationFetcher: imagesLockLocationConfig}
 	imgsLock.generateImagesLocations(relativeToRepo)
+
 	return imgsLock
 }
 
+func (o *ImagesLock) SyncImageRefs() error {
```
cppforlife: it seems that this method is mostly used to mutate ImagesLock with isBundle information. from the design perspective it seems quite surprising that i need to call this method myself, instead of ImagesLock managing such information internally (aka internal detail is leaking out to consumers). additionally i would have expected that we would be recording that image is at a particular location (that's why we introduced locations in the first place), but i dont see happening.

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: > what happens when err==nil?

See line 179. It is the outside of the if err != nil block.
 
when err == nil (the locations OCI *was* found) the images ref is 'localized' with the destination repo.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: The usage of isBundle is outside the scope of the if statement. Hence declaring the variable there.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: I don't understand your concern or reason to change this part of the code. 

But It sounds like you are asking is there a reason I didn't instead just keep a variable reference to the `ImageLocationsConfig` and *probably* `error`? and pass those variables around to the things that need it? and call it a CachedLoctions object?

I think that would work too.

The rationale I had originally with `cachedLocationFetcher` was that there is already an API defined to `fetch` an `ImageLocationsConfig` and an `error` for a specific bundle. So decorate that with a simple caching layer since it is used for 2 reasons. Hence avoiding an unnecessary http call since the Location OCI shouldn't change between those 2 calls.

pkg/imgpkg/bundle/bundle.go
cppforlife: isBundle yeah, but subBundle could just use :=

pkg/imgpkg/bundle/images_lock.go
cppforlife: how is it localized? im guessing you are saying .imagesRefs is somehow holding localized references (`for _, imgRef := range o.imageRefs.ImageRefs() {` ) but from this function there is no way to tell how that happened. is it stateful? is that what that Sync method is for?

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: as discussed over zoom, i moved that function into `ImageRefs`

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -257,3 +229,13 @@ func (o *singleLayerReader) Read(img regv1.Image) (lockconfig.ImagesLock, error)
 
 	return lockconfig.NewImagesLockFromBytes(bs)
 }
+
+type LocationFetcher struct {
```
cppforlife: i forget what was the name we wanted to rename this to: LocationsConfig?

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -93,9 +100,37 @@ func (o *ImagesLock) generateImagesLocations(relativeToRepo string) {
...
+		return ImageRefs{}, err
+	}
+	return o.imageRefs, nil
+}
+
+func (o *ImagesLock) syncImageRefs() error {
+	locationsConfig, err := o.imagesLockLocationFetcher.Config()
+	if _, ok := err.(*LocationsNotFound); ok {
+		return nil
+	}
```
cppforlife, line 123: nit: may be move this inside `err != nil`?

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -93,9 +100,37 @@ func (o *ImagesLock) generateImagesLocations(relativeToRepo string) {
...
+		return err
+	}
+
+	imgRefs := o.imageRefs.ImageRefs()
+	for i, imgRef := range imgRefs {
+		for _, imgLoc := range locationsConfig.Images {
+			if imgLoc.Image == imgRef.Image {
+				isBundle := imgLoc.IsBundle
+				imgRefs[i].IsBundle = &isBundle
+			}
```
cppforlife: lets move this into ImageRefs class, so that it reads something like `o.imageRefs.MarkAsBundle(imgLoc.Image, imgLock.IsBundle)`. otherwise it's reading like we are modifying some local array (imgsRefs) and throwing it away.

pkg/imgpkg/bundle/locations.go
```diff
@@ -34,7 +34,7 @@ func (n LocationsNotFound) Error() string {
 	return fmt.Sprintf("Locations image in %s could not be found", n.image)
 }
 
-type Locations struct {
+type LocationsConfigs struct {
```
cppforlife, line 43: should change file name as well

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -66,20 +66,27 @@ func (i ImageRefs) DeepCopy() ImageRefs {
 	return result
 }
 
-func NewImagesLock(imagesLock lockconfig.ImagesLock, imgRetriever ctlimg.ImagesMetadata, relativeToRepo string) *ImagesLock {
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . ImagesLockLocationConfig
+type ImagesLockLocationConfig interface {
```
cppforlife: ...LocationConfig -> ...LocationsConfig

pkg/imgpkg/bundle/images_lock.go
```diff
@@ -66,20 +66,27 @@ func (i ImageRefs) DeepCopy() ImageRefs {
...
+
 	return imgsLock
 }
 
 type ImagesLock struct {
-	imageRefs    ImageRefs
-	imgRetriever ctlimg.ImagesMetadata
+	imageRefs                 ImageRefs
+	imgRetriever              ctlimg.ImagesMetadata
+	imagesLockLocationFetcher ImagesLockLocationConfig
```
cppforlife: imagesLockLocationFetcher -> locationsConfig for consistency

pkg/imgpkg/bundle/bundle_images_lock.go
DennisDenuto: ok renamed to LocationsConfig

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: i'm going to leave it as-is since it is a 'nit' and also to me a flatter structure is a lot easier to comprehend (and is the reason I decided to avoid nesting it in the first place)

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: ok done

pkg/imgpkg/bundle/locations.go
DennisDenuto: done

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: sure!

pkg/imgpkg/bundle/images_lock.go
DennisDenuto: changed

joaopapereira (approved):
LGTM

PR #162, opened 2021-06-08
Fix oci locations setting images to null

### LINKED ISSUE #161: OCI locations has images field set to null
**What steps did you take:**
Run the following e2e test:
https://github.com/vmware-tanzu/carvel-imgpkg/blob/0bd8263585390fcd01e3c197c1ca1c27cf7f4313/test/e2e/pull_test.go#L19

**What happened:**
an oci location is generated. pull down the image and inspect the contents.

**What did you expect:**

Contents to contain the `hello-world` image referenced in the e2e test
```
---
apiVersion: imgpkg.carvel.dev/v1alpha1
images: 
 - hello-world@sha...
kind: ImageLocations
```

Got:
```
---
apiVersion: imgpkg.carvel.dev/v1alpha1
images: null
kind: ImageLocations
```

This doesn't seem to be an issue if the image referenced lives in the same repo as the bundle being copied.

**Environment:**

- imgpkg version (use `imgpkg --version`): develop
- Docker registry used (e.g. `Docker HUB`):  dockerhub
- OS (e.g. from `/etc/os-release`): osx

---
Vote on this request

DennisDenuto:
Attempts to fix: https://github.com/vmware-tanzu/carvel-imgpkg/issues/161

There seems to be a couple of issues:
- Working around issue around setting a bundle image's ref, if that image is missing repo information then the oci locations sets images to null
- when imgpkg copy runs twice, the second time creates an oci location with images set to null

### DIFF
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -58,6 +58,15 @@ func (o *Bundle) addImageRefs(refs ...ImageRef) {
 
 func (o *Bundle) imageRef(imageDigest string) (ImageRef, bool) {
 	ref, found := o.imagesRef[imageDigest]
+	if !found {
+		for _, imgRef := range o.imagesRef {
+			for _, loc := range imgRef.Locations() {
+				if loc == imageDigest {
+					return imgRef, true
+				}
+			}
+		}
+	}
 	return ref, found
 }
 
@@ -86,6 +95,10 @@ func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesM
 		if image.UnprocessedImageRef.DigestRef == o.DigestRef() {
 			bundleProcessedImage = image
 		}
+	}
+
+	if len(locationsCfg.Images) != len(o.imagesRef) {
+		panic(fmt.Sprintf("Expected: %d images to be written to Location OCI. Actual: %d were written", len(o.imagesRef), len(locationsCfg.Images)))
 	}
 
 	destinationRef, err := regname.NewDigest(bundleProcessedImage.DigestRef)
--- a/pkg/imgpkg/bundle/bundle_images_lock.go
+++ b/pkg/imgpkg/bundle/bundle_images_lock.go
@@ -35,7 +35,6 @@ func (o *Bundle) AllImagesRefs(concurrency int, logger util.LoggerWithLevels) ([
 			if !found {
 				panic(fmt.Sprintf("Internal inconsistency: The Image '%s' cannot be found in the total list of images", ref.Image))
 			}
-			imgRef = imgRef.DeepCopy()
 			bundle.addImageRefs(imgRef)
 		}
 	}
@@ -61,7 +60,6 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 	mutex := &sync.Mutex{}
 
 	for _, image := range imageRefsToProcess.ImageRefs() {
-		image = image.DeepCopy()
 		o.addImageRefs(image)
 
 		if skip := processedImgs.CheckAndAddImage(image.Image); skip {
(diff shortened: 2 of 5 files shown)


pkg/imgpkg/bundle/bundle.go
```diff
@@ -58,6 +58,15 @@ func (o *Bundle) addImageRefs(refs ...ImageRef) {
 
 func (o *Bundle) imageRef(imageDigest string) (ImageRef, bool) {
 	ref, found := o.imagesRef[imageDigest]
+	if !found {
+		for _, imgRef := range o.imagesRef {
+			for _, loc := range imgRef.Locations() {
```
cppforlife, line 63: hmm, i dont think i understand how locations become involved here. afaik everywhere the primary key is `image` field value. are we screwing that up somewhere and that's why we need to start looking at locations?

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -35,7 +35,12 @@ func (o *Bundle) AllImagesRefs(concurrency int, logger util.LoggerWithLevels) ([
 			if !found {
 				panic(fmt.Sprintf("Internal inconsistency: The Image '%s' cannot be found in the total list of images", ref.Image))
 			}
-
+			imgRef = imgRef.DeepCopy()
+			parsedImageRefName, err := regname.ParseReference(ref.Image)
```
cppforlife: if we are changing this stuff here this would affect places that end up reading stuff (second copy or pull). i dont think added test accounts for that currently.

DennisDenuto: > affect places that end up reading stuff

affect in a negative way? I fail to see how using a fully qualified name would cause issues when reading? 

> second copy or pull

Ok i'll dig into this concern.
fwiw the modified e2e test in the PR effectively does this scenario already and runs assertions on the oci location file too.
Although I can add a test to assert that when copying twice that the bundle sha doesn't change.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: discussed over zoom. short version is that for optimization reasons the imageDigest sometimes references the image ref of the target repo. However the image ref in the images lock (belonging to the bundle) has the original source reference.

Thus, since they are sometimes different they don't match in this function and never make it to the Locations image.

The proposed fix is to iterate over all the locations to try and 'map' an image to a bundle's image ref to decide whether to include it as part of the location OCI.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -88,6 +97,10 @@ func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesM
 		}
 	}
 
+	if len(locationsCfg.Images) != len(o.imagesRef) {
+		return fmt.Errorf("Expected: %d images to be written to Location OCI. Actual: %d were written", len(o.imagesRef), len(locationsCfg.Images))
```
cppforlife: let's change this to panic since this is a logic error in the program vs error in a correctly functioning program. (e.g. https://github.com/vmware-tanzu/carvel-imgpkg/blob/a1a755aab6644dc9dc0b70c5ce26d7d7d2a7d0f7/pkg/imgpkg/bundle/images_lock.go#L148)

pkg/imgpkg/lockconfig/images_lock.go
```diff
@@ -58,6 +58,17 @@ func NewImagesLockFromBytes(data []byte) (ImagesLock, error) {
 		return lock, fmt.Errorf("Validating images lock: %s", err)
 	}
 
+	// Update the image lock file to use a fully qualified name
+	// i.e. if a user provides ubuntu (short hand for library/ubuntu) in the ImageLock file,
+	// downstream processing will fail when comparing if images match.
+	for i, img := range lock.Images {
+		parsedImageRefName, err := regname.NewDigest(img.Image)
+		if err != nil {
+			return lock, err
```
cppforlife: let's change this to panic as well (per similar comment above). Validate() checks that everything is a digest already, so if this error something is terribly wrong with the program (not the data).

pkg/imgpkg/bundle/bundle.go
DennisDenuto: done

pkg/imgpkg/lockconfig/images_lock.go
DennisDenuto: done

PR #164, opened 2021-06-10
Add test assets to .gitignore

DennisDenuto:
- e2e test generates cosign.key and cosign.pub


cppforlife:
can these files be generated inside our tmp/ instead of top level?

DennisDenuto:
@cppforlife 

Ok generated cosign test assets under tmp/

Also installed cosign via `go install` simplifying the test-e2e.sh script

PR #165, opened 2021-06-10
More deterministic push

cppforlife:
turns out dir.Size() is not necessary and more importantly not deterministic across different OSes. remove it so that imgpkg push generates same digest everywhere.


vijaykatam (approved):

DennisDenuto:
As discussed, this PR causes a breaking change. The same tarball created with an older version of imgpkg will result in a different sha. However this is a similar breaking change to the file permission change introduced earlier.

The advantages of having reproducible builds across diff os's outweighs the breaking change.

PR #175, opened 2021-06-24
Generate pulled ImagesLock based on the file in the Bundle

joaopapereira:


pkg/imgpkg/bundle/images_lock.go
```diff
@@ -200,13 +206,16 @@ func (o *ImagesLock) LocalizeImagesLock() (ImageRefs, lockconfig.ImagesLock, boo
 			bundleImageRefs = NewImageRefs()
 			// Remove the bundle location on all the Images, which is present due to the constructor call to
 			// ImagesLock.generateImagesLocations
-			for _, image := range o.imageRefs.ImageRefs() {
-				lockImgRef := image.DiscardLocationsExcept(image.Image)
```
cppforlife, line 204: do we no longer need to call DiscardLocationsExcept?

joaopapereira: We only do `generateImagesLocations` on the `refs` and the imagesLock that we  store in this struct is not changed

PR #177, opened 2021-06-24
Refactor remove images lock

DennisDenuto:
The goal of this PR is to consolidate the 'localization' and 'image lock re-writing' into ImageRefs and removing the bundle 'imagelock' struct.

- ImageRefs has an ImagesLock (not the same as the above which is a now removed bundle ImagesLock) and a LocationConfig 
- UI logic when pulling a bundle has also been moved. This reduces the need of using a NoOp logger. Instead the top of the recursion stack (root bundle) is responsible for printing out logs about when the imageslock file is re-written.


pkg/imgpkg/bundle/bundle.go
```diff
@@ -135,32 +145,36 @@ func (o *Bundle) pull(baseOutputPath string, ui goui.UI, pullNestedBundles bool,
...
 	}
 
 	imagesLock, err := lockconfig.NewImagesLockFromPath(filepath.Join(baseOutputPath, bundlePath, ImgpkgDir, ImagesLockFile))
 	if err != nil {
-		return err
+		return false, err
 	}
 
-	lock := NewImagesLock(imagesLock, o.imgRetriever, o.Repo(), LocationsConfig{
+	bundleImageRefs, err := NewImageRefs(imagesLock, LocationsConfig{
```
cppforlife: NewImageRefs -> NewImageRefsFromImagesLock since you can imagine plain NewImageRefs constructor wouldnt have to rely on images lock

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -53,7 +53,7 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 		return nil, ImageRefs{}, err
 	}
 
-	processedImageRefs := NewImageRefs()
+	processedImageRefs := NewEmptyImageRefs()
```
cppforlife: NewEmptyImageRefs -> NewImageRefs()

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func NewImageRef(imgRef lockconfig.ImageRef, isBundle bool) ImageRef {
+	return ImageRef{ImageRef: imgRef, IsBundle: &isBundle}
+}
+
+func (i ImageRef) DeepCopy() ImageRef {
+	return ImageRef{
+		ImageRef: i.ImageRef.DeepCopy(),
+		IsBundle: i.IsBundle,
```
cppforlife: pointer is being copied. need to deref and make a new pointer.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+}
+func (i *ImageRef) updateRepo(relativeToRepo string) *ImageRef {
```
cppforlife: newline before

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	i.Image = i.replaceImageRepo(relativeToRepo)
+	return i
+}
+func (i *ImageRef) replaceImageRepo(relativeToRepo string) string {
```
cppforlife: newline before

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	updatedRepo := relativeToRepo + "@" + imgParts[1]
+	return updatedRepo
+}
+
+type ImageRefs struct {
+	refs                 []ImageRef
+	imageLocationsConfig *ImageLocationsConfig
+	originalImagesLock   lockconfig.ImagesLock
+
+	*sync.Mutex
```
cppforlife: mutex shouldnt be public.

cppforlife: it seems that this lock is meant for refs, so lets call it refsLock. read uses of i.refs should be using lock as well.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func NewEmptyImageRefs() ImageRefs {
+	return ImageRefs{
+		Mutex: &sync.Mutex{},
+	}
+}
+
+func NewImageRefs(imagesLock lockconfig.ImagesLock, imagesLockLocationConfig ImagesLockLocationsConfig) (ImageRefs, error) {
```
cppforlife: naming is off for imagesLockLocationConfig. ImagesLock is different thing from Locations.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+
+	for _, ref := range i.refs {
+		image, err := name.ParseReference(ref.updateRepo(relativeToRepo).Image)
+		if err != nil {
+			return false, err
+		}
+		_, err = imgRetriever.Digest(image)
+		if err != nil {
+			if terr, ok := err.(*transport.Error); ok {
+				if _, ok := imageNotFoundStatusCode[terr.StatusCode]; ok {
```
cppforlife: we should probably pull this check out as a package level method:

_, ok := imageNotFoundStatusCode[terr.StatusCode]; ok

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (i *ImageRefs) UpdateRelativeToRepo(imgRetriever ctlimg.ImagesMetadata, relativeToRepo string) (bool, error) {
+	if i.imageLocationsConfig != nil {
+		i.LocalizeToBundle(relativeToRepo)
+		return true, nil
+	}
+
+	for _, ref := range i.refs {
+		image, err := name.ParseReference(ref.updateRepo(relativeToRepo).Image)
```
cppforlife: it seems that updateRepo is mutating image ref. are we meaning to mutate some refs and then return false when finding one that isnt found? if our goal is to just get an updated location, then we should just use replaceImageRepo.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+			return false, err
+		}
+	}
+
+	i.LocalizeToBundle(relativeToRepo)
+
+	return true, nil
+}
+func (i *ImageRefs) AddImagesRef(refs ...ImageRef) {
```
cppforlife, line 109: newline

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+		for j, imageRef := range i.refs {
+			if imageRef.Image == ref.Image {
+				found = true
+				i.refs[j] = ref.DeepCopy()
+				break
+			}
+		}
+
+		if !found {
+			i.refs = append(i.refs, ref)
```
cppforlife: ref.DeepCopy()

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+}
+func (i *ImageRefs) Find(ref string) (ImageRef, bool) {
```
cppforlife, line 129: newline

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	for _, imageRef := range i.refs {
+		if imageRef.Image == ref {
+			return imageRef, true
```
cppforlife: should we be doing DeepCopy() here?

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+
+	return ImageRef{}, false
+}
+func (i *ImageRefs) MarkAsBundle(image string, isBundle bool) {
```
cppforlife, line 142: newline

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	for j, ref := range i.refs {
+		if ref.Image == image {
+			i.refs[j].IsBundle = &isBundle
+		}
+	}
+}
+func (i ImageRefs) ImagesLock() lockconfig.ImagesLock {
```
cppforlife, line 153: newline

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	imgLock := lockconfig.NewEmptyImagesLock()
+	for _, originalImg := range i.originalImagesLock.Images {
```
cppforlife, line 159: lets turn i.originalImagesLock into a pointer and panic with internal inconsistency message if it's nil. we always working against a real ImagesLock

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+		imgLock.Images = append(imgLock.Images, lockconfig.ImageRef{
+			Image:       ref.PrimaryLocation(),
+			Annotations: originalImg.Annotations,
+		})
+	}
+
+	return imgLock
+}
+func (i ImageRefs) ImageRefs() []ImageRef {
+	return i.refs
```
cppforlife: i think we should be deep copying everything coming out of this class.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+}
+func (i *ImageRefs) syncImageRefsWithLocationConfig(imagesLockLocationConfig ImagesLockLocationsConfig) error {
+	locationsConfig, err := imagesLockLocationConfig.Config()
+	if _, ok := err.(*LocationsNotFound); ok {
+		return nil
+	}
```
cppforlife, line 191: would be good to move this if into below if err != nil. would read better imho.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return ImageRefs{}, err
+	}
+	return imageRefs, nil
+}
+
+func (i *ImageRefs) LocalizeToBundle(relativeToRepo string) {
+	for _, imgRef := range i.refs {
+		imgRef.AddLocation(imgRef.replaceImageRepo(relativeToRepo))
+		i.AddImagesRef(imgRef)
```
cppforlife: `i.AddImagesRef(imgRef)` feels very weird here. it reads like we are double adding refs. since we have access to index in the loop, lets do i.refs[i] = imgRef

cppforlife: arguably we should just have AddLocation be mutable, so there is no need to screw around with refs

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	err := imageRefs.syncImageRefsWithLocationConfig(imagesLockLocationConfig)
+	if err != nil {
+		return ImageRefs{}, err
+	}
+	return imageRefs, nil
+}
+
+func (i *ImageRefs) LocalizeToBundle(relativeToRepo string) {
```
cppforlife: LocalizeToBundle -> LocalizeToRepo. ImageRefs doesnt know anything bundles.

pkg/imgpkg/bundle/images_refs.go
cppforlife: replaceImageRepo -> ImageRelativeToRepo (public). we arent really replacing anything, just calculating location. public because its used outside of ImageRef class.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: I don't understand the proposed function name change here. Today a bundle.ImageRefs is a bunch of image references from an images lock *and* a Location OCI. 

ImageRefs are *also* used to keep track of image references that have been 'processed'. But the original source of these image references came from an images lock *and* a Location OCI.

Also this is also 'relying' on a LocationsConfig which the proposed naming change doesn't capture...

pkg/imgpkg/bundle/bundle_images_lock.go
DennisDenuto: hmm `NewImageRefs` function name is already taken and used to create an ImageRefs with its dependencies.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: ok done.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: i'm curious... are these new line comments 'nits' or do you see these as required changes for this PR?

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: ok made it private

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: ok i renamed the struct.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: ok done.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: ok done!

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: I don't see why.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: I understand what is being asked. I don't understand the 'why'

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: why?

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: I prefer the flatter structure of these 2 different error scenarios. it has helped me differentiate the difference in return values based on these different classes of errors. 1 is a genuine error (something went wrong), the other is an error that represents something *not* going wrong, but that a bundle doesn't have a Location OCI.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: I disagree. I think we are replacing something. Specifically we are replacing the image's repository portion of its name with the argument provided.

I'm seeing this function being similar to `strings.Replace` where it takes an input and returns a copy of the input, replaced with something else.

It currently lives on the ImageRef struct. I think it could also live as a package function. And take 2 inputs. 1) an Image (string) and 2) relativeToRepo (string).

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: ok done.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: The ref in the for loop is copied by value. `ref.updateRepo` does 'mutate' a locally copied ref value, but `i.refs` does not contain that change. fwiw updateRepo also only mutates the `.Image` string field on an ImageRef and doesn't update any pointers or maps etc... 

if we return false because an image is found to not be relocated... then `i.refs` is not mutated at all.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: Ok i'm hoping to make this feel less weird by `AddLocation` now mutates the ref in the loop.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: ok done!

pkg/imgpkg/bundle/bundle.go
cppforlife: imageRefs is a just a collection of refs. one could imagine to be able to instantiate that plain collection via default constructor aka NewImageRefs() (i see that you called it NewEmptyImageRefs elsewhere, but we dont typically call other constructors NewEmpty* since that's typically the default state e.g. NewEmptyArray() vs NewArray()).

in this case we are instantiating such collection of ImageRefs _plus_ populating it with content from ImagesLock (primary source). technically, we are also using locations data to augment ImagesLock images but it's not the primary source of data, hence just the name NewImageRefsFromImagesLock().

pkg/imgpkg/bundle/bundle_images_lock.go
cppforlife: (see above naming comment why these names should be flipped.)

pkg/imgpkg/bundle/images_refs.go
cppforlife: it's a nit since it's a small stylistic suggestion. whether it's required is a tough question. in this case it's very atypical to have bunch of functions bunched up together without a newline separator (i have not seen that done in many go code bases) and given our existing code base style i would want these to get fixed unless there is a really good reason not to. imho its good to do a pass to fix things like these before putting PR in the review state.

pkg/imgpkg/bundle/images_refs.go
cppforlife: i see, so the replacement is really related to image contents rather than related to ImageRef structure itself. my default instinct was that it's related to ImageRef since that's where it's at.

> It currently lives on the ImageRef struct. I think it could also live as a package function. And take 2 inputs. 1) an Image (string) and 2) relativeToRepo (string).

i would agree with you on this, either on ImageRefs or package.

pkg/imgpkg/bundle/images_refs.go
cppforlife: i see. given this explanation whats the reasoning behind having updateRepo at all vs just doing something like `imgRef.replaceImageRepo(relativeToRepo)` like you were doing elsewhere?

pkg/imgpkg/bundle/images_refs.go
cppforlife: in general its typically good to return copies of objects to avoid accidental mutation down the line. in this case each ImageRef contains some mutable data which we wouldnt want any caller to modify (e.g. AddLocations on lockconfig.ImageRef for example) and affect internal state of ImageRefs struct.

pkg/imgpkg/bundle/images_refs.go
cppforlife: > we always working against a real ImagesLock

to expand on this: it seems that the intent for this class is to be able to generate ImagesLock based on _existing images lock_ + image refs. using this method against an ImageRefs instance that was never initialized with ImagesLock should be a programmer's mistake because we know that we must always be basing new images lock on existing images lock -- double-checking system invariants is generally a good pattern (unless its in the middle of the perf sensitive code). currently if this method is called on NewEmptyImagesRefs, it will just return empty ImagesLock. it's probably better to blow up to catch that or any other possibility instead of generating erroneous ImagesLock. this change ultimately firms up the contract of this class.

pkg/imgpkg/bundle/images_refs.go
cppforlife: (same as above deep copy comment)

pkg/imgpkg/bundle/images_refs.go
cppforlife: ok sounds good.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: `updateRepo` calls `imgRef.replaceImageRepo(relativeToRepo)`.  `replaceImageRepo` doesn't mutate anything, it is similar to a `strings.Replace`. `updateRepo` updates / mutates the `.Image` repo portion on the imgref object.

(as per our below convo) I'm going to move `imgRef.replaceImageRepo` to instead be a package level function that accepts 2 arguments.

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: Thanks for the explanation. good point i'll make the change.

pkg/imgpkg/bundle/bundle.go
DennisDenuto: ok thanks for the explanation. renamed the function NewImageRefsFromImagesLock

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: went with package

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: I did take a pass through and did intentionally group the functions for the structs together. I can adjust to different code styles. I think if this is a coding style for the existing code base (which I'm learning about now) I'll see If I can incorporate that into our linter. (as part of a separate PR)

DennisDenuto: i'll also add the new lines

pkg/imgpkg/bundle/images_refs.go
DennisDenuto: i'll make the change.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (i *ImageRefs) UpdateRelativeToRepo(imgRetriever ctlimg.ImagesMetadata, relativeToRepo string) (bool, error) {
+	if i.imageLocationsConfig != nil {
+		i.LocalizeToRepo(relativeToRepo)
+		return true, nil
+	}
+
+	for _, ref := range i.refs {
+		image, err := name.ParseReference(ref.updateRepo(relativeToRepo).Image)
```
cppforlife: so should this change to just using replaceImageRepo and then we remove updateRepo() method?

DennisDenuto: ok done.

PR #185, opened 2021-07-07
feat: introduce flag to set the http response timeout value

DennisDenuto:
- increased the default timeout from 10s -> 30s
- refactor: Remove copied golang http.DefaultTransport
- It is easier to see what has been 'changed' from the DefaultTransport.

### DIFF
--- a/pkg/imgpkg/cmd/registry_flags.go
+++ b/pkg/imgpkg/cmd/registry_flags.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"os"
+	"time"
 
 	"github.com/k14s/imgpkg/pkg/imgpkg/registry"
 	"github.com/spf13/cobra"
@@ -19,6 +20,8 @@ type RegistryFlags struct {
 	Password string
 	Token    string
 	Anon     bool
+
+	ResponseHeaderTimeout time.Duration
 }
 
 func (r *RegistryFlags) Set(cmd *cobra.Command) {
@@ -30,6 +33,8 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&r.Password, "registry-password", "", "Set password for auth ($IMGPKG_PASSWORD)")
 	cmd.Flags().StringVar(&r.Token, "registry-token", "", "Set token for auth ($IMGPKG_TOKEN)")
 	cmd.Flags().BoolVar(&r.Anon, "registry-anon", false, "Set anonymous auth ($IMGPKG_ANON)")
+
+	cmd.Flags().DurationVar(&r.ResponseHeaderTimeout, "registry-response-header-timeout", 30*time.Second, "Maximum time to allow a request to wait for a server's response headers from the registry (ms|s|m|h)")
 }
 
 func (r *RegistryFlags) AsRegistryOpts() registry.Opts {
@@ -42,6 +47,8 @@ func (r *RegistryFlags) AsRegistryOpts() registry.Opts {
 		Password: r.Password,
 		Token:    r.Token,
 		Anon:     r.Anon,
+
+		ResponseHeaderTimeout: r.ResponseHeaderTimeout,
 	}
 
 	if len(opts.Username) == 0 {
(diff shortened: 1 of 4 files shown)


pkg/imgpkg/registry/registry.go
```diff
@@ -222,25 +221,13 @@ func newHTTPTransport(opts Opts) (*http.Transport, error) {
...
-		TLSHandshakeTimeout:   10 * time.Second,
-		ResponseHeaderTimeout: 10 * time.Second,
-		ExpectContinueTimeout: 1 * time.Second,
-		// Use the cert pool with k8s cert bundle appended.
-		TLSClientConfig: &tls.Config{
-			RootCAs:            pool,
-			InsecureSkipVerify: (opts.VerifyCerts == false),
-		},
-	}, nil
+	clonedDefaultTransport := http.DefaultTransport.(*http.Transport).Clone()
```
cppforlife, line 226: the only downside i see here is that if something changes default transport this code will be affected. do we know if anything changing default transport anywhere?

DennisDenuto: the upside is that as we upgrade the http package, we get any changes / updates made to the DefaultTransport. (we avoid drifting from DefaultTransport)

i.e. since inlining the DefaultTransport, we drifted in 2 fields:
- `ForceAttemptHTTP2` setting to true 
- `DialContext.DualStack` has been deprecated 

sounds like a trade-off to me. While the risk you outline is true, currently nothing is mutating the `DefaultTransport` in our codebase or any of our libraries. 

My assumption is that the chance of drifting from stdlib is higher than the chance something will mutate DefaultTransport, and that's why I think this tradeoff is 'worth it'. 

Another assumption is that we don't want to drift from stdlib because the go folk are making reasonable sane changes to the DefaultTransport that we *want* in imgpkg.

nithinmy:
Not sure if I was doing it right, but I just tried to push a local branch and got this error -
```
08:38:00.829: [carvel-imgpkg] git -c credential.helper= -c core.quotepath=false -c log.showSignature=false push --progress --porcelain origin refs/heads/config-cli-responseheadertimeout:config-cli-responseheadertimeout --set-upstream
remote: Permission to vmware-tanzu/carvel-imgpkg.git denied to nithinmy.
fatal: unable to access 'https://github.com/vmware-tanzu/carvel-imgpkg.git/': The requested URL returned error: 403
```

pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -19,6 +20,8 @@ type RegistryFlags struct {
 	Password string
 	Token    string
 	Anon     bool
+
+	ResponseTimeout time.Duration
```
cppforlife: this should be ResponseHeaderTimeout. ResponseTimeout implies that it's a timeout for the entire response which isnt the case of how we are using this value.

pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -30,6 +33,8 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&r.Password, "registry-password", "", "Set password for auth ($IMGPKG_PASSWORD)")
 	cmd.Flags().StringVar(&r.Token, "registry-token", "", "Set token for auth ($IMGPKG_TOKEN)")
 	cmd.Flags().BoolVar(&r.Anon, "registry-anon", false, "Set anonymous auth ($IMGPKG_ANON)")
+
+	cmd.Flags().DurationVar(&r.ResponseTimeout, "registry-response-timeout", 30*time.Second, "Maximum time to allow a request to wait for a http response from the registry (ms|s|m|h)")
```
cppforlife: registry-response-timeout -> registry-response-header-timeout

pkg/imgpkg/cmd/registry_flags.go
DennisDenuto: ok done

pkg/imgpkg/cmd/registry_flags.go
DennisDenuto: ok done

PR #188, opened 2021-07-07
bug fix: ensure correct bundle is referenced when generating lock file

### LINKED ISSUE #159: when working with nested bundles, "imgpkg copy --lock-output" always returns digest for the "first" bundle, instead of bundle that was given in -b
**What steps did you take:**

```
imgpkg copy --lock staging.lock --to-repo ${TEST_REGISTRY}/imgpkg-debug-release --lock-output release.lock
```

assuming staging.lock points to a bundle that references nested bundles, `release.lock` (may) end up pointing to a nested bundle (instead of root bundle referenced by staging.lock) because writeLockOutput [1] picks last bundle that it finds within processed images.

[1] https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/pkg/imgpkg/cmd/copy.go#L160-L172

- imgpkg version (use `imgpkg --version`): 0.6.0, 0.7.0

---
Vote on this request

DennisDenuto:
Attempts to fix: https://github.com/vmware-tanzu/carvel-imgpkg/issues/159

### DIFF
--- a/pkg/imgpkg/cmd/copy.go
+++ b/pkg/imgpkg/cmd/copy.go
@@ -168,6 +168,19 @@ func (c *CopyOptions) writeLockOutput(bundleProcessedImage *ctlimgset.ProcessedI
 		if !ok {
 			panic(fmt.Errorf("Internal inconsistency: '%s' should be a bundle but it is not", bundleProcessedImage.DigestRef))
 		}
+	} else {
+		for _, item := range processedImages.All() {
+			plainImg := plainimage.NewFetchedPlainImageWithTag(item.DigestRef, item.UnprocessedImageRef.Tag, item.Image, item.ImageIndex)
+			bundle := bundle.NewBundleFromPlainImage(plainImg, registry)
+
+			ok, err := bundle.IsBundle()
+			if err != nil {
+				return fmt.Errorf("Check if '%s' is bundle: %s", item.DigestRef, err)
+			}
+			if ok {
+				return fmt.Errorf("Unable to determine correct root bundle to use for lock-output. hint: if copying from a tarball, try re-generating the tarball")
+			}
+		}
 	}
 
 	if c.LockOutputFlags.LockFilePath != "" {
--- a/pkg/imgpkg/cmd/copy.go
+++ b/pkg/imgpkg/cmd/copy.go
@@ -105,13 +105,13 @@ func (c *CopyOptions) Run() error {
 		imageSet := ctlimgset.NewImageSet(c.Concurrency, prefixedLogger)
 		tarImageSet := ctlimgset.NewTarImageSet(imageSet, c.Concurrency, prefixedLogger)
 
-		processedImages, err := tarImageSet.Import(c.TarFlags.TarSrc, importRepo, regWithProgress)
+		bundleProcessedImageRef, processedImages, err := tarImageSet.Import(c.TarFlags.TarSrc, importRepo, regWithProgress)
 		if err != nil {
 			return err
 		}
 
 		informUserToUseTheNonDistributableFlagWithDescriptors(levelLogger, c.IncludeNonDistributable, processedImagesMediaType(processedImages))
-		return c.writeLockOutput(processedImages, reg)
+		return c.writeLockOutput(bundleProcessedImageRef, processedImages, reg)
 
 	case c.isRepoSrc():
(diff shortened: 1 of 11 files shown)


DennisDenuto:
TODO:
Current WIP is a POC to ensure it 'fixes' the bug. Validation was mainly done via manual testing...

- [x] refactor - think about code change and whether it applies to the affected layers.
- [x] review error messages and ensure wording / phrasing is correct
- [x] change label constants from `main.bundle: true` to a namespace that communicates what the label is for.
- [x] backfill tests -> WIP is currently a proof of concept that I ran with manual tests to ensure the fix would work.
- [x] test to make sure that a 'newer' version of imgpkg that generates a tarball (with the modified manifest.json) still works with an 'older' version of imgpkg. Reading the code I think it will fine.

pkg/imgpkg/cmd/copy.go
```diff
@@ -146,28 +146,40 @@ func (c *CopyOptions) Run() error {
...
+			bundle := bundle.NewBundleFromPlainImage(plainImg, registry)
+
+			ok, err := bundle.IsBundle()
+			if err != nil {
+				return fmt.Errorf("Check if '%s' is bundle: %s", item.DigestRef, err)
+			}
+			if ok {
+				return fmt.Errorf("Unable to determine correct root bundle to use for lock-output. hint: if copying from a tarball, try re-generating the tarball")
+			}
 		}
```
cppforlife, line 183: im not sure i understand purpose of this code block.

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -23,7 +25,7 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, logger Logger) TarImageS
 	return TarImageSet{imageSet, concurrency, logger}
 }
 
-func (i TarImageSet) Export(foundImages *UnprocessedImageRefs, outputPath string, registry ImagesReaderWriter, imageLayerWriterCheck imagetar.ImageLayerWriterFilter) (*imagedesc.ImageRefDescriptors, error) {
+func (i TarImageSet) Export(bundleUnprocessedImageRef *UnprocessedImageRef, foundImages *UnprocessedImageRefs, outputPath string, registry ImagesReaderWriter, imageLayerWriterCheck imagetar.ImageLayerWriterFilter) (*imagedesc.ImageRefDescriptors, error) {
```
cppforlife, line 28: imageset package should not have any idea of what bundles are (its a lower level package that knows how to work with lower level primitives). how do we pull out such bundle related logic into packages (e.g. cmd package) that do/should know about bundles?

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -46,18 +48,50 @@ func (i TarImageSet) Export(foundImages *UnprocessedImageRefs, outputPath string
...
+			return nil, err
+		}
+	}
 
 	return ids, imagetar.NewTarWriter(ids, outputFileOpener, opts, i.logger, imageLayerWriterCheck).Write()
 }
 
 func (i *TarImageSet) Import(path string,
-	importRepo regname.Repository, registry ImagesReaderWriter) (*ProcessedImages, error) {
+	importRepo regname.Repository, registry ImagesReaderWriter) (*ProcessedImage, *ProcessedImages, error) {
```
cppforlife, line 62: same goes for this method (too low level to know about bundles).

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/k14s/imgpkg/pkg/imgpkg/imagetar"
 )
 
+const rootBundleLabelKey string = "root.bundle"
```
cppforlife: "root.bundle" -> "dev.carvel.imgpkg.copy.root-bundle"

cppforlife:
talked with @DennisDenuto. will do a refactor to move bundle related logic out of imagetar/imageset packages after release is cut (given its urgency)

pkg/imgpkg/cmd/copy.go
DennisDenuto: discussed with dk over zoom. it is to provide a friendly error to the user if a tarball is provided (with an older version of imgpkg that doesn't contain the labels to the root bundle).

The error nudges the user to re-generate the tarball with the newer version of imgpkg

PR #189, opened 2021-07-12
Refactor - remove bundle information from iamgedesc layer

DennisDenuto:
follow up refactor for: https://github.com/vmware-tanzu/carvel-imgpkg/issues/159

### DIFF
--- a/pkg/imgpkg/cmd/copy_repo_src_test.go
+++ b/pkg/imgpkg/cmd/copy_repo_src_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/k14s/imgpkg/pkg/imgpkg/bundle"
 	ctlimg "github.com/k14s/imgpkg/pkg/imgpkg/image"
+	"github.com/k14s/imgpkg/pkg/imgpkg/imagedesc"
 	"github.com/k14s/imgpkg/pkg/imgpkg/imageset"
 	"github.com/k14s/imgpkg/pkg/imgpkg/imagetar"
 	"github.com/k14s/imgpkg/pkg/imgpkg/lockconfig"
@@ -384,13 +385,18 @@ func TestToRepoBundleContainingANestedBundle(t *testing.T) {
 		subject.registry = fakeRegistry.Build()
 
 		destRepo := fakeRegistry.ReferenceOnTestServer("library/bundle-copy")
-		processedBundle, processedImages, err := subject.CopyToRepo(destRepo)
+		processedImages, err := subject.CopyToRepo(destRepo)
 		require.NoError(t, err)
 
 		require.Len(t, processedImages.All(), 4)
+
+		var processedBundle imageset.ProcessedImage
 		processedImageDigest := []string{}
 		for _, processedImage := range processedImages.All() {
 			processedImageDigest = append(processedImageDigest, processedImage.DigestRef)
+			if _, ok := processedImage.Labels[rootBundleLabelKey]; ok {
+				processedBundle = processedImage
+			}
 		}
 		assert.ElementsMatch(t, processedImageDigest, []string{
 			destRepo + "@" + bundleWithNestedBundle.Digest,
@@ -422,13 +428,18 @@ bundle:
 		subject.registry = fakeRegistry.Build()
 
 		destRepo := fakeRegistry.ReferenceOnTestServer("library/bundle-copy")
-		processedBundle, processedImages, err := subject.CopyToRepo(destRepo)
+		processedImages, err := subject.CopyToRepo(destRepo)
 		require.NoError(t, err)
 
 		require.Len(t, processedImages.All(), 4)
+
+		var processedBundle imageset.ProcessedImage
 		processedImageDigest := []string{}
 		for _, processedImage := range processedImages.All() {
 			processedImageDigest = append(processedImageDigest, processedImage.DigestRef)
+			if _, ok := processedImage.Labels[rootBundleLabelKey]; ok {
+				processedBundle = processedImage
+			}
 		}
 		assert.ElementsMatch(t, processedImageDigest, []string{
 			destRepo + "@" + bundleWithNestedBundle.Digest,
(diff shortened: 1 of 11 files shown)


pkg/imgpkg/imagedesc/image_ref_descriptors.go
```diff
@@ -28,8 +28,9 @@ type Registry interface {
 }
 
 type Metadata struct {
-	Ref regname.Reference
-	Tag string
+	Ref    regname.Reference
+	Tag    string
+	Labels map[string]interface{}
```
cppforlife: i think it might be worth making this map[string]string (follows more conventional labels/annotations signature within kubernetes ecosystem)

DennisDenuto: ok done

DennisDenuto: I think this is a breaking change...

if someone were to use a tarball generated with v0.14.0, the tarball has `nil` as the map value.

I *think* they might get a `panic` complaining that they are unable to convert to a string.

anyway, if that is true, the workaround would be to re-generate the tarball with the newer version of imgpkg.

pkg/imgpkg/imageset/unprocessed_image_refs.go
```diff
@@ -11,37 +11,50 @@ import (
 	regname "github.com/google/go-containerregistry/pkg/name"
 )
 
+const RootBundleLabelKey string = "dev.carvel.imgpkg.copy.root-bundle"
```
cppforlife: imageset does not know anything about root bundles. we should move this into a package that contains copy command (that's the only place where it's used afaik).

pkg/imgpkg/imageset/unprocessed_image_refs.go
```diff
@@ -11,37 +11,50 @@ import (
...
+
 type UnprocessedImageRef struct {
 	DigestRef string
 	Tag       string
+
+	Labels map[string]string
+}
+
+func (u UnprocessedImageRef) Key() string {
+	return u.DigestRef + u.Tag
```
cppforlife: just in case, lets add a separator / between the two since we are combining two unrelated pieces into a single data.

pkg/imgpkg/imageset/unprocessed_image_refs.go
```diff
@@ -11,37 +11,50 @@ import (
...
-	i.imgRefs[imgRef] = struct{}{}
+	i.imgRefs[imgRef.Key()] = imgRef
 }
 
 func (i *UnprocessedImageRefs) Length() int {
 	return len(i.imgRefs)
 }
 
 func (i *UnprocessedImageRefs) All() []UnprocessedImageRef {
+	i.Mutex.Lock()
```
cppforlife: while we are here, why do we need Mutex to be public here?

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -48,40 +46,32 @@ func (i TarImageSet) Export(bundleUnprocessedImageRef *UnprocessedImageRef, foun
...
 
 	reader := imagetar.NewTarReader(path)
 	imgOrIndexes, err := reader.Read()
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	rootBundleImageDesc, err := reader.FindByLabelKey(rootBundleLabelKey)
+	rootBundleImageDesc, err := reader.FindByLabelKey(RootBundleLabelKey)
```
cppforlife: TarImageSet should not know about RootBundleLabelKey.

pkg/imgpkg/imageset/unprocessed_image_refs.go
DennisDenuto: ok moved to the `cmd` package

pkg/imgpkg/imageset/unprocessed_image_refs.go
DennisDenuto: ok. done!

pkg/imgpkg/imageset/unprocessed_image_refs.go
DennisDenuto: for... transparency.... jk
made it private!

pkg/imgpkg/imageset/tar_image_set.go
DennisDenuto: ok good call. moved that logic out of tar_image_set.

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -384,13 +385,18 @@ func TestToRepoBundleContainingANestedBundle(t *testing.T) {
...
+		processedImages, err := subject.CopyToRepo(destRepo)
 		require.NoError(t, err)
 
 		require.Len(t, processedImages.All(), 4)
+
+		var processedBundle imageset.ProcessedImage
 		processedImageDigest := []string{}
 		for _, processedImage := range processedImages.All() {
 			processedImageDigest = append(processedImageDigest, processedImage.DigestRef)
+			if _, ok := processedImage.Labels[rootBundleLabelKey]; ok {
```
cppforlife, line 397: ill channel joao here, we should make a copy of rootBundleLabelKey for tests. otherwise tests do not catch breaking change if rootBundleLabelKey changes.

also i see that this file is under `package cmd`. we should move it to `cmd_test`.

DennisDenuto: Hmm, that was kinda intentional actually.

during test assertions of ensuring that the tarball contains the correct label, I do make a copy of that constant. See https://github.com/vmware-tanzu/carvel-imgpkg/pull/189/files/a114b01fb828548b594390c5739c3a98ecd66be5#diff-21cee4994b10fb08208390667e81d81f7e176e9b62caee769d3c42cc4a9cac42R910

That part of the test you have mentioned, digs into an internal constant to find a reference to make a specific type of assertion. which is that the digest ref is correct.

If the label constant were to change accidentally, a test would fail due to the tarball not containing the correct label. 

> also i see that this file is under package cmd. we should move it to cmd_test.

Every test in that package is currently in `cmd` not `cmd_test`. That is a larger change that I think should be done in a separate PR. There is dependencies in those tests that access private members.

Tell Joao I say hi btw.

cppforlife: 👍 i think separating that into a dedicated PR is prolly good.

cppforlife:
double checked changes via `git diff 82ca9f1...head`

PR #194, opened 2021-07-19
PlainImage does not allow ImageIndex

DennisDenuto:
A/C

for `copy`:

When specifying an image, but a user provides an `ImageIndex` `imgpkg copy --image imageindex` then an error message will be returned explaining that only plain images (not image indexes) are accepted.

When specifying an image lock file `imgpkg copy --lock images.lock` and that images.lock file contains an ImageIndex, then an error message will be returned explaining that only plain images (not image indexes) are accepted.

for `pull`:
When specifying an image, but a user provides an `ImageIndex` then an error message will be returned explaining that only plain images (not image indexes) are accepted.

## EDIT:

Will allow copying an image index but continue to return an error message if pulling an imageindex

### Out of scope for this story:

Updating the output during a copy to include the 'image reference' of the image belonging to the index that is copied.

Currently it will output the 'image index reference', (but not the images that belong to it) 


DennisDenuto:
A/C

for `copy`:

When specifying an image, but a user provides an `ImageIndex` `imgpkg copy --image imageindex` then an error message will be returned explaining that only plain images (not image indexes) are accepted.

When specifying an image lock file `imgpkg copy --lock images.lock` and that images.lock file contains an ImageIndex, then an error message will be returned explaining that only plain images (not image indexes) are accepted.


for `pull`:
When specifying an image, but a user provides an `ImageIndex` then an error message will be returned explaining that only plain images (not image indexes) are accepted.

pkg/imgpkg/cmd/pull.go
```diff
@@ -89,6 +86,15 @@ func (po *PullOptions) Run() error {
 
 	case len(po.ImageFlags.Image) > 0:
 		plainImg := plainimage.NewPlainImage(po.ImageFlags.Image, reg)
+		isImage, err := plainImg.IsImage()
+		if err != nil {
+			return err
+		}
+
+		if !isImage {
+			return fmt.Errorf("Unable to pull non-images, such as ImageIndexes. (hint: provide a specific digest to the image instead)")
```
cppforlife: nit: i think it's probably worth casing "ImageIndex" as "image index"

pkg/imgpkg/plainimage/plain_image.go
```diff
@@ -9,27 +9,34 @@ import (
...
-	fetchedIndex regv1.ImageIndex
 }
 
-func NewPlainImage(ref string, imagesMetadata ctlimg.ImagesMetadata) *PlainImage {
-	return &PlainImage{ref: ref, registry: imagesMetadata}
-}
+func NewPlainImage(ref string, imgDescriptor ImagesDescriptor) *PlainImage {
+	parsedRef, err := regname.ParseReference(ref, regname.WeakValidation)
+	if err != nil {
+		panic(fmt.Sprintf("Expected valid Tag Ref: %s", err))
```
cppforlife: if we panic here i think `imgpkg pull -i :123` would panic instead of previous behaviour of returning an error from Fetch. i think we should keep setting parsedRef in fetch.

pkg/imgpkg/plainimage/plain_image.go
```diff
@@ -86,28 +92,31 @@ func (i *PlainImage) Fetch() (regv1.Image, error) {
...
+	if !imgDescriptor.MediaType.IsImage() {
+		if imgDescriptor.MediaType.IsIndex() {
+			index, err := imgDescriptor.ImageIndex()
+			if err != nil {
+				return nil, notAnImageError{imgDescriptor.MediaType}
+			}
+			digest, err := index.Digest()
+			if err != nil {
+				return nil, notAnImageError{imgDescriptor.MediaType}
+			}
```
cppforlife: in both of these error cases, should we just be returning underlying error instead of notAnImageError, or at least including underyling error in the resulting error message?

pkg/imgpkg/plainimage/plain_image.go
```diff
@@ -86,28 +92,31 @@ func (i *PlainImage) Fetch() (regv1.Image, error) {
...
+			i.parsedDigest = digest.String()
```
cppforlife: i believe imgDescriptor already has a digest, so no need to check if it's an index and go down that path.

pkg/imgpkg/cmd/pull.go
DennisDenuto: ok done. changed to "image index"

pkg/imgpkg/plainimage/plain_image.go
DennisDenuto: true. ah so much nicer.

pkg/imgpkg/plainimage/plain_image.go
DennisDenuto: true. true...

Ok, i went a bit back and forth between moving that error checking back into the `Fetch` vs returning an error from `NewPlainImage`

My main thought was we have 2 constructors, `NewPlainImage` and `NewFetchedPlainImageWithTag`. where `NewFetchedPlainImageWithTag` does panic if the provided ref is incorrect. Although *today*, due to 'when' and 'how' it is used, should *never* be in a position to be provided an incorrect ref (i think). Although we *could* introduce user validation around the `-i` flag and when reading image refs from the ImagesLock file to ensure that the ref passed into `NewPlainImage` is valid (just like its sibling NewFetchedPlainImageWithTag)

Went with removing the panic from `NewPlainImage` and moving the error back into `Fetch`.

However, I did add one more additional check in `NewFetchedPlainImageWithTag` to ensure that an image *is* in fact provided (and not nil), otherwise when `Fetch` is called, it won't return the fetched image, and instead the digest ref provided to the sibling constructor will be 'clobbered' by the 'unparsedRef' (which will be an empty string).

cppforlife: got it. 👍

PR #198, opened 2021-07-20
improve --help short description

DennisDenuto:
Removes references to "Docker images".

This PR attempts to make the 'front page' of the help command a bit more useful.

### DIFF
--- a/pkg/imgpkg/cmd/imgpkg.go
+++ b/pkg/imgpkg/cmd/imgpkg.go
@@ -28,7 +28,7 @@ func NewDefaultImgpkgCmd(ui *ui.ConfUI) *cobra.Command {
 func NewImgpkgCmd(o *ImgpkgOptions) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:               "imgpkg",
-		Short:             "imgpkg stores files as Docker images",
+		Short:             "imgpkg creates an immutable OCI artifact suitable for containing configuration and images used in that configuration",
 		SilenceErrors:     true,
 		SilenceUsage:      true,
 		DisableAutoGenTag: true,


pkg/imgpkg/cmd/imgpkg.go
```diff
@@ -28,7 +28,7 @@ func NewDefaultImgpkgCmd(ui *ui.ConfUI) *cobra.Command {
 func NewImgpkgCmd(o *ImgpkgOptions) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:               "imgpkg",
-		Short:             "imgpkg stores files as Docker images",
+		Short:             "imgpkg creates an immutable OCI artifact suitable for containing configuration and images used in that configuration",
```
cppforlife, line 31: the word "configuration" seems to be overly skewed towards kubernetes use case. i think "files" is more generic. images do not also have to be used in that configuration. it could be any set of images.

DennisDenuto: hehe I did originally have the word kubernetes, and decided to remove it to make it more generic... 

> overly skewed towards kubernetes use case

It *is* predicated with the phrase 'it is suitable for configuration'. I was hoping it would come off as a soft opinion on our biggest use case, and not a hard rule for folk to follow.

However, using "files".... That isn't how we are communicating the tool in our docs though.

Here are some headlines from carvel.dev

"Package, distribute, and relocate your Kubernetes configuration and dependent OCI images as one OCI artifact"

"Bundle and relocate application configuration (with images) via Docker registries. Be assured app contents are immutable."

^ would be nice to change the phrase "Docker registries". 

Also, at risk of bike shedding here, if you reply with just "use files" I would `s/configuration/files`

cppforlife:
changed in https://github.com/vmware-tanzu/carvel-imgpkg/commit/60dae93db3d391f17cd5ad8e88323950515ba7f6

Issue #199, opened 2021-07-21
imgpkg copy should be re-runnable for registries that disallow mutating tags


DennisDenuto:
**Describe the problem/challenge you have**

Based on the following slack conversation: https://kubernetes.slack.com/archives/CH8KCCKA5/p1626878622031300

There appears to be errors when re-running `imgpkg copy` with the same bundle, targeting a registry that does not allow mutating tags.

Error received:
```
TAG_INVALID: The image tag 'sha256-ABCDEF.image-locations.imgpkg' already exists and cannot be overwritten because the repository is immutable
```

**Describe the solution you'd like**
Ensure re-running `imgpkg copy` with the same input works on registries that dis-allow mutating tags.

**Anything else you would like to add:**
linked thread was using version: v0.14.0

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

gabegorelick:
> There appears to be errors when re-running imgpkg copy with the same bundle, targeting a registry that does not allow mutating tags.

One (big) caveat: it doesn't appear to be a consistent issue. Much of the time, rerunning `imgpkg copy` works fine. 

It's possible there is some kind of non-determinism going into the image-locations generation, but there doesn't seem to be a good way to grab the `image-locations.yml` that `imgpkg` is attempting to upload to confirm this.

DennisDenuto:
Hi @gabegorelick 

I have spent some time trying to get oci-location to 'mutate' across runs. But I have been unable to do so.

Check out this branch for a e2e test https://github.com/vmware-tanzu/carvel-imgpkg/tree/try-find-oci-location-mutate-exploration

Are you able to provide more information:
- bundle / nested-bundle structure
-- images.yml for each bundle / nested-bundle


In the meanwhile, I have submitted a [PR](https://github.com/vmware-tanzu/carvel-imgpkg/pull/205) in an attempt to prevent saving an oci-location image if it already exists

DennisDenuto:
From the slack convo:

We were provided with the following info https://kubernetes.slack.com/archives/CH8KCCKA5/p1627569740185300?thread_ts=1626878622.031300&cid=CH8KCCKA5:

```
---
apiVersion: imgpkg.carvel.dev/v1alpha1
[...]

PR #205, opened 2021-07-28
Attempt to prevent location-oci from mutating un-intentionally

DennisDenuto:
Attempts to address https://github.com/vmware-tanzu/carvel-imgpkg/issues/199

I was unable to pin-point where / how a location-oci might mutate across runs.

This PR is a 'defensive' measure to try and prevent mutations from occurring in the future. Really to prevent issues for folk who are using a registry with immutable tags.

- does not handle concurrency race conditions i.e. more than 1 imgpkg copy from
running at the same time might still mutate an oci-location image.
- introduces a slight perf penalty. although the location-oci should be
'small' in size

### DIFF
--- a/pkg/imgpkg/bundle/bundle.go
+++ b/pkg/imgpkg/bundle/bundle.go
@@ -117,11 +117,17 @@ func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesM
 
 	logger.Debugf("creating Locations OCI Image\n")
 	// Using NewNoopUI because we do not want to have output from this push
-	err = NewLocations(logger).Save(reg, destinationRef, locationsCfg, goui.NewNoopUI())
-	if err != nil {
-		return err
+
+	locations := NewLocations(logger)
+	_, err = locations.Fetch(reg, destinationRef)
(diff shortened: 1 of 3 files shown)


cppforlife:
i think we should start a bit conservative -- lets detect when this error happens (error is raised and we catch it), and include as part of error existing and new images locations content.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -117,11 +117,35 @@ func (o *Bundle) NoteCopy(processedImages *imageset.ProcessedImages, reg ImagesM
...
+		return err
+	}
+	existingLocationCfgUploadBytes, debugLocationErr := existingLocationsCfg.AsBytes()
+	if debugLocationErr != nil {
+		return err
+	}
+
+	return fmt.Errorf(`Unable to write location oci: %s
+location oci already uploaded contents: %s
+attempted to write oci contents: %s`,
```
cppforlife: nit: in general its handy to add some deliminator like quotes or >><< around unknown multi line text so that we now within the error if there is any suspicious whitespace etc.

DennisDenuto: ok done

PR #207, opened 2021-07-29
Minor renames in Bundle class

cppforlife:


pkg/imgpkg/bundle/bundle.go
```diff
@@ -40,49 +40,56 @@ type Bundle struct {
...
 func NewBundleFromPlainImage(plainImg *plainimg.PlainImage, imagesMetadata ImagesMetadata) *Bundle {
-	return &Bundle{plainImg: plainImg, imgRetriever: imagesMetadata, imagesLockReader: &singleLayerReader{}, imagesRef: map[string]ImageRef{}}
+	return &Bundle{plainImg: plainImg, imgRetriever: imagesMetadata,
+		imagesLockReader: &singleLayerReader{}}
 }
 
 func NewBundleWithReader(ref string, imagesMetadata ImagesMetadata, imagesLockReader ImagesLockReader) *Bundle {
-	return &Bundle{plainImg: plainimg.NewPlainImage(ref, imagesMetadata), imgRetriever: imagesMetadata, imagesLockReader: imagesLockReader, imagesRef: map[string]ImageRef{}}
+	return &Bundle{plainImg: plainimg.NewPlainImage(ref, imagesMetadata),
+		imgRetriever: imagesMetadata, imagesLockReader: imagesLockReader}
```
DennisDenuto, line 61: Whats the rationale for not initializing `imagesRef` in the constructor anymore?

Is this a defensive measure? to only permit writing to the map when `bundle.buildAllImagesLock` is called, and to panic *if* anywhere else were to try and update the 'cached' image refs before first calling that function?

DennisDenuto (changes requested):

pkg/imgpkg/bundle/bundle.go
cppforlife: yup, exactly that. we also dont want to reuse it every time you call AllImagesLock.

PR #208, opened 2021-07-29
add --image-is-bundle-check=bool flag to allow fetching bundles as plain images via imgpkg pull

### LINKED ISSUE #206: add --image-is-bundle-check=true to disable failure when trying "imgpkg pull -i <bundle>"

i'd like to use plain image functionality to fetch bundle, so that i can see exact bits of the bundle (without any rewriting of images.yml, etc.)

`imgpkg pull -i my-bundle:v1` would continue result in error
`imgpkg pull -i my-bundle:v1 --image-is-bundle-check=true` would succeed

---
Vote on this request

cppforlife:
fixes https://github.com/vmware-tanzu/carvel-imgpkg/issues/206

works:

```
imgpkg pull -i k8slt/kc-e2e-test-repo@sha256:ddd93b67b97c1460580ca1afd04326d16900dc716c4357cade85b83deab76f1c -o /tmp/lol --image-is-bundle-check=false
```

### DIFF
--- a/pkg/imgpkg/cmd/pull.go
+++ b/pkg/imgpkg/cmd/pull.go
@@ -18,6 +18,7 @@ type PullOptions struct {
 	ui ui.UI
 
 	ImageFlags           ImageFlags
+	ImageIsBundleCheck   bool
 	RegistryFlags        RegistryFlags
 	BundleFlags          BundleFlags
 	LockInputFlags       LockInputFlags
@@ -42,6 +43,7 @@ func NewPullCmd(o *PullOptions) *cobra.Command {
   imgpkg pull -i repo/app1-image -o /tmp/app1-image`,
 	}
 	o.ImageFlags.Set(cmd)
+	cmd.Flags().BoolVar(&o.ImageIsBundleCheck, "image-is-bundle-check", true, "Error when image is a bundle (disable pulling bundles via -i)")
 	o.RegistryFlags.Set(cmd)
 	o.BundleFlags.Set(cmd)
 	o.BundleRecursiveFlags.Set(cmd)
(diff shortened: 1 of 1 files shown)


DennisDenuto:
The `--image-is-bundle-check` is starting to blur the difference between the `-i` and `-b` functionality. I think the blurring is due to the `--image` having the concept of a 'bundle' being introduced.

From a ux perspective this adds to the `pull` command another flag that only works in the context of a previous flag being set. This one isn't a *massive* issue, but it can add to confusion. It is something i'm noticing, and perhaps one day if enough divergence is encountered, a splitting of sub-commands might be required. i.e. `imgpkg pull image <flags>` and `imgpkg pull bundle <flags>`

for e.g. the `--recursive` flag only makes sense when pulling a bundle (not an image)
similarly the `--image-is-bundle-check` flag only makes sense when pulling an image (not a bundle)

I'm thinking that, if the goal is to pull a bundle, without re-writing the images.yml, then that is a toggle for the `--bundle` flag and not the `--image` flag. That way, bundle logic belongs to the bundle flag. 

So, instead of `--image-is-bundle-check`, a `--flag-to-communicate-that-images-yml-are-not-rewritten-when-pulled` that works on the `--bundle` flag. This (to me) helps with the blurring, but still introduces a flag that only works in the context of `-b` and not `-i`.


Note:
An assumption I had while suggesting the above is: We do not want to *completely* drop the the validation logic in `--images` to [assert that the image isn't a bundle](https://github.com/vmware-tanzu/carvel-imgpkg/blob/3ba94c8c804ffb778fb2fdc50426bb78b589ac96/pkg/imgpkg/cmd/pull.go#L106).

cppforlife:
> I'm thinking that, if the goal is to pull a bundle, without re-writing the images.yml, then that is a toggle for the --bundle flag and not the --image flag. That way, bundle logic belongs to the bundle flag.

goal is not pull a bundle. the goal is to pull an image _that_ happens to be a bundle. arguably we should get rid of bundle check entirely from pull -i but that might just save some folks from mistakenly using -i where they wanted to use -b.

in other words i should be able to use pull -i against any artifact in the registry regardless whether it's a bundle, runnable container image, image made via imgpkg, etc.

DennisDenuto:
Spoke over zoom. We think this flag is being set at the correct 'layer'. Meaning, this is a low level image interaction and hence doesn't belong to a higher level abstraction (-b flag)

Also, if we decide to rename the flag, it *should* be trivial to do so.

Issue #154, opened 2021-06-02
Incorrect parsing of image URL provided in the command line


joaopapereira:
**What steps did you take:**
Try to pull a bundle while providing the protocol (e.g. `https://`, `http://`) as part URL fails with a strange error

**What happened:**
```
imgpkg pull -b https://index.docker.io/k8slt/demo-repo@sha256:aef6d96cde8e41fbdd67202106c89a5c55bd0cd47398de619a60a4ae06a3e388  -o /tmp/lol
Error: Checking if image is bundle: Collecting images: Working with https://index.docker.io/k8slt/demo-repo@sha256:aef6d96cde8e41fbdd67202106c89a5c55bd0cd47398de619a60a4ae06a3e388: Get "https://https/v2/": dial tcp: lookup https on 10.21.82.100:53: no such host
```

**What did you expect:**
The bundle to be pulled. If you execute without the protocol is works
```
imgpkg pull -b https://index.docker.io/k8slt/demo-repo@sha256:aef6d96cde8e41fbdd67202106c89a5c55bd0cd47398de619a60a4ae06a3e388  -o /tmp/lol
Pulling bundle 'index.docker.io/k8slt/demo-repo@sha256:aef6d96cde8e41fbdd67202106c89a5c55bd0cd47398de619a60a4ae06a3e388'
  Extracting layer 'sha256:34de6e22982f89d95517c1cdb76e4d81f6a0735c78f89c56ab7a627b785b0aad' (1/1)

Locating image lock file images...
One or more images not found in bundle repo; skipping lock file update

Succeeded
```

**Anything else you would like to add:**
[Additional information that will assist in solving the issue.]

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.8.0
- Docker registry used (e.g. `Docker HUB`): docker hub, but is unrelated
- OS (e.g. from `/etc/os-release`): MacOS

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help to work on this issue.

aaronshurley:
_I think_ we can consider this issue `carvel-accepted` because we have an undesirable result at the moment by having an unhelpful error message. However, we have a few potential solutions that we could considder so that we can agree on a way forward. A couple of options that come to mind:

[...]

PR #214, opened 2021-08-04
warn on protocol: want to check if any tests fail if we panic and bail

joe-kimmel-vmw:


joe-kimmel-vmw:
just leaving a note that the prior art/discussion for this is here: https://github.com/vmware-tanzu/carvel-imgpkg/issues/154

and maybe the best would be to find the corresponding discussion in the underlying library and see what they think.

pkg/imgpkg/cmd/imgpkg.go
```diff
@@ -64,6 +66,21 @@ func NewImgpkgCmd(o *ImgpkgOptions) *cobra.Command {
...
+		bundleFlag := cmd.Flag("bundle")
+		if bundleFlag != nil && protocolMatcher.MatchString(bundleFlag.Value.String()) {
+			return fmt.Errorf("bundle flag %v starts with protocol: remove http(s):// from bundle", bundleFlag.Value)
+		}
+
+		imageFlag := cmd.Flag("image")
+		if imageFlag != nil && protocolMatcher.MatchString(imageFlag.Value.String()) {
+			return fmt.Errorf("image flag %v starts with protocol: remove http(s):// from image", imageFlag.Value)
+		}
+		return nil
```
joe-kimmel-vmw: not sure if there's a third example of bundle/image flag that we should check,  if so obvs should extract the pattern to one more func and just call it for each flag. 

also could get fancier w. regexps / protocols ,  e.g. could match `^.{1,10}://` to catch any protocol in case somebody has their bundle sitting on a gopher server or something.  (e.g. FTP seems vaguely possible).   If we did this then seems reasonable to use a capture group or something to show exactly what was matched.

pkg/imgpkg/registry/registry.go
```diff
@@ -74,7 +75,18 @@ func NewRegistry(opts Opts, regOpts ...regremote.Option) (Registry, error) {
 	}, nil
 }
 
+func validateRef(ref regname.Reference) error {
```
cppforlife: lets make this a private method on Registry class (nothing else should be using this method). you can use `func (Registry)` to even skip specifying receiver name. also lets move this internal/misc function to the end of the file.

pkg/imgpkg/registry/registry.go
```diff
@@ -74,7 +75,18 @@ func NewRegistry(opts Opts, regOpts ...regremote.Option) (Registry, error) {
...
+	protocolMatcher := regexp.MustCompile(`^(https?://)`)
```
cppforlife: nits:
- regular double quotes (since there is nothing special inside the string)
- no need to wrap in paranthesis
- use \A to start from content instead of ^
- typically it's MustCompile decl is set at the package level so that it can blow up at program startup if regexp is wrong.

pkg/imgpkg/registry/registry.go
```diff
@@ -164,6 +191,9 @@ func (r Registry) Index(ref regname.Reference) (regv1.ImageIndex, error) {
 }
 
 func (r Registry) WriteIndex(ref regname.Reference, idx regv1.ImageIndex) error {
+	if err := validateRef(ref); err != nil {
```
cppforlife: also add to WriteTag

pkg/imgpkg/registry/registry.go
```diff
@@ -74,7 +75,18 @@ func NewRegistry(opts Opts, regOpts ...regremote.Option) (Registry, error) {
 	}, nil
 }
 
+func validateRef(ref regname.Reference) error {
+	protocolMatcher := regexp.MustCompile(`^(https?://)`)
+	if match := protocolMatcher.FindString(ref.String()); len(match) > 0 {
+		return fmt.Errorf("Reference %v should not include %v protocol prefix", ref, match)
```
cppforlife: use `'%s'` (with single quotes) instead of %v

pkg/imgpkg/registry/registry.go
joe-kimmel-vmw: - once i swapped in `\A` it wanted the backtick-string back, but otherwise i think i got it.

pkg/imgpkg/registry/registry.go
```diff
@@ -244,3 +269,12 @@ func newHTTPTransport(opts Opts) (*http.Transport, error) {
 
 	return clonedDefaultTransport, nil
 }
+
+var protocolMatcher *regexp.Regexp = regexp.MustCompile(`\Ahttps?://`)
```
cppforlife: nit: i dont believe you need to specify a type (*regexp.Regexp)

pkg/imgpkg/registry/registry.go
```diff
@@ -244,3 +269,12 @@ func newHTTPTransport(opts Opts) (*http.Transport, error) {
...
+
+func (Registry) validateRef(ref regname.Reference) error {
+	if match := protocolMatcher.FindString(ref.String()); len(match) > 0 {
+		return fmt.Errorf("Reference '%s' should not include %v protocol prefix", ref, match)
```
cppforlife: nit: lets turn match into `'%s'` as well

Issue #138, opened 2021-05-03
Add automated testing against well known registries on commits and releases


joaopapereira:
### Acceptance Criteria:
**As a contributor to imgpkg**
I want to have confidence that my changes did not introduce inconsistent behavior on different registries.

**When a new commit is pushed to imgpkg**, I expect to see github actions running integration tests on:
- [x] Harbor
- [x] GCR

**When a new release of imgpkg is cut**, I expect to see github actions running integration tests on:
- [x] Harbor
- [x] GCR
- [ ] https://cloud.google.com/artifact-registry
- [ ] DockerHub

**Notes:**
- Since DockerHub has rate limiting, running the integration tests only during a release is sufficient.
- For the integration with harbor, Joao found that we could stand up an instance using https://github.com/bitnami/bitnami-docker-harbor-portal/blob/master/docker-compose.yml, so maybe we could just start a container to do this.
- **Feel free to tackle this work all at once, or one registry at a time using 3 different PRs.**
  - in fact, if supporting GCR and DockerHub is not "just" an incremental investment, consider whether it makes sense to tackle those registries, later.

👇🏻 Original request👇🏻 

---

- **Describe the problem/challenge you have**
As a user, I want to be sure that `imgpkg` works with different registries

---
**Describe the solution you'd like**
When testing `imgpkg` using Github action, we should try to test against a local registry and against more widely spread registries, like gcr, harbor, and dockerhub.

**Anything else you would like to add:**
I spun up this issue from #131, thanks for the idea @DennisDenuto 
For this issue, we should do the following:
- [x] Run integration tests against a gcr instance
- [x] Run integration tests against an Harbor instance
  - Is there a Harbor instance that we can use?
  - Should we instantiate a Harbor instance and use it for testing?
  -  Can we run Harbor as a container similar to what we are doing with the registry:2?
- [ ] Run integration tests against a dockerhub, we should be careful about this due to rate limiting. Maybe we could consider running these tests as part of a Release process

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

joaopapereira:
[...]

PR #219, opened 2021-08-18
Test with gcr registry

DennisDenuto:
Addresses #138

Add GH Action to run tests against GCR
 
- After tests are run, the images created by the tests are deleted.
-- IAM was configured by following: https://cloud.google.com/container-registry/docs/access-control

See test run here:
https://github.com/vmware-tanzu/carvel-imgpkg/runs/3364818396

*** Prior to Merging ***
~~The "GCR e2e" environment needs to be created and configured with a secret `GCR_PASSWORD`~~

### DIFF
--- a/.github/workflows/gh-test-external-registry.yml
+++ b/.github/workflows/gh-test-external-registry.yml
@@ -91,11 +91,20 @@ jobs:
           # pull registry for e2e tests that require a locally running docker registry. i.e. airgapped env tests
           docker pull registry:2
 
+          function cleanup {
+            set +e
+            {
+              gcloud auth activate-service-account --key-file <(echo "$GCR_PASSWORD")
+              gcloud config set project cf-k8s-lifecycle-tooling-klt > /dev/null
+              gcloud container images list --filter=$GITHUB_RUN_ID | while read img; do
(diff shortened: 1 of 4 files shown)


DennisDenuto:
There appears to be an incompatibility between the air-gap workflow (when using non-distributable layers) and GCR. 

See the failure around copy shown in: https://github.com/vmware-tanzu/carvel-imgpkg/runs/3364818396

Short version is: gcr seems ok to commit a manifest that references `application/vnd.docker.image.rootfs.foreign.diff.tar.gzip` layers that are not uploaded as blobs.

However, if the layer is of type `application/vnd.oci.image.layer.nondistributable.v1.tar` or `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip` then GCR errors out.

I've submitted a [bug to the GCR issue tracker](https://issuetracker.google.com/issues/197146608) and will probably modify the e2e accordingly.

I've also submitted a [bug to dockerhub](https://github.com/docker/hub-feedback/issues/2132), as it has a similar issue when the media type is foreign layers.

DennisDenuto:
See run with GCR as the registry: https://github.com/vmware-tanzu/carvel-imgpkg/runs/3386269802?check_suite_focus=true

PR #222, opened 2021-08-25
Stop modifying local filesystem when running gcloud e2e test

### LINKED ISSUE #57: e2e tests are messing with global docker config.json
this is surprising as a user of e2e tests since changes to .docker/config.json affects my other operations on the computer. can we find an alternative way of testing whatever needs to be tested?

DennisDenuto:
Attempts to fix #57 

- Uses a temp directory to contain the docker config.json

### DIFF
--- a/test/e2e/gcloud_expired_token_test.go
+++ b/test/e2e/gcloud_expired_token_test.go
@@ -4,14 +4,12 @@
 package e2e
 
 import (
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"testing"
-	"time"
 
 	"github.com/k14s/imgpkg/test/helpers"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


test/e2e/gcloud_expired_token_test.go
```diff
@@ -45,49 +44,49 @@ images:
...
+					"gcr.io": "gcloud-race-condition-db-error"
+			}
+		}`), os.ModePerm)
 
-	err = ioutil.WriteFile(dockerConfigPath, dockerConfigJSONContents, os.ModePerm)
 	require.NoError(t, err)
 
-	// restore docker config
+	// Retrieve the docker image
+	exec.Command("docker", "pull", "ubuntu:21.04").Run()
```
cppforlife, line 86: what is the purpose of this command? also same for docker volume rm

DennisDenuto: It is to cache the image before the `gcloud-race-condition-db-error` plugin is called.

The `gcloud-race-condition-db-error` runs a docker command (using the ubuntu:21.04) image. If it isn't cached, then that plugin will download that image (which takes time), and the keychain will timeout/fail. (We want it to fail for a different reason)

DennisDenuto: `docker volume rm` is run during the cleanup phase of the test. It is to remove a docker volume created by the test.

cppforlife: @DennisDenuto i think we should add this as a comment on top of the docker pull command. no way somebody would get this context from reading that line.

DennisDenuto: done

PR #225, opened 2021-08-27
Require godocs

DennisDenuto:
Inspired from https://github.com/vmware-tanzu/carvel-ytt/pull/476

To generate godocs run:
```
 godoc -http localhost:6060
```

then in the browser goto:
http://localhost:6060/pkg/github.com/k14s/imgpkg/


joe-kimmel-vmw:
LGTM

PR #228, opened 2021-08-31
Fix grammar

seemiller:
Replace 2 instances of `It it` with `It`

Signed-off-by: Nicholas Seemiller <nseemiller@vmware.com>


DennisDenuto:
LGTM

I'm going to merge this in. The harbor tests has failed however that is due to flakiness from TanzuNet and not related to this change.

The local-registry and GCP tests have passed.

PR #242, opened 2021-09-10
Add dev docs for godoc process

cari-lynn:
Added docs to the main readme. I think this could also fit in a new file under docs/dev.md.


DennisDenuto:
Thanks @cari-lynn !!

LGTM. 

Also tested the godoc commands and it renders and works.

Issue #235, opened 2021-09-02
reconsider how registry matching is done in env keychain


cppforlife:
**Describe the problem/challenge you have**

currently we only match by hostname. to make imgpkg more interoperable with k8s we may want to support more flexible matching:
Fix soon:
- [x] support ignoring of http/https scheme
  - https://github.com/vmware-tanzu/carvel-imgpkg/issues/234
- [x] ignore v1/v2 suffix on dockerhub? 

Requires more investigation:
- [x] support matching by globs? k8s cred provider related code has some matching
- [x] support matching by repository path (e.g. support credentials for same registry but under different paths)
    - https://github.com/vmware-tanzu/carvel-imgpkg/issues/226

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

DennisDenuto:
Closing. This has been resolved in https://github.com/vmware-tanzu/carvel-imgpkg/releases/tag/v0.18.0

Issue #234, opened 2021-09-02
Make IMGPKG_REGISTRY_HOSTNAME more permissive


DennisDenuto:
**Describe the problem/challenge you have**
- when setting a registry hostname via env variables (`IMGPKG_REGISTRY_HOSTNAME`) containing a `scheme` such as `http` or `https`, authentication is failing with a validation error:
```
Parsing registry hostname: registries must be valid RFC 3986 URI authorities: https://<registry> (e.g. gcr.io, index.docker.io)
```

**Describe the solution you'd like**
Allow specifying a scheme when setting the `IMGPKG_REGISTRY_HOSTNAME` env variable

**Anything else you would like to add:**
Slack: https://kubernetes.slack.com/archives/CH8KCCKA5/p1630593329071400

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

DennisDenuto:
Closing. This has been resolved in https://github.com/vmware-tanzu/carvel-imgpkg/releases/tag/v0.18.0

PR #220, opened 2021-08-24
Use k8s keychain to auth against IaaS

DennisDenuto:

PR #245, opened 2021-09-14
Make HOSTNAME matching more permissive

DennisDenuto:
Related to: 
https://github.com/vmware-tanzu/carvel-imgpkg/issues/234
https://github.com/vmware-tanzu/carvel-imgpkg/issues/235

`IMGPKG_REGISTRY_HOSTNAME` matching now allows:
- schemes (http/https)
- v1 and v2 in the path to be the equivalent to just passing the hostname
- glob patterns in the HOSTNAME (currently wildcards are not supported in ports and paths)

# Before merging

to avoid merge difficulty, this PR should be merged after https://github.com/vmware-tanzu/carvel-imgpkg/pull/220

TODO:

- [ ] Update docs to reflect new behavior of `IMGPKG_REGISTRY_HOSTNAME`

### DIFF
--- a/pkg/imgpkg/registry/keychain_test.go
+++ b/pkg/imgpkg/registry/keychain_test.go
@@ -240,6 +240,58 @@ func TestAuthProvidedViaEnvVars(t *testing.T) {
 		}), auth)
 	})
 
+	for i, validHostname := range []string{
+		"localhost:9999",
+		"http://localhost:9999",
+		"https://localhost:9999",
+		"localhost:9999/v1/",
+		"localhost:9999/v2/",
+	} {
+		t.Run(fmt.Sprintf("IMGPKG_HOSTNAME %d", i), func(t *testing.T) {
+			envVars := []string{
+				"IMGPKG_REGISTRY_USERNAME=user",
+				"IMGPKG_REGISTRY_PASSWORD=pass",
+				fmt.Sprintf("IMGPKG_REGISTRY_HOSTNAME=%s", validHostname),
+			}
+
+			keychain, err := registry.Keychain(auth.KeychainOpts{}, func() []string { return envVars })
+			assert.NoError(t, err)
+
+			resource, err := name.NewRepository("localhost:9999/imgpkg_test")
+			assert.NoError(t, err)
+
+			auth, err := keychain.Resolve(resource)
+			assert.NoError(t, err)
+
+			assert.Equal(t, authn.FromConfig(authn.AuthConfig{
+				Username: "user",
+				Password: "pass",
+			}), auth)
+		})
+	}
+
+	for i, invalidHostname := range []string{
+		"http://[::1]:namedport", // rfc3986 3.2.3
+		"http://[%10::1]",        // no %xx escapes in IP address
+		"http://%41:8080/",       // not allowed: % encoding only for non-ASCII
+	} {
+		t.Run(fmt.Sprintf("IMGPKG_HOSTNAME %d", i), func(t *testing.T) {
+			envVars := []string{
+				"IMGPKG_REGISTRY_USERNAME=user",
+				"IMGPKG_REGISTRY_PASSWORD=pass",
+				fmt.Sprintf("IMGPKG_REGISTRY_HOSTNAME=%s", invalidHostname),
+			}
+
+			keychain, err := registry.Keychain(auth.KeychainOpts{}, func() []string { return envVars })
+			assert.NoError(t, err)
+
+			resource, err := name.NewRepository("localhost:9999/imgpkg_test")
+			assert.NoError(t, err)
+
+			_, err = keychain.Resolve(resource)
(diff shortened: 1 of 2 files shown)


pkg/imgpkg/registry/keychain.go
```diff
@@ -49,7 +50,12 @@ func (k *envKeychain) Resolve(target regauthn.Resource) (regauthn.Authenticator,
 	}
 
 	for _, info := range infos {
-		if info.Hostname == target.RegistryStr() {
+		registryURLMatches, err := credentialprovider.URLsMatchStr(info.Hostname, target.RegistryStr())
```
cppforlife: might be good to document for users (lightly) how URLsMatchStr behaves.

pkg/imgpkg/registry/keychain_test.go
```diff
@@ -169,6 +170,58 @@ func TestAuthProvidedViaEnvVars(t *testing.T) {
...
 
+	for i, validHostname := range []string{
+		"localhost:9999",
+		"http://localhost:9999",
+		"https://localhost:9999",
+		"localhost:9999/v1/",
+		"localhost:9999/v2/",
+		"*:9999/v2/",
+		"local*:9999/v2/",
+		"localhost:9999/imgpkg_test",
```
cppforlife: do we allow globs within path?

pkg/imgpkg/registry/keychain_test.go
```diff
@@ -169,6 +170,58 @@ func TestAuthProvidedViaEnvVars(t *testing.T) {
...
 	})
 
+	for i, validHostname := range []string{
+		"localhost:9999",
+		"http://localhost:9999",
+		"https://localhost:9999",
+		"localhost:9999/v1/",
+		"localhost:9999/v2/",
+		"*:9999/v2/",
+		"local*:9999/v2/",
```
cppforlife: can we throw in a test for multi domain globs: *.*.foo.com?

pkg/imgpkg/registry/keychain.go
```diff
@@ -49,7 +50,12 @@ func (k *envKeychain) Resolve(target regauthn.Resource) (regauthn.Authenticator,
 	}
 
 	for _, info := range infos {
```
cppforlife: i think we need to consider how should ordering of infos works out here e.g.

foo.com/image

vs

foo.com

i think we would want to be matching more specific cases. do we have some precedent in credential provider hows thats done?

pkg/imgpkg/registry/keychain_test.go
DennisDenuto: No thats a limitation with the package (k8s-pkg-credentialprovider) we are using

https://github.com/vmware-tanzu/carvel-imgpkg/blob/3a779ca0886ab989a22c6928a1eb58cd36dc0bbd/vendor/github.com/vdemeester/k8s-pkg-credentialprovider/keyring.go#L203

pkg/imgpkg/registry/keychain.go
DennisDenuto: that makes sense... i'll add some ordering logic where the more specific match should take precedence.

pkg/imgpkg/registry/keychain_test.go
DennisDenuto: yep done!

pkg/imgpkg/registry/keychain.go
DennisDenuto: agreed. I had a TODO to document the new behavior of `IMGPKG_REGISTRY_HOSTNAME` 

I was thinking carvel.dev doc website. did you have something else in mind? (--help?)

pkg/imgpkg/registry/keychain.go
DennisDenuto: took a very similar approach to what `BasicDockerKeyring` does in terms of handling multiple urls.

https://github.com/vmware-tanzu/carvel-imgpkg/blob/3a779ca0886ab989a22c6928a1eb58cd36dc0bbd/vendor/github.com/vdemeester/k8s-pkg-credentialprovider/keyring.go#L42

pkg/imgpkg/registry/keychain.go
cppforlife: nah, just regular docs on carvel.dev

pkg/imgpkg/registry/keychain_test.go
cppforlife: ah interesting, i guess thats ok for now. could always expand later. what happens if somebody does throw in a star into a path? error, or something else?

DennisDenuto: not an error. it won't match (it will return an Anonymous) and will try the next keychain (iaas, cli, config.json etc)

DennisDenuto:
@cppforlife as discussed, merging into develop since tests went green.

PR #247, opened 2021-09-17
Allow setting azure-config-file via env variable

DennisDenuto:
POC

- before running a command, replace the string pointer used in the azure provider, to whatever was provided to the imgpkg registry azure flag (or env var)

### DIFF
--- a/pkg/imgpkg/cmd/registry_flags.go
+++ b/pkg/imgpkg/cmd/registry_flags.go
@@ -4,6 +4,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"time"
 
@@ -35,6 +36,29 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&r.Anon, "registry-anon", false, "Set anonymous auth ($IMGPKG_ANON)")
 
 	cmd.Flags().DurationVar(&r.ResponseHeaderTimeout, "registry-response-header-timeout", 30*time.Second, "Maximum time to allow a request to wait for a server's response headers from the registry (ms|s|m|h)")
+
+	cmd.Flags().String("registry-azure-cr-config", "", "Path to the file containing Azure container registry configuration information. ($IMGPKG_REGISTRY_AZURE_CR_CONFIG)")
+
+	err := cmd.LocalFlags().MarkHidden("azure-container-registry-config")
+	if err != nil {
+		panic(fmt.Sprintf("Unable to hide flag: %s", err))
+	}
+
+	if cmd.PersistentPreRunE != nil {
+		panic("Internal inconsistency: PersistentPreRunE was already set")
+	}
+
+	cmd.PersistentPreRunE = func(subCmd *cobra.Command, args []string) error {
+		registryAzureContainerConfigFlag := subCmd.Flag("registry-azure-cr-config")
+		if registryAzureContainerConfigFlag == nil {
+			return nil
+		}
+
+		if registryAzureContainerConfigFlag.Value.String() != "" {
+			return cmd.Flags().Set("azure-container-registry-config", registryAzureContainerConfigFlag.Value.String())
+		}
+		return cmd.Flags().Set("azure-container-registry-config", os.Getenv("IMGPKG_REGISTRY_AZURE_CR_CONFIG"))
+	}
 }
 
 func (r *RegistryFlags) AsRegistryOpts() registry.Opts {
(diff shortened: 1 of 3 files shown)


pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -35,6 +36,18 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
...
+		//TODO: manually test this alias is applying to the correct cmd...
+		switch name {
+		case "azure-container-registry-config":
+			name = "registry-azure-container-registry-config"
+			break
+		}
+		return pflag.NormalizedName(name)
+	})
+
+	cmd.Flags().String("registry-azure-container-registry-config", "", "Path to the file containing Azure container registry configuration information.")
```
cppforlife: wdyt? `registry-azure-container-registry-config` -> `registry-azure-cr-config`

i would also add `($IMGPKG_REGISTRY_AZURE_CR_CONFIG)` to the description (similar to other env var setters)

pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -35,6 +36,18 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&r.Anon, "registry-anon", false, "Set anonymous auth ($IMGPKG_ANON)")
 
 	cmd.Flags().DurationVar(&r.ResponseHeaderTimeout, "registry-response-header-timeout", 30*time.Second, "Maximum time to allow a request to wait for a server's response headers from the registry (ms|s|m|h)")
+
+	cmd.Flags().SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
```
cppforlife: for the sake of simplifying things, lets not alias original flag to the new one. original flag should not be used by anybody and if it is somehow discovered -- that would just be undefined behaviour from imgpkg's pov.

pkg/imgpkg/cmd/imgpkg.go
```diff
@@ -69,6 +70,24 @@ func NewImgpkgCmd(o *ImgpkgOptions) *cobra.Command {
...
+		registryAzureContainerConfigFlag := subCmd.Flag("registry-azure-container-registry-config")
+		if registryAzureContainerConfigFlag == nil {
+			return nil
+		}
+
+		if registryAzureContainerConfigFlag.Value.String() != "" {
+			return cmd.Flags().Set("azure-container-registry-config", registryAzureContainerConfigFlag.Value.String())
+		}
+		return cmd.Flags().Set("azure-container-registry-config", os.Getenv("IMGPKG_REGISTRY_AZURE_CONFIG"))
+	}
```
cppforlife: should we move this whole thing into RegistryFlags. it has access to cmd and this code would be closer where we define "our" flag.

pkg/imgpkg/cmd/imgpkg.go
```diff
@@ -69,6 +70,24 @@ func NewImgpkgCmd(o *ImgpkgOptions) *cobra.Command {
 	// This configurations forces all nodes to do not accept extra args, but the completion requires 1 extra arg
 	cmd.AddCommand(NewCompletionCmd())
 
+	err := cmd.LocalFlags().MarkHidden("azure-container-registry-config")
+	if err != nil {
+		//TODO: wrap around a useful err msg
+		panic(err.Error())
+	}
+
+	cmd.PersistentPreRunE = func(subCmd *cobra.Command, args []string) error {
```
cppforlife: i would throw in a quick panic if cmd.PersistentPreRunE is already defined. to make sure we dont override things accidently.

DennisDenuto:
Manually tested on Azure IaaS

- Using the old flag `azure-container-registry-config` does *not* work, since the new flag `registry-azure-cr-config` takes precedence with a default value of empty string
- `--registry-azure-cr-config` takes precedence over the env variable `IMGPKG_REGISTRY_AZURE_CR_CONFIG`

pkg/imgpkg/cmd/registry_flags.go
DennisDenuto: done

pkg/imgpkg/cmd/imgpkg.go
DennisDenuto: done

pkg/imgpkg/cmd/imgpkg.go
DennisDenuto: good idea. done

pkg/imgpkg/cmd/registry_flags.go
DennisDenuto: done.

DennisDenuto:
As discussed over slack, merging into develop:

 https://kubernetes.slack.com/archives/CH8KCCKA5/p1632166438343400?thread_ts=1632165510.340600&cid=CH8KCCKA5

PR #255, opened 2021-09-22
Disable default docker config provider

DennisDenuto:
TODO:

- [x] Add a bunch of comments and documentation around this hack-around
~- [ ] Submit an issue/PR upstream to allow disabling the default docker provider. Extend this change: https://github.com/kubernetes/kubernetes/commit/a947c32783e9027922762d6f76f5c27926f9bc0a#diff-0ebbc5bb19e182a89f5992c919d5fe2254af767f4611835010367b324b271df9R51-R53~

### DIFF
--- a/hack/build-binaries.sh
+++ b/hack/build-binaries.sh
@@ -12,6 +12,12 @@ go fmt ./cmd/... ./pkg/... ./test/...
 go mod vendor
 go mod tidy
 
+# related to https://github.com/vmware-tanzu/carvel-imgpkg/pull/255
+# there doesn't appear to be a simple way to disable the defaultDockerConfigProvider
+# Having defaultDockerConfigProvider enabled by default results in the imgpkg auth ordering not working correctly
+# Specifically, the docker config.json is loaded before cli flags (and maybe even IaaS metadata services)
+git apply ./hack/patch-k8s-pkg-credentialprovider.patch
+
 # makes builds reproducible
 export CGO_ENABLED=0
 LDFLAGS="-X github.com/k14s/imgpkg/pkg/imgpkg/cmd.Version=$VERSION -buildid="
--- a/pkg/imgpkg/registry/keychain_test.go
+++ b/pkg/imgpkg/registry/keychain_test.go
@@ -36,6 +36,9 @@ func TestMain(m *testing.M) {
 	gcpRegistryURL, server = registerGCPProvider()
 	defer server.Close()
 
+	tempConfigJSONProviderDir := registerDefaultDockerProvider()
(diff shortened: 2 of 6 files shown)


DennisDenuto:
@cppforlife 

FYI: There appears to be some incompatible changes made upstream (kubernetes) to auth providers. Not sure when / if those changes will eventually land on the fork we are using... 

Specifically...

amazon-ecr is always enabled: https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/aws/aws_credentials.go#L77-L80

And will return an empty dockerconfig if it is 'disabled': https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/aws/aws_credentials.go#L173

This change will require changes on imgpkg, since we will treat that case as 'we found a cred for your image... try using an empty username / password'

pkg/imgpkg/registry/keychain_test.go
```diff
@@ -664,6 +667,28 @@ func TestOrderingOfAuthOpts(t *testing.T) {
 	})
 }
 
+func registerDefaultDockerProvider() string {
+	tempConfigJSONProviderDir, err := ioutil.TempDir(os.TempDir(), "test-default-keychain-provider")
+	if err != nil {
+		panic(fmt.Errorf("unable to run test: %s", err))
+	}
+
+	err = ioutil.WriteFile(filepath.Join(tempConfigJSONProviderDir, "config.json"), []byte(`{
```
cppforlife, line 684: im assuming we are adding so that if somehow its getting picked up tests would fail. if so, should we make this file invalid, so that it makes things blow up.

hack/build-binaries.sh
```diff
@@ -12,6 +12,8 @@ go fmt ./cmd/... ./pkg/... ./test/...
 go mod vendor
 go mod tidy
 
+git apply ./hack/patch-k8s-pkg-credentialprovider.patch
```
cppforlife, line 19: lets add a comment why we do this.

cppforlife: (saw your TODO, ignore)

pkg/imgpkg/registry/keychain_test.go
DennisDenuto: It is a good idea but....

the `defaultDockerConfigProvider` won't blow up. it will just try the next path in its list of paths to try reading a docker config.json from.

On most dev machines and even our e2e tests on GH we will have a `~/.docker/config.json` and it will try to read and load that file instead.

With the above test change made in this PR:
- every `CLI flag` test fails
- 2 of the ordering tests fail (one being that cli > config.json)

PR #265, opened 2021-10-04
Fix verifying certs on windows

### LINKED ISSUE #22: Certs do not verify when running on windows
This issue was discussed in https://github.com/vmware-tanzu/carvel-kbld/issues/75.

Since `imgpkg` inherited some functionality from `kbld` we will need to fix it in `imgpkg` as well.

Notes
- see [kbld's comment](https://github.com/vmware-tanzu/carvel-kbld/issues/75#issuecomment-906670951) about reproducing this again before starting implementation.

DennisDenuto:
Attempts to fix #22 

TODO:

- [x] document how to use custom ca certs with windows root store
- [x] figure out why windows still has a file handle on the bundle.tar file. TODO in test code
- [x] confirm ca certs work in an airgapped env
- [x] investigate if theres any performance cost due to using windows root ca store (I think it downloads ca certs using network)

### DIFF
--- a/.github/workflows/gh-test.yml
+++ b/.github/workflows/gh-test.yml
@@ -8,7 +8,7 @@ on:
     types: [opened, reopened, synchronize]
 jobs:
   test-all:
-    name: Test GH
+    name: Test GH - Linux
     runs-on: ubuntu-latest
     steps:
     - name: Set up Go 1.x
@@ -29,3 +29,35 @@ jobs:
 
         # deploy local registry and run tests
         ./hack/test-all-local-registry.sh
+
+  test-all-windows:
+    name: Test GH - Windows
+    runs-on: windows-latest
+    steps:
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v1
+      with:
+        go-version: "1.17"
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+      with:
+        path: src/github.com/${{ github.repository }}
+    - name: Run Tests
+      run: |
+        bash -c '
+        set -e -x
+
+        export GOPATH=$(echo `pwd`)
+        export PATH="$GOPATH/bin:$PATH"
+        cd src/github.com/${{ github.repository }}
+
+        export IMGPKG_E2E_IMAGE="ttl.sh/github-action-test-relocation-$GITHUB_RUN_ID"
+        export IMGPKG_E2E_RELOCATION_REPO="ttl.sh/github-action-imgpkg-test-$GITHUB_RUN_ID"
+        # deploy local registry and run tests
+
+        go install github.com/sigstore/cosign/cmd/cosign@v0.5.0
+
+        alias cosign=cosign.exe
+
+        ./hack/test-all.sh ".exe"
+        '
--- a/.github/workflows/gh-test.yml
+++ b/.github/workflows/gh-test.yml
@@ -53,11 +53,11 @@ jobs:
 
         export IMGPKG_E2E_IMAGE="ttl.sh/github-action-test-relocation-$GITHUB_RUN_ID"
         export IMGPKG_E2E_RELOCATION_REPO="ttl.sh/github-action-imgpkg-test-$GITHUB_RUN_ID"
-        # deploy local registry and run tests
 
         go install github.com/sigstore/cosign/cmd/cosign@v0.5.0
 
         alias cosign=cosign.exe
 
-        ./hack/test-all.sh ".exe"
+        export IMGPKG_BINARY_EXT=".exe"
+        ./hack/test-all.sh
         '
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -2,6 +2,8 @@
 
 set -e -x -u
 
+FILE_EXT="${1-}"
+
 # makes builds reproducible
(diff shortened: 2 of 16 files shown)


pkg/imgpkg/cmd/push_test.go
```diff
@@ -56,7 +57,7 @@ func TestMultiImgpkgDirError(t *testing.T) {
 		t.Fatalf("Expected validations to err, but did not")
 	}
 
-	if !strings.Contains(err.Error(), "This directory contains multiple bundle definitions. Only a single instance of .imgpkg/images.yml can be provided") {
+	if !strings.Contains(err.Error(), fmt.Sprintf("This directory contains multiple bundle definitions. Only a single instance of .imgpkg%simages.yml can be provided", string(os.PathSeparator))) {
```
cppforlife, line 60: im not sure how i feel about us flipping slashes in error messages. should we do this or just leave it as is?

DennisDenuto: hmmm i'm not sure i fully understand here...

This is within a test asserting an error message though... so the implementation on windows is going to return a different error message when running this test in linux (the diff being the slash). So this aims to make the test make the right assertion based on the OS that is running said test.

cppforlife: i see, we are generating that stuff on the fly.

pkg/imgpkg/registry/registry.go
```diff
@@ -250,8 +251,17 @@ func (r Registry) FirstImageExists(digests []string) (string, error) {
 }
 
 func newHTTPTransport(opts Opts) (*http.Transport, error) {
-	pool, err := x509.SystemCertPool()
-	if err != nil {
+	var pool *x509.CertPool
+
+	if runtime.GOOS != "windows" {
```
cppforlife, line 259: lets throw in a comment here why we are doing this.

DennisDenuto: done

test/helpers/assets.go
```diff
@@ -94,6 +94,7 @@ func (a *Assets) CleanCreatedFolders() {
 	a.T.Helper()
 	for _, folder := range a.CreatedFolders {
 		err := os.RemoveAll(folder)
+
```
cppforlife: nit: remove newline?

hack/build.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e -x -u
 
+FILE_EXT="${1-}"
```
cppforlife: instead of turning this into an arg wdyt about just using $IMGPKG_BINARY_EXT env var

.github/workflows/gh-test.yml
```diff
@@ -28,4 +28,36 @@ jobs:
...
+        bash -c '
+        set -e -x
+
+        export GOPATH=$(echo `pwd`)
+        export PATH="$GOPATH/bin:$PATH"
+        cd src/github.com/${{ github.repository }}
+
+        export IMGPKG_E2E_IMAGE="ttl.sh/github-action-test-relocation-$GITHUB_RUN_ID"
+        export IMGPKG_E2E_RELOCATION_REPO="ttl.sh/github-action-imgpkg-test-$GITHUB_RUN_ID"
+        # deploy local registry and run tests
```
cppforlife: this comment doesnt seem to belong to anything

test/e2e/gcloud_expired_token_test.go
```diff
@@ -9,13 +9,17 @@ import (
...
 
 	"github.com/k14s/imgpkg/test/helpers"
 	"github.com/stretchr/testify/require"
 )
 
 func TestCopyWithBundleLockInputToRepoDestinationUsingGCloudWithAnExpiredToken(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Skipping test as docker image used requires linux")
+	}
 	env := helpers.BuildEnv(t)
```
cppforlife, line 24: nit: seems like in other places you gotta newline after this windows if

test/e2e/copy_from_image_test.go
```diff
@@ -116,7 +113,7 @@ func TestCopyAnImageFromATarToARepoThatDoesNotContainNonDistributableLayersButTh
 		// copy to tar skipping NDL
 		imgpkg.Run([]string{"copy", "-i", airgappedRepo, "--to-tar", tarFilePath})
 
-		stopRegistryForAirgapTesting(t, env)
+		stopRegistryForAirgapTesting(fakeRegistry)
```
cppforlife, line 116: should this function be replaced with defer fakeRegistry.CleanUp() after startRegistryForAirgapTesting()?

cppforlife: oh i see what's going on you are clearing it before copy... nm

.github/workflows/gh-test.yml
DennisDenuto: true. removed.

hack/build.sh
DennisDenuto: done

test/helpers/assets.go
DennisDenuto: done

PR #266, opened 2021-10-04
Disable defaultDockerConfigProvider

DennisDenuto:
- It was enabled by dependabot. When tests are run a git patch is
applied which is why the tests passed on the dependabot PR

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -16,6 +16,11 @@ go mod tidy
 # Specifically, the docker config.json is loaded before cli flags (and maybe even IaaS metadata services)
 git apply ./hack/patch-k8s-pkg-credentialprovider.patch
 
+git diff --exit-code vendor/ || {
+  echo 'found changes in the project. when expected none. exiting'
+  exit 1
+}
+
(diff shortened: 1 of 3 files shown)


DennisDenuto:
@cppforlife added a check in the build scripts to fail if unstaged git changes are found

hack/build.sh
```diff
@@ -16,6 +16,11 @@ go mod tidy
 # Specifically, the docker config.json is loaded before cli flags (and maybe even IaaS metadata services)
 git apply ./hack/patch-k8s-pkg-credentialprovider.patch
 
+git diff --exit-code || {
```
cppforlife: not sure if im getting this right, would this script always fail when there are changes. i use it for development heavily (it's initial purpose).

DennisDenuto: yeah you're right i should be running that diff on the vendor directory... oops good catch. I think theres value on leaving the build-binaries.sh remain on running diff on the entire project.

DennisDenuto: made the change to run diff on vendor.

Issue #55, opened 2021-01-12
Create a bundle or image directly as a tar file


joaopapereira:
**Describe the problem/challenge you have**
In a scenario where the artifact that will be shared is a tar, there might not be a need to use the registry as an intermediary step. This would allow the users to combine the commands `imgpkg push` + `imgpkg copy --to-tar` in a single command and without storing extra images in a registry

**Describe the solution you'd like**

Scenario for images
**Given** a set of files in examples/basic-step-1
**When** execute `imgpkg build --tag v0.5.0 -f examples/basic-step-1 --to-tar some.tar`
**Then** copy the image to a registry`imgpkg copy --tar some.tar --to-repo localhost:5000/some-image`
**And** retrieve it using `imgpkg pull -i localhost:5000/some-image:v0.5.0`

Scenario for bundles
**Given** a bundle in examples/basic-step-2
**When** execute `imgpkg build --tag v0.5.0 -f examples/basic-step-1 --to-tar some-bundle.tar`
**Then** copy the bundle to a registry`imgpkg copy --tar some-bundle.tar --to-repo localhost:5000/some-bundle`
**And** retrieve it using `imgpkg pull -b localhost:5000/some-image:v0.5.0`

Notes:
- Tags used during the build step should be kept when copying

Based on a request from Zihan Jiang: [Slack conversations](https://kubernetes.slack.com/archives/CH8KCCKA5/p1610484485443500)

StevenLocke:
This feels like a good enhancement for imgpkg. One question we'd like to pose is whether this fits into a top level command or should be separated?
i.e. Could this be a subset of `copy` or should we extract it to a `build` command?

At the moment, my understanding is that `tar`s do not have any tagging metadata. If that's a firm requirement, we may want to add that functionality to all of our tar creating.
We have an old desired feature to support a `--to-tag` flag, maybe that would be sufficient for the "note" behaviour.

joaopapereira:
>i.e. Could this be a subset of copy or should we extract it to a build command?

The idea behind doing this as a separated command is to separate these 2 concerns and also do not overload the `copy` command. Sounds more comprehensive to `build an OCI Image tar from a set of files` than `copy a set of files into an OCI Image tar`. But we can try to talk with some users to better understand if this is the best UX option.

>At the moment, my understanding is that tars do not have any tagging metadata. If that's a firm requirement, we may want to add that functionality to all of our tar creating.
[...]

PR #273, opened 2021-10-11
Create a bundle or image directly as a tar file

DennisDenuto:
Attempts to implement: https://github.com/vmware-tanzu/carvel-imgpkg/issues/55

TODO:

- [x] manual test spike and ensure it works writing to a tarball directly not requiring a registry
- [x] look into providing a registry that can serve local images and images from the registry as an alternate way to implement this story
- [x] throw spike away and TDD implementation
- [x] consider whether a new command is needed "build" or whether this is part of the "push" command


pivotaljohn:
FWIW, there _may_ be an additional use-case for this feature: vmware-tanzu/carvel-kbld#158.

`kbld` has a mode where it does not push built artifacts. Having the ability to do so with `imgpkg` bundles fits nicely with that mode.

cppforlife:
we'll try to rework based on https://github.com/vmware-tanzu/carvel-imgpkg/issues/55#issuecomment-962209740

cppforlife:
per discussion with dennis, will close. not sure if above proposed solution will result in some code reuse from this PR.

parteekcoder:
@cppforlife I want to work on `Package Repository to tar` this project but does not know where can i start from. I saw this PR which relates to this project.

Could you please help me

PR #276, opened 2021-10-13
ux: improve error message when timeout occurs trying to auth via IaaS

DennisDenuto:
- hint around the feature flag to disable auth via IaaS

### DIFF
--- a/pkg/imgpkg/registry/auth/iaas_keychain.go
+++ b/pkg/imgpkg/registry/auth/iaas_keychain.go
@@ -61,7 +61,7 @@ func NewIaasKeychain(ctx context.Context, environFunc func() []string) (authn.Ke
 			keyring: keyring,
 		}, nil
 	case <-timeout.Done():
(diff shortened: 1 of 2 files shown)


pkg/imgpkg/registry/auth/iaas_keychain.go
```diff
@@ -61,7 +61,7 @@ func NewIaasKeychain(ctx context.Context, environFunc func() []string) (authn.Ke
 			keyring: keyring,
 		}, nil
 	case <-timeout.Done():
-		return nil, fmt.Errorf("Timeout occurred trying to enable iaas provider")
+		return nil, fmt.Errorf("Timeout occurred trying to enable IaaS provider. Hint: To skip authenticating via IaaS set the environment variable IMGPKG_ENABLE_IAAS_AUTH=false")
```
cppforlife: carvel tools typically use `...provider (hint: ...)` format for hints.

DennisDenuto: good catch. done

PR #281, opened 2021-10-21
Images are only tagged after all blobs are uploaded

### LINKED ISSUE #211: ensure that root bundle tag is added/updated last during "imgpkg copy"

currently, if you are continuously running `imgpkg copy -b foo:latest --to-repo bar`, `bar:latest` will get updated in parallel with all the other referenced assets. this may present a challenge to consumers who are pulling `bar:latest` since they could receive `bundle:latest` before referenced assets are copied (and *.image-locations.imgpkg is made).

this ultimately is related to when we add/update tag (in above example latest tag). we need to update tag at the last moment when all the related resources are in place: referenced blobs, root bundle blob, *.image-locations.imgpkg. we probably need to do this in order from inner bundles, to outer bundles (but can still parallelize writing of all image-locations.imgpkg images).

i dont believe we do this today.

---
Vote on this request

joaopapereira:
The images were being tagged after all the images (that were part of a bundle) were uploaded to the registry. Nevertheless, this was happening before the creation of the Images Location image. This could cause some concurrency problems if someone tried to pull the images as soon as it was tagged.

Fixes #211

### DIFF
--- a/pkg/imgpkg/cmd/copy_repo_src.go
+++ b/pkg/imgpkg/cmd/copy_repo_src.go
@@ -59,6 +59,7 @@ func (c CopyRepoSrc) CopyToTar(dstPath string) error {
 func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error) {
 	c.ui.Tracef("CopyToRepo(%s)\n", repo)
 
+	var processedImages *ctlimgset.ProcessedImages
 	importRepo, err := regname.NewRepository(repo)
 	if err != nil {
 		return nil, fmt.Errorf("Building import repository ref: %s", err)
@@ -69,45 +70,35 @@ func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error)
 			return nil, fmt.Errorf("Cannot use tar source (--tar) with tar destination (--to-tar)")
 		}
 
-		processedImages, err := c.tarImageSet.Import(c.TarFlags.TarSrc, importRepo, c.registry)
+		processedImages, err = c.tarImageSet.Import(c.TarFlags.TarSrc, importRepo, c.registry)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		unprocessedImageRefs, bundles, err := c.getAllSourceImages()
 		if err != nil {
 			return nil, err
 		}
 
-		informUserToUseTheNonDistributableFlagWithDescriptors(
-			c.ui, c.IncludeNonDistributable, processedImagesMediaType(processedImages))
-
-		err = c.tagAllImages(processedImages)
+		processedImages, err = c.imageSet.Relocate(unprocessedImageRefs, importRepo, c.registry)
 		if err != nil {
-			return nil, fmt.Errorf("Tagging images: %s", err)
+			return nil, err
 		}
 
-		return processedImages, nil
-	}
-
-	unprocessedImageRefs, bundles, err := c.getAllSourceImages()
-	if err != nil {
-		return nil, err
-	}
-
-	processedImages, ids, err := c.imageSet.Relocate(unprocessedImageRefs, importRepo, c.registry)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, bundle := range bundles {
-		if err := bundle.NoteCopy(processedImages, c.registry, c.ui); err != nil {
-			return nil, fmt.Errorf("Creating copy information for bundle %s: %s", bundle.DigestRef(), err)
+		for _, bundle := range bundles {
+			if err := bundle.NoteCopy(processedImages, c.registry, c.ui); err != nil {
+				return nil, fmt.Errorf("Creating copy information for bundle %s: %s", bundle.DigestRef(), err)
+			}
 		}
 	}
+
+	informUserToUseTheNonDistributableFlagWithDescriptors(
+		c.ui, c.IncludeNonDistributable, processedImagesMediaType(processedImages))
 
 	err = c.tagAllImages(processedImages)
 	if err != nil {
 		return nil, fmt.Errorf("Tagging images: %s", err)
 	}
-
-	informUserToUseTheNonDistributableFlagWithDescriptors(
-		c.ui, c.IncludeNonDistributable, imageRefDescriptorsMediaTypes(ids))
 
 	return processedImages, nil
 }
--- a/pkg/imgpkg/cmd/copy_repo_src.go
+++ b/pkg/imgpkg/cmd/copy_repo_src.go
@@ -77,6 +77,11 @@ func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error)
 		informUserToUseTheNonDistributableFlagWithDescriptors(
 			c.ui, c.IncludeNonDistributable, processedImagesMediaType(processedImages))
 
+		err = c.tagAllImages(processedImages)
+		if err != nil {
+			return nil, fmt.Errorf("Tagging images: %s", err)
+		}
+
 		return processedImages, nil
 	}
 
@@ -94,6 +99,11 @@ func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error)
 		if err := bundle.NoteCopy(processedImages, c.registry, c.ui); err != nil {
 			return nil, fmt.Errorf("Creating copy information for bundle %s: %s", bundle.DigestRef(), err)
 		}
+	}
+
+	err = c.tagAllImages(processedImages)
+	if err != nil {
+		return nil, fmt.Errorf("Tagging images: %s", err)
 	}
 
 	informUserToUseTheNonDistributableFlagWithDescriptors(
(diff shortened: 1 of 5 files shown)


pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -242,3 +252,60 @@ func imageRefDescriptorsMediaTypes(ids *imagedesc.ImageRefDescriptors) []string
...
+				digest, err := regname.NewDigest(item.DigestRef)
+				if err != nil {
+					panic(fmt.Sprintf("Internal consistency: %s should be a digest", item.DigestRef))
+				}
+
+				uploadOriginalTagRef := digest.Tag(item.Tag)
+
+				switch {
+				case item.Image != nil:
+					err = c.registry.WriteTag(uploadOriginalTagRef, item.Image)
```
cppforlife: hmm. should we be doing this instead? https://github.com/vmware-tanzu/carvel-kbld/blob/develop/pkg/kbld/registry/registry.go#L118 (seems a bit weird to pass in entire image/index in there)

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -242,3 +252,60 @@ func imageRefDescriptorsMediaTypes(ids *imagedesc.ImageRefDescriptors) []string
...
+			go func() {
+				importThrottle.Take()
+				defer importThrottle.Done()
+
+				digest, err := regname.NewDigest(item.DigestRef)
+				if err != nil {
+					panic(fmt.Sprintf("Internal consistency: %s should be a digest", item.DigestRef))
+				}
+
+				uploadOriginalTagRef := digest.Tag(item.Tag)
```
cppforlife: uploadOriginalTagRef -> customTagRef

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -242,3 +252,60 @@ func imageRefDescriptorsMediaTypes(ids *imagedesc.ImageRefDescriptors) []string
...
+
+func (c CopyRepoSrc) tagAllImages(processedImages *ctlimgset.ProcessedImages) error {
+	importThrottle := util.NewThrottle(c.Concurrency)
+
+	totalThreads := 0
+	errChVerifyImages := make(chan error, processedImages.Len())
+	for _, item := range processedImages.All() {
+		item := item // copy
+
+		if item.Tag != "" {
```
cppforlife: might be better to do `item.Tag == "" continue` to avoid extra indention.

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -242,3 +252,60 @@ func imageRefDescriptorsMediaTypes(ids *imagedesc.ImageRefDescriptors) []string
 	}
 	return mediaTypes
 }
+
+func (c CopyRepoSrc) tagAllImages(processedImages *ctlimgset.ProcessedImages) error {
+	importThrottle := util.NewThrottle(c.Concurrency)
+
+	totalThreads := 0
+	errChVerifyImages := make(chan error, processedImages.Len())
```
cppforlife: errChVerifyImages -> errCh

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -242,3 +252,60 @@ func imageRefDescriptorsMediaTypes(ids *imagedesc.ImageRefDescriptors) []string
...
+				}
+
+				errChVerifyImages <- nil
+			}()
+		}
+	}
+	return checkForAnyAsyncErrors(totalThreads, errChVerifyImages)
+}
+
+func checkForAnyAsyncErrors(numAsyncCalls int, errCh chan error) error {
```
cppforlife: i would favor inlining since its a pretty common pattern to happen after for-loop + goroutines

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -242,3 +252,60 @@ func imageRefDescriptorsMediaTypes(ids *imagedesc.ImageRefDescriptors) []string
...
+					panic(fmt.Sprintf("Internal consistency: %s should be a digest", item.DigestRef))
+				}
+
+				uploadOriginalTagRef := digest.Tag(item.Tag)
+
+				switch {
+				case item.Image != nil:
+					err = c.registry.WriteTag(uploadOriginalTagRef, item.Image)
+					if err != nil {
+						errChVerifyImages <- fmt.Errorf("Importing image as %s: %s", digest.Name(), err)
```
cppforlife: "Importing image as" -> "Tagging image"

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -242,3 +252,60 @@ func imageRefDescriptorsMediaTypes(ids *imagedesc.ImageRefDescriptors) []string
 	}
 	return mediaTypes
 }
+
+func (c CopyRepoSrc) tagAllImages(processedImages *ctlimgset.ProcessedImages) error {
+	importThrottle := util.NewThrottle(c.Concurrency)
```
cppforlife: importThrottle -> throttle

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -439,6 +439,31 @@ func TestToRepoBundleContainingANestedBundle(t *testing.T) {
...
+		requestLog := fakeDestRegistry.WithRequestLogging()
+		fakeDestRegistry.Build()
+
+		destRepo := fakeDestRegistry.ReferenceOnTestServer("library/bundle-copy")
+		_, err := subject.CopyToRepo(destRepo)
+		require.NoError(t, err)
+
+		for _, log := range *requestLog {
+			fmt.Printf("%s %s\n", log.Method, log.URL)
+		}
```
cppforlife: keep or not to keep?

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -96,6 +101,11 @@ func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error)
 		}
 	}
 
+	err = c.tagAllImages(processedImages)
+	if err != nil {
+		return nil, fmt.Errorf("Tagging images: %s", err)
+	}
+
 	informUserToUseTheNonDistributableFlagWithDescriptors(
 		c.ui, c.IncludeNonDistributable, imageRefDescriptorsMediaTypes(ids))
```
cppforlife: should we consolidate this into single exit point (instead of 2 currently)?

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: You need to provide ggcr with a `remote.Taggable` which is implemented by the `v1.Image` and `v1.ImageIndex` if you see that code from `kbld` there is a `remote.Get` call (https://github.com/vmware-tanzu/carvel-kbld/blob/develop/pkg/kbld/registry/registry.go#L130) that we do not need to do again since we already have a reference to a `remote.Taggable`.

If we did move to a similar interface as we do in `kbld` we would have to retrieve the Image that we already have from the registry. Do you think we need the extra step?

cppforlife: oh doh. i totally glanced over that bit.

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: I make sense, fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: I was trying to keep the same pattern we already have in the code in a different place, but for now we only need it in one spot so it make sense to do what you suggest

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: this one is redundant

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: this one is redundant.

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: This one is redundant

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: fixed

pkg/imgpkg/cmd/copy_repo_src_test.go
joaopapereira: Left over from debugging the test. Removed

pkg/imgpkg/cmd/copy_repo_src.go
joaopapereira: I was able to refactor the function and consolidate both the exit points, it should be good now

PR #283, opened 2021-10-21
Use cosign to sign release binary files

DennisDenuto:
- Attaches the signature and public key required to verify the binary file

Example release: https://github.com/DennisDenuto/carvel-imgpkg/releases/tag/v0.40.0

TODO:
- [x] Also sign via github, use a carvel owned private/public key. Each carvel tool would require its own public key to be provided to verify all-the-blobs across multiple versions.
- [x] find out if verify-blob using keyless has a simpler way to find the public key (similar to how sign works)
- [x] Pin to a specific version of cosign-installer (instead of main)
- [x] figure out why windows signature file didn't get uploaded (but the windows crt and other distro signatures did)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,11 +16,37 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - uses: sigstore/cosign-installer@main
+      - uses: sigstore/cosign-installer@v1.3.0
       - name: Set up Go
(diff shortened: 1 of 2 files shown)


DennisDenuto:
@cppforlife 
Before merging:

- requires configuring new secrets 
- requires configuring a Workload Identity Provider. See https://githubmemory.com/index.php/repo/google-github-actions/auth#setup

PR #289, opened 2021-10-26
Create ImagesLocation when copying images from a tar file

### LINKED ISSUE #212: ImageLocations is not generated when source is tar
**What steps did you take:**
looked at code [0]

**What happened:**
it does not look like when tar is source we call NoteCopy which means we are not generating *.image-locations.imgpkg images later used for pull optimization.

**What did you expect:**
expected ImageLocations record to be generated.

this issue should probably be dealt in a context of https://github.com/vmware-tanzu/carvel-imgpkg/issues/211

[0] https://github.com/vmware-tanzu/carvel-imgpkg/blob/7d2fd285eaa493c5460b5739104e6876fcc04217/pkg/imgpkg/cmd/copy.go#L110-L116

**Environment:**

- imgpkg version (use `imgpkg --version`):

---
Vote on this request

joaopapereira:
- Fixed issue found in the function that Uncompresses the layers, the function was not being used in the code prior to this change
- Rename the struct to better depict what it represents

Fixes #212


pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -43,6 +43,35 @@ func (o *Bundle) AllImagesRefs(concurrency int, ui util.UIWithLevels) ([]*Bundle
...
 }
 
+// UpdateImageRefs updates the bundle cached images
+func (o *Bundle) UpdateImageRefs(bundles []*Bundle, ui util.UIWithLevels) error {
+	o.cachedImageRefs = map[string]ImageRef{}
+
+	img, err := o.checkedImage()
+	if err != nil {
+		return err
+
```
cppforlife: nit: remove newline

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -43,6 +43,35 @@ func (o *Bundle) AllImagesRefs(concurrency int, ui util.UIWithLevels) ([]*Bundle
...
+	}
+
+	imageRefsToProcess, err := o.fetchImagesRef(img, ui)
+	if err != nil {
+		return fmt.Errorf("fetching images of %s: %s", o.DigestRef(), err)
```
cppforlife: nit: capitalize first letter in error message

pkg/imgpkg/imageset/unprocessed_image_refs.go
```diff
@@ -19,6 +19,9 @@ type UnprocessedImageRef struct {
 }
 
 func (u UnprocessedImageRef) Key() string {
+	if u.Tag == "" {
+		return u.DigestRef
+	}
```
cppforlife: was absence of this breaking anything? my 2c, keeping this as simple as possible is better since contents of this string are "internal" (if anything i was just thinking may be we should add `"internal-"+` prefix)

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -43,6 +43,35 @@ func (o *Bundle) AllImagesRefs(concurrency int, ui util.UIWithLevels) ([]*Bundle
 	return bundles, allImageRefs, err
 }
 
+// UpdateImageRefs updates the bundle cached images
+func (o *Bundle) UpdateImageRefs(bundles []*Bundle, ui util.UIWithLevels) error {
```
cppforlife, line 47: - UpdateImageRefs seems to be suggesting details about some internal state of the Bundle (which is what it does, but we shouldnt be expsosing that in naming)
- feels a little bit weird to rely on bundles to give you authoritative information for all bundles collected. this i guess assumes that copy had all the information available to it
  - this becomes a problem if we do not copy all the images for example

second bullet is most concerning to me... might be worth discussing a bit.

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: fixed

pkg/imgpkg/bundle/bundle_images_lock.go
joaopapereira: fixed

pkg/imgpkg/imageset/unprocessed_image_refs.go
joaopapereira: In my first implementation it was cause issues, but in the current solution it is not, so I removed this change

cppforlife:
dont have any better ideas on naming of function. merging. can always circle back with better ideas.

Issue #248, opened 2021-09-17
push adds a default, unique tag


seemiller:
**Describe the problem/challenge you have**

OCI registries can perform a garbage collection process where any un-tagged image is deleted.

This can cause a problem when tags are re-used, with the tag migrating from an older image to a newer one, but with the digest of the older image still being referenced.

Here's how the problem arises.

1. An image with URL/Digest `image@sha256:ABC123` is tagged `v.1.2.3`.
2. Image is referenced in the package repository with the digest `image@sha256:ABC123`
3. The new image uploaded, gets the digest, `image@sha256:ZCYD23`
4. The tag is move from `ABC123` to `ZCYD23`
5. `image@sha256:ABC123` _no longer has a tag_.
5. Package repository still points to `image@sha256:ABC123`
6. Garbage collector deletes `image@sha256:ABC123`
7. Users still using the original package repo will now experience `image not found` errors when attempting to install the package or package repository.

![Registry GC](https://user-images.githubusercontent.com/5900/133817744-6e8bcc27-acd3-4691-ad71-afafdc4a2e58.png)

**Describe the solution you'd like**

imgpkg should provide a means to add a unique/arbitrary tag when pushing images. Any image with this unique/arbitrary tag would never be garbage collected unless that tag was manually removed.

This would be optional and driven by a command line argument or environment variable.

`imgpkg push --bundle registry.example.com/foo/image:v1.2.3 --file ./bundle --unique-tag`

**Anything else you would like to add:**
[Additional information that will assist in solving the issue.]

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

jorgemoralespou:
@seemiller I'm assuming the `unique-tag` will be in addition to the provided tag for the repository. 

- Any recommendation on the tag format?
- Should it be configurable, or default is ok?

seemiller:
At the moment, I have no opinion on the format of the tag. Options include:
* A (partial) SHA
* The time stamp of the upload
* Randomly generated string
* A user provided value
* `fred`

I'm not to attached to any particular format as all that is needed is another tag to prevent garbage collection.

cppforlife:
[...]

PR #292, opened 2021-10-29
A default tag is added to all pushed images and bundles #248

seemiller:
A default tag in the form of `{digest.Algorithm}-{digest.Hex}.imgpkg`
will be added to all pushed images and bundles.

### DIFF
--- a/pkg/imgpkg/bundle/contents.go
+++ b/pkg/imgpkg/bundle/contents.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cppforlife/go-cli-ui/ui"
 	regname "github.com/google/go-containerregistry/pkg/name"
 	regv1 "github.com/google/go-containerregistry/pkg/v1"
+	regremote "github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/plainimage"
 )
 
@@ -30,6 +31,7 @@ type Contents struct {
 type ImagesMetadataWriter interface {
 	ImagesMetadata
 	WriteImage(regname.Reference, regv1.Image) error
+	WriteTag(ref regname.Tag, taggagle regremote.Taggable) error
 }
 
 func NewContents(paths []string, excludedPaths []string) Contents {
--- a/pkg/imgpkg/imageset/image_set.go
+++ b/pkg/imgpkg/imageset/image_set.go
@@ -147,7 +147,7 @@ func checkForAnyAsyncErrors(imgOrIndexes []imagedesc.ImageOrIndex, errCh chan er
 }
 
 func (i ImageSet) getImageOrImageIndexForMultiWrite(item imagedesc.ImageOrIndex, importRepo regname.Repository, registry registry.ImagesReaderWriter) (regname.Tag, regremote.Taggable, error) {
-	uploadTagRef, err := buildUploadTagRef(item, importRepo)
+	uploadTagRef, err := util.BuildDefaultUploadTagRef(item, importRepo)
 	if err != nil {
 		return regname.Tag{}, nil, err
 	}
@@ -191,20 +191,6 @@ func (i ImageSet) mountableImage(imageWithRef imagedesc.ImageWithRef, uploadTagR
 	return regv1.Image(imageWithRef), nil
 }
 
-func buildUploadTagRef(item imagedesc.ImageOrIndex, importRepo regname.Repository) (regname.Tag, error) {
-	itemDigest, err := item.Digest()
-	if err != nil {
-		return regname.Tag{}, err
-	}
-
-	tag := fmt.Sprintf("%s-%s.imgpkg", itemDigest.Algorithm, itemDigest.Hex)
-	uploadTagRef, err := regname.NewTag(fmt.Sprintf("%s:%s", importRepo.Name(), tag))
-	if err != nil {
-		return regname.Tag{}, fmt.Errorf("Building upload tag image ref: %s", err)
-	}
-	return uploadTagRef, nil
-}
-
 func (i *ImageSet) verifyImageOrIndex(item imagedesc.ImageOrIndex, importRepo regname.Repository, registry registry.ImagesReaderWriter) (ProcessedImage, error) {
 	existingRef, err := regname.NewDigest(item.Ref())
 	if err != nil {
@@ -244,7 +230,7 @@ func (i *ImageSet) verifyItemCopied(item imagedesc.ImageOrIndex, importRepo regn
 	}
 
 	// AWS ECR doesnt like using digests for manifest uploads
-	uploadTagRef, err := buildUploadTagRef(item, importRepo)
+	uploadTagRef, err := util.BuildDefaultUploadTagRef(item, importRepo)
 	if err != nil {
 		return regname.Digest{}, err
 	}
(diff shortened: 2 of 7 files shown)


pkg/imgpkg/plainimage/contents.go
```diff
@@ -53,6 +53,16 @@ func (i Contents) Push(uploadRef regname.Tag, labels map[string]string, writer I
 		return "", err
 	}
 
+	uploadTagRef, err := util.BuildDefaultUploadTagRef(digest, uploadRef.Repository)
+	if err != nil {
+		return "", fmt.Errorf("Building default upload tag image ref: %s", err)
+	}
+
+	err = writer.WriteImage(uploadTagRef, img)
```
joaopapereira: We are not really writing an Image here so I would suggest extending the Interface to add the WriteTag function that we have in the [Registry Interface](https://github.com/vmware-tanzu/carvel-imgpkg/blob/develop/pkg/imgpkg/registry/registry.go#L52)

pkg/imgpkg/imageset/image_set.go
```diff
@@ -164,7 +164,12 @@ func checkForAnyAsyncErrors(imgOrIndexes []imagedesc.ImageOrIndex, errCh chan er
 }
 
 func (i ImageSet) getImageOrImageIndexForMultiWrite(item imagedesc.ImageOrIndex, importRepo regname.Repository, registry ImagesReaderWriter) (regname.Tag, regremote.Taggable, error) {
-	uploadTagRef, err := buildUploadTagRef(item, importRepo)
+	digest, err := item.Digest()
```
joaopapereira: I think it would be more useful in this file if we completely remove buildUploadTagRef and only call the new function in the `util` package.

pkg/imgpkg/util/tag.go
```diff
@@ -0,0 +1,20 @@
...
+package util
+
+import (
+	"fmt"
+	regname "github.com/google/go-containerregistry/pkg/name"
+	regv1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+// BuildDefaultUploadTagRef Builds a tag from the digest Algorithm and Digest
+func BuildDefaultUploadTagRef(digest regv1.Hash, importRepo regname.Repository) (regname.Tag, error) {
```
joaopapereira: Since when we call this function everywhere we need to do a call to the `Digest` function maybe it would be better if we move that call to this function.
This would change the signature of the function to:
`func BuildDefaultUploadTagRef(digest WithDigest, importRepo regname.Repository) (regname.Tag, error)`
Create the following interface in `util`
```
type WithDigest interface {
   Digest() (regv1.Hash, error)
}
```
Not a fan of the name, maybe `Digestable` or something?

What do you think?

pkg/imgpkg/plainimage/contents.go
```diff
@@ -5,14 +5,14 @@ package plainimage
...
-	"os"
-	"path/filepath"
-	"strings"
-
 	"github.com/cppforlife/go-cli-ui/ui"
 	regname "github.com/google/go-containerregistry/pkg/name"
 	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	ctlimg "github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/image"
+	"github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/util"
+	"os"
```
joaopapereira: Do you mind moving the Golang imports to the top?

test/e2e/bundle_test.go
```diff
@@ -18,6 +18,29 @@ import (
...
+	imgpkg := helpers.Imgpkg{T: t, ImgpkgPath: env.ImgpkgPath}
+	defer env.Cleanup()
+
+	bundleDir := env.BundleFactory.CreateBundleDir(helpers.BundleYAML, helpers.ImagesYAML)
+
+	out := imgpkg.Run([]string{"push", "--tty", "-b", env.Image, "-f", bundleDir})
+	bundleDigest := fmt.Sprintf("@%s", helpers.ExtractDigest(t, out))
+
+	outDir := env.Assets.CreateTempFolder("imgpkg-test-basic")
+	imgpkg.Run([]string{"pull", "-b", env.Image + bundleDigest, "-o", outDir})
```
joaopapereira: I think in this case you want to pull by tag so you would have to pull using:
`env.Image:fmt.Sprintf("%s-%s.imgpkg", strings.Split(bundleDigest, ":")...)`

test/e2e/image_test.go
```diff
@@ -17,8 +18,10 @@ func TestPushPull(t *testing.T) {
 
 	testDir := env.Assets.CreateTempFolder("imgpkg-test-basic")
 
-	imgpkg.Run([]string{"push", "-i", env.Image, "-f", env.Assets.SimpleAppDir()})
-	imgpkg.Run([]string{"pull", "-i", env.Image, "-o", testDir})
+	out := imgpkg.Run([]string{"push", "--tty", "-i", env.Image, "-f", env.Assets.SimpleAppDir()})
+	imageDigest := fmt.Sprintf("@%s", helpers.ExtractDigest(t, out))
+
+	imgpkg.Run([]string{"pull", "-i", env.Image + imageDigest, "-o", testDir})
```
joaopapereira: Same as the previous test comment.

joaopapereira (changes requested):
Thanks for the PR, I did make some comments.

pkg/imgpkg/plainimage/contents.go
seemiller: Thought I had this one, but the FakeImagesMetadataWriter doesn't have a WriteTag method. This is the first time I've encountered counterfeiter, so I may need a pairing session to help me along on this...

pkg/imgpkg/bundle/contents.go
```diff
@@ -5,6 +5,7 @@ package bundle
 
 import (
 	"fmt"
+	regremote "github.com/google/go-containerregistry/pkg/v1/remote"
```
joaopapereira: This import should be right after the `regname` one.

We try to keep the imports separated into 2 groups the first one will contain the language imports and the second one will contain dependency imports.

pkg/imgpkg/imageset/image_set.go
```diff
@@ -209,17 +209,7 @@ func (ImageSet) mountableImage(imageWithRef imagedesc.ImageWithRef, uploadTagRef
 }
 
 func buildUploadTagRef(item imagedesc.ImageOrIndex, importRepo regname.Repository) (regname.Tag, error) {
```
joaopapereira: Do we still use this function? If not I think it is safe to remove it

pkg/imgpkg/plainimage/contents.go
```diff
@@ -5,6 +5,7 @@ package plainimage
 
 import (
 	"fmt"
+	regremote "github.com/google/go-containerregistry/pkg/v1/remote"
```
joaopapereira: Same comment as prior import

pkg/imgpkg/bundle/contents.go
joaopapereira: goland screenshot of the configuration that we use
Screen Shot 2021-12-06 at 11 27 00 AM

joaopapereira (changes requested):
Added some comments on the PR. They are just import and used function-related issues.

pkg/imgpkg/plainimage/contents.go
seemiller: Figured it out!

joaopapereira (approved):
LGTM

PR #295, opened 2021-11-08
Same registry different credentials

### LINKED ISSUE #294: Copying to same registry with authentication per repository fails
**What steps did you take:**
When a bundle is being copied to the same registry, but to a different repository, if both repositories have different authentication `imgpkg` fails with a 401 when trying to mount the repositories.

Let's assume we have a bundle `my.registry.io/my-bundle:tag` with the following image
`my.registry.io/application1/controller` and to be able to copy from this image the user needs the following authentication
```bash
export IMGPKG_REGISTRY_HOSTNAME_0=my.registry.io/application1
export IMGPKG_REGISTRY_USERNAME_0=user1
export IMGPKG_REGISTRY_PASSWORD_0=some-password
```

When we are copying the bundle to `my.registry.io/release/bundle:tag` the user needs the following authentication
```bash
export IMGPKG_REGISTRY_HOSTNAME_1=my.registry.io/release
export IMGPKG_REGISTRY_USERNAME_1=user2
export IMGPKG_REGISTRY_PASSWORD_1=some-other-password
```

To copy the user executes (in the same shell where the prior exports were made):
`imgpkg copy -b my.registry.io/my-bundle:tag --to-repo my.registry.io/release/bundle:tag`

**What happened:**
This is the output of `imgpkg`
```
copy | importing 1 images...

 4.16 MiB / 197.57 MiB [=========>-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------]   2.11% 8.99 MiB/s 0s

copy | done uploading images

imgpkg: Error: Non-retryable error:
  POST https://my.registry.io/v2/release/bundle/blobs/uploads/?from=application1%2Fcontroller&mount=sha256%3A1739aaf524aa3b31288b8d2008ec831618d3a58b721161dfba96e8adafa9f417:
    UNAUTHORIZED: unauthorized to access repository: application1/controller, action: pull: unauthorized to access repository: application1/controller, action: pull
``` 

**What did you expect:**
The copy to succeed since all the needed authentication was provided

The problem in this situation is that `imgpkg` is trying to mount an image into a different repository and the keychain is providing the authentication information for the destination. In this case, the authentication for the destination does not allow `imgpkg` to read the source.

One possible solution for this problem might be using the destination credentials when checking to see if an image is mountable. If the destination credentials cannot read the image, it should not try to mount it.

**Note:** Due to the implementation of `imgpkg` this same issue occurs even if the user tries to copy to tar and afterward copy it back to the registry because `imgpkg` will understand that the images to copy are in the same registry so it will try to mount them.

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.23.0
- Docker registry used (e.g. `Docker HUB`):  Harbour

---
Vote on this request

joaopapereira:
This PR contains the fix for issue #294. 
In the commit history, there is a rollbacked commit because we had to do a fix on the Registry struct of ggcr to ensure that we were able to test the failure. The revert commit falls back to the previous version of ggcr. What happens with the current implementation of the Registry from ggcr is that it does not take into account the repository that pushed a particular blob into it. So when checking if a blob exists it will always find it even if it was not provided by a particular repository. This behavior is not specified in the spec, so it is not a bug. A [PR](https://github.com/google/go-containerregistry/pull/1158) was created in ggcr that could allow us to provide the blob storage, which would help us create the behavior we want, but it was not merged.

Also, some refactoring is done in our Registry package to convert the Registry struct into an interface.

Fixes #294


pkg/imgpkg/registry/keychain.go
```diff
@@ -14,6 +15,10 @@ import (
 // keychains that contain credentials for 'any' target. i.e. env keychain takes precedence over the custom keychain.
 // Since env keychain contains credentials per HOSTNAME, and custom keychain doesn't.
 func Keychain(keychainOpts auth.KeychainOpts, environFunc func() []string) (regauthn.Keychain, error) {
+	if environFunc == nil {
+		environFunc = os.Environ
+	}
```
cppforlife: should this be pushed down to EnvKeychain?

pkg/imgpkg/registry/registry.go
```diff
@@ -34,17 +34,64 @@ type Opts struct {
...
+	WriteTag(tag regname.Tag, taggable regremote.Taggable) error
+
+	ListTags(repo regname.Repository) ([]string, error)
+
+	CloneWithSingleAuth(imageRef regname.Tag) (Registry, error)
+}
+
+// ImagesMetadata Interface for Reading Images
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . ImagesMetadata
+type ImagesMetadata interface {
```
cppforlife: should this be called ImagesReader to be consistent with ImagesReaderWriter naming?

pkg/imgpkg/registry/with_progress.go
```diff
@@ -58,3 +60,22 @@ func (w WithProgress) WriteIndex(reference regname.Reference, index regv1.ImageI
...
+// A Registry need to be provided as the first parameter or the function will panic
+func (w WithProgress) CloneWithSingleAuth(imageRef regname.Tag) (Registry, error) {
+	delegate, err := w.delegate.CloneWithSingleAuth(imageRef)
+	if err != nil {
+		return nil, err
+	}
+
+	return &WithProgress{
+		delegate: delegate,
+	}, nil
```
cppforlife: style nit: seems like a oneliner

pkg/imgpkg/registry/registry.go
```diff
@@ -34,17 +34,64 @@ type Opts struct {
...
+// SimpleRegistry Implements Registry interface
+type SimpleRegistry struct {
+	remoteOpts []regremote.Option
+	refOpts    []regname.Option
+	keychain   regauthn.Keychain
 }
 
-func NewRegistry(opts Opts, regOpts ...regremote.Option) (Registry, error) {
+// NewRegistry Builder for a Simple Registry
+func NewRegistry(opts Opts, regOpts ...regremote.Option) (*SimpleRegistry, error) {
```
cppforlife: NewRegistry -> NewSimpleRegistry

cppforlife (commented):
aside from minor comments, lgtm

pkg/imgpkg/registry/keychain.go
joaopapereira: Created a builder for the EnvKeychain that takes care of this, good catch

pkg/imgpkg/registry/registry.go
joaopapereira: Fixed

pkg/imgpkg/registry/with_progress.go
joaopapereira: fixed

pkg/imgpkg/registry/registry.go
joaopapereira: fixed

cppforlife:
@joaopapereira can you squash "Change in ggcr..." and "Revert..."?

joaopapereira:
> @joaopapereira can you squash "Change in ggcr..." and "Revert..."?

I left them there so that we know how might we run the skipped tests. If you think that it is not very relevant I can squash it

Issue #231, opened 2021-09-01
Allow specifying the number of retries


voor:
I am trying to do the unachievable, which is download gigabytes of data through an enterprise proxy server.  It's failing, a lot.  I've tried changing the concurrency, and that seems to help, but things are still failing.  I know if I keep trying it'll eventually work, and rather than creating a bash script to just forever loop this, I'd like to pass something into imgpkg.

Possibly add an option like `--num-retries=XX` or another logical flow that would allow me to `--just-go-until-it-works` might help this problem.

Note:
Slack conversation: https://kubernetes.slack.com/archives/CH8KCCKA5/p1630527017057500

---

Related work summary:

- [x] https://github.com/google/go-containerregistry/issues/1114
- [x] https://github.com/google/go-containerregistry/pull/1122

DennisDenuto:
Related to: https://github.com/google/go-containerregistry/issues/1114

pivotaljohn:
Triage:
- split out work to change the imgpkg default to 15.
- @DennisDenuto and @cppforlife to huddle to decide whether this is configurable to the imgpkg user.

vrabbi:
this would be great to have. im trying to create a VERY large imgpkg bundle for an airgapped scenario with 942 images in it. this is a package repository. the file size is 67GB when copied to TAR. eventually after about 1000 layers there is usually some temporary network issue which unfortunately isnt fixed in time for the current retry number to take continue on successfully.
copying to tar is taking ~3+ hours and has yet to finish because of issues like this. having the retry count configurable would make this much smoother

aaronshurley:
Just as a reference, the gcr work was in the last release: https://github.com/google/go-containerregistry/releases/tag/v0.7.0

joaopapereira:
The PR was merged and will be released in version 0.24.0

joaopapereira:
Released on version 0.24.0

PR #299, opened 2021-11-16
Allow users to defined number of retries

joaopapereira:
Add a new flag `--registry-retry-count` that enables the users to define the number of retries done by `imgpkg` when a temporary network error happens while talking with the registry

Related #231

### DIFF
--- a/pkg/imgpkg/cmd/registry_flags.go
+++ b/pkg/imgpkg/cmd/registry_flags.go
@@ -22,9 +22,12 @@ type RegistryFlags struct {
 	Token    string
 	Anon     bool
 
+	RetryCount int
+
 	ResponseHeaderTimeout time.Duration
 }
 
+// Set Registers the flags available to the provided command
 func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().StringSliceVar(&r.CACertPaths, "registry-ca-cert-path", nil, "Add CA certificates for registry API (format: /tmp/foo) (can be specified multiple times)")
 	cmd.Flags().BoolVar(&r.VerifyCerts, "registry-verify-certs", true, "Set whether to verify server's certificate chain and host name")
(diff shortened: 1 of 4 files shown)


pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -36,6 +39,7 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&r.Anon, "registry-anon", false, "Set anonymous auth ($IMGPKG_ANON)")
 
 	cmd.Flags().DurationVar(&r.ResponseHeaderTimeout, "registry-response-header-timeout", 30*time.Second, "Maximum time to allow a request to wait for a server's response headers from the registry (ms|s|m|h)")
+	cmd.Flags().IntVar(&r.NumberRetries, "registry-num-retries", 5, "Set the number of times imgpkg retries to send requests to the registry in case of an error")
```
cppforlife: may be --registry-retry-num/count? would be nice to group retry settings under --registry-retry-*

joaopapereira: Fixed it

pkg/imgpkg/registry/registry.go
```diff
@@ -122,6 +122,13 @@ func NewSimpleRegistry(opts Opts, regOpts ...regremote.Option) (*SimpleRegistry,
 		regRemoteOptions = append(regRemoteOptions, regOpts...)
 	}
 
+	regRemoteOptions = append(regRemoteOptions, regremote.WithRetryBackoff(regremote.Backoff{
+		Duration: 1.0 * time.Second,
+		Factor:   3.0,
+		Jitter:   0.1,
+		Steps:    opts.RetryCount,
+	}))
```
cppforlife: i think that's where im leaning towards is we should just keep retry backoff more linear, retry after [100ms, 300ms, 500ms, 1s (cap at 1s)]

PR #301, opened 2021-11-19
Do not access origin registry when copying from tar

### LINKED ISSUE #300: Copy from tar to repo fails on airgapped environment
**What steps did you take:**
Create a Bundle and copy it to a tar file
Move the tar to a place that has no access to the original place where the Bundle was created

**What happened:**
```
imgpkg: Error: Updating Image Refs registry.local:5000/service-bindings/bundle@sha256:8db097671917f25f32d5a7afe54f95a0c112728e024c9f80524d45b11438ea82:
  Fetching images of registry.local:5000/service-bindings/bundle@sha256:8db097671917f25f32d5a7afe54f95a0c112728e024c9f80524d45b11438ea82:
    Fetching location image:
      Get "https://registry.local:5000/v2/": dial tcp: lookup registry.local on [2001:558:feed::1]:53: no such host; Get "http://registry.local:5000/v2/": dial tcp: lookup registry.local on [2001:558:feed::1]:53: no such host

```

**What did you expect:**
`imgpkg` not to try to contact the origin Registry

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.23.0
- Docker registry used (e.g. `Docker HUB`): Dockerhub

---
Vote on this request

joaopapereira:
Fixes #300

### DIFF
--- a/pkg/imgpkg/bundle/bundle_images_lock.go
+++ b/pkg/imgpkg/bundle/bundle_images_lock.go
@@ -43,8 +43,8 @@ func (o *Bundle) AllImagesRefs(concurrency int, ui util.UIWithLevels) ([]*Bundle
 	return bundles, allImageRefs, err
 }
 
-// UpdateImageRefs updates the bundle cached images
-func (o *Bundle) UpdateImageRefs(bundles []*Bundle, ui util.UIWithLevels) error {
+// UpdateImageRefs updates the bundle cached images without talking to the registry
+func (o *Bundle) UpdateImageRefs(bundles []*Bundle) error {
 	o.cachedImageRefs = map[string]ImageRef{}
 
 	img, err := o.checkedImage()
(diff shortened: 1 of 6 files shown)


test/e2e/copy_from_tar_test.go
```diff
@@ -151,4 +151,31 @@ func TestCopyTarSrc(t *testing.T) {
...
+		env := helpers.BuildEnv(t)
+		imgpkg := helpers.Imgpkg{T: t, L: helpers.Logger{}, ImgpkgPath: env.ImgpkgPath}
+		defer env.Cleanup()
+
+		destinationFakeRegistry := helpers.NewFakeRegistry(t, &helpers.Logger{LogLevel: helpers.LogDebug})
+		defer destinationFakeRegistry.CleanUp()
+
+		var tempBundleTarFile string
+		logger.Section("create tar from bundle", func() {
+			originFakeRegistry := helpers.NewFakeRegistry(t, &helpers.Logger{LogLevel: helpers.LogDebug})
```
cppforlife: can we have an assertion on FakeRegistry checking that it was not accessed at all.

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -137,7 +137,7 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 	return bundles, processedImageRefs, nil
 }
 
-func (o *Bundle) fetchImagesRef(img regv1.Image, ui util.UIWithLevels) (ImageRefs, error) {
+func (o *Bundle) fetchImagesRef(img regv1.Image, ui util.UIWithLevels, offline bool) (ImageRefs, error) {
```
cppforlife: im ok with this for now in favor of getting the build out, but we should probably discuss a more elegant approach.

test/e2e/copy_from_tar_test.go
joaopapereira: I can add a comment in the next line, but the CleanUp stops the server when this `logger.Section` ends. So the server will not be on.

Issue #331, opened 2022-02-04
Introduce tag that gives visibility into source image / repository


joshrosso:
When using `imgpkg copy -b`, especially recursively against many packages, we end up with a repository full of tags. This is great, but makes things hard to identify without extra tooling.

For example:

![image](https://user-images.githubusercontent.com/6200057/152581938-5dd5c59f-9d70-4341-94f0-18e6b56d91e1.png)

Ideally, there'd be a tag that could help us better identify the source image. For example:

![image](https://user-images.githubusercontent.com/6200057/152582004-ac484353-31a1-42ef-a9e0-cf766bbd6a6e.png)

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help work on this issue.

alexgaganashvili:
@joshrosso , could you provide more context/details into this issue? I'm just wondering how I can reproduce it so that I can better understand what should be improved.
Thanks.

cppforlife:
> could you provide more context/details into this issue

if you run `imgpkg copy -b ... --to-repo bar.com/foo/blah/dst/repo` then all assets are copied over into the same repo (e.g. /foo/blah/dst/repo). such images will have standard imgpkg tags like `sha256-....imgpkg` and will not have any indication from the registry perspective that it was cert-manager image for example.

a practical improvement that could be made is: in addition to regular imgpkg tags, one can add "mangled" human friendly tag (e.g. based on screentshot above, tag could be: `tce-cert-manager-<some randome bits>`).

alexgaganashvili:
@cppforlife , thanks for providing more details.
Here's a PR: https://github.com/vmware-tanzu/carvel-imgpkg/pull/333

joaopapereira:
I was talking with Dmitriy and we were thinking about the impact of `imgpkg` creating 2 tags per image. In our point of view, it feels like we do not need 2 tags.
The suggestion that we would like to do is to instead replace the current tag that is generated by `imgpkg` with a new format.

Let's imagine the following image:
`my.registry.io/some/application@sha256:669e010b58baf5beb2836b253c1fd5768333f0d1dbcb834f7c07a4dc93f474be`
[...]

PR #333, opened 2022-02-16
Add legible tags issue 331.

alexgaganashvili:


pkg/imgpkg/util/tag.go
```diff
@@ -28,3 +30,20 @@ func BuildDefaultUploadTagRef(item WithDigest, importRepo regname.Repository) (r
 	}
 	return uploadTagRef, nil
 }
+
+// BuildLegibleUploadTagRef Builds a legible tag from original image reference
+func BuildLegibleUploadTagRef(itemRef string, importRepo regname.Repository) (regname.Tag, error) {
+	// ex.:
+	// 1. registry.f.q.d.n/myorg/my-img@sha256:012345679e2a06317dc02abcde123759eb9e6105a0731224a0c63898abcde123
+	// 2. registry.f.q.d.n/myorg/my-img:some-tag@sha256:012345679e2a06317dc02abcde123759eb9e6105a0731224a0c63898abcde123
+	origRepoAndTag := strings.Split(itemRef, "@")
```
joaopapereira: This can be accomplished with 
```go
digest, err := regname.NewDigest(itemRef)

origRepoPath := a.RepositoryStr()
dashedRepo := strings.ReplaceAll(origRepoPath, "/", "-")
tag := fmt.Sprintf("%s-%s-%s.imgpkg", dashedRepo[len(dashedRepo)-49:], digest.Algorithm, digest.Hex)
```

I did not try to compile this, so please take it with a grain of salt, also there is the problem of `dashedRepo` being smaller than 49

joaopapereira (changes requested):
The issue was changed to accommodate a different syntax, do you mind updating your PR to match the new requirements?

pkg/imgpkg/imageset/image_set.go
```diff
@@ -31,7 +31,7 @@ func NewImageSet(concurrency int, ui goui.UI) ImageSet {
 }
 
 func (i ImageSet) Relocate(foundImages *UnprocessedImageRefs,
-	importRepo regname.Repository, registry registry.ImagesReaderWriter) (*ProcessedImages, error) {
+	importRepo regname.Repository, registry registry.ImagesReaderWriter, useRepoBasedTags bool) (*ProcessedImages, error) {
```
joaopapereira: Instead of the Relocate to take a boolean flag I personally would prefer to have

```go
type TagGenerator interface {
   Generate(sourceDigest regname.Digest, destinationRepo regname.Repository) (regname.Tag, error)
}

func NewImageSet(concurrency int, ui goui.UI, tagGen TagGenerator) ImageSet {
	return ImageSet{concurrency, UI, tagGen}
}
```

This would allow us to have eventually other types of tag generation in the future if we want and the decision of a particular tag type is not made deep inside the copy code but is done by the caller of ImageSet

pkg/imgpkg/imageset/image_set.go
```diff
@@ -146,27 +148,40 @@ func checkForAnyAsyncErrors(imgOrIndexes []imagedesc.ImageOrIndex, errCh chan er
...
 	if err != nil {
-		return regname.Tag{}, nil, err
+		return []regname.Tag{}, nil, err
 	}
 
-	var artifactToWrite regremote.Taggable
-	switch {
-	case item.Image != nil:
-		artifactToWrite, err = i.mountableImage(*item.Image, uploadTagRef, registry)
+	tagArr := make([]regname.Tag, 1, 2)
```
joaopapereira: I think we talked about only having 1 tag, so we do not need a slice here

joaopapereira, line 157: I think we talked about only having 1 tag, so we do not need a slice here

pkg/imgpkg/internal/util/tag.go
```diff
@@ -29,3 +30,27 @@ func BuildDefaultUploadTagRef(item WithDigest, importRepo regname.Repository) (r
 	}
 	return uploadTagRef, nil
 }
+
+// BuildLegibleUploadTagRef Builds a legible tag from original image reference
+func BuildLegibleUploadTagRef(item WithDigest, itemRef string, importRepo regname.Repository) (regname.Tag, error) {
```
joaopapereira: I was looking at this code and I am not 100% sure this works in the scenario:
1. copy image from registry A to registry B
2. copy image from registry B to registry C

Looking at the parameters that are provided to this function I believe that `itemRef` on the second copy will be the reference from registry B. 

What we need is the original location of the Image. This might require more changes in the `imagedesc` package to save the original location of the Image as part of the ImageDescriptor type.

Can you validate if the case I pointed out in this comment is really happening?
Let me know if this comment is a bit to much information and might confuse you, I am free to jump on a Zoom call and look into this issue with you if it helps

joaopapereira, line 35: I was looking at this code and I am not 100% sure this works in the scenario:
1. copy image from registry A to registry B
2. copy image from registry B to registry C

Looking at the parameters that are provided to this function I believe that `itemRef` on the second copy will be the reference from registry B. 

What we need is the original location of the Image. This might require more changes in the `imagedesc` package to save the original location of the Image as part of the ImageDescriptor type.

Can you validate if the case I pointed out in this comment is really happening?
Let me know if this comment is a bit to much information and might confuse you, I am free to jump on a Zoom call and look into this issue with you if it helps

joaopapereira (changes requested):
I added some comments on this PR. Not 100% sure if this implementation will work on all cases. Let me know if you need more information about my comments.

pkg/imgpkg/imageset/image_set.go
alexgaganashvili: Got it. That's because of my comment: https://github.com/vmware-tanzu/carvel-imgpkg/issues/331#issuecomment-1046136094

pkg/imgpkg/cmd/copy.go
```diff
@@ -74,6 +84,8 @@ func NewCopyCmd(o *CopyOptions) *cobra.Command {
 	cmd.Flags().IntVar(&o.Concurrency, "concurrency", 5, "Concurrency")
 	cmd.Flags().BoolVar(&o.IncludeNonDistributable, "include-non-distributable-layers", false,
 		"Include non-distributable layers when copying an image/bundle")
+	cmd.Flags().BoolVar(&o.UseRepoBasedTags, "use-repo-based-tags", false,
```
cppforlife: use-repo-based-tags => repo-based-tags (to be more consistent with other carvel tool flag naming)

pkg/imgpkg/internal/util/tag.go
```diff
@@ -5,16 +5,103 @@ package util
...
+	tag := strings.ReplaceAll(dashedRepo, ":", "-")
+	uploadTagRef, err := regname.NewTag(fmt.Sprintf("%s:%s", importRepo.Name(), tag))
+	if err != nil {
+		return regname.Tag{}, fmt.Errorf("building repo-based tag: %s", err)
+	}
+	return uploadTagRef, nil
+}
+
+// NewTagGenerator returns different TagGenerator types
+func NewTagGenerator(tagGenType string) TagGenerator {
```
cppforlife: instead of introducing this function lets just instantiate RepoBasedTagGenerator or DefaultTagGenerator in call sites.

test/helpers/env.go
```diff
@@ -13,15 +13,16 @@ import (
...
-	BundleFactory  BundleFactory
-	Assets         *Assets
-	Assert         Assertion
-	ImageFactory   ImageFactory
-	Logger         *Logger
-	cleanupFuncs   []func()
+	Image           string
+	ImgpkgPath      string
+	RelocationRepo  string
+	RelocationRepo2 string
```
cppforlife: what is RelocationRepo2?

alexgaganashvili: Forgot to remove that one from env.go. I don't need it. Thx.

joaopapereira (approved):
LGTM

pkg/imgpkg/imagedesc/types.go
```diff
@@ -21,8 +21,9 @@ type ImageOrIndex struct {
 	Index *ImageIndexWithRef
 
 	Labels map[string]string
-}
 
+	OrigRef string
```
joaopapereira:
```suggestion
	OrigRef string `json:"-"`
```

pkg/imgpkg/imagedesc/types.go
```diff
@@ -57,6 +58,7 @@ type ImageIndexDescriptor struct {
 	Digest    string
 	Raw       string
 	Tag       string
+	OrigRef   string
```
joaopapereira, line 61:
```suggestion
	OrigRef   string `json:"-"`
```

joaopapereira (changes requested):
I made these 2 suggestions after talking with @cppforlife. By doing this we ensure that the same tar created with an old version of imgpkg and a version with this feature will have the same `sha256sum`

pkg/imgpkg/imagedesc/types.go
joaopapereira: Ignore this

pkg/imgpkg/imagedesc/types.go
joaopapereira: Ignore this

joaopapereira (approved):
LGTM

PR #334, opened 2022-02-16
Reuse transport when talking with registries

joaopapereira:
This PR adds makes an improvement on the communication with the registries. Previously for each request done to the registry, a new transport was created. The main issue is that the authentication part was lost between requests.
We are introducing a store that will maintain the transport information that will be reused whenever possible, this will lower the number of times that `imgpkg` needs to talk with authentication servers.

### DIFF
--- a/pkg/imgpkg/registry/registry.go
+++ b/pkg/imgpkg/registry/registry.go
@@ -11,12 +11,15 @@ import (
 	"net/http"
 	"regexp"
 	"runtime"
+	"sync"
 	"time"
 
 	regauthn "github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/logs"
 	regname "github.com/google/go-containerregistry/pkg/name"
 	regv1 "github.com/google/go-containerregistry/pkg/v1"
 	regremote "github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/registry/auth"
 )
 
@@ -82,9 +85,11 @@ var _ Registry = &SimpleRegistry{}
 
 // SimpleRegistry Implements Registry interface
 type SimpleRegistry struct {
-	remoteOpts []regremote.Option
-	refOpts    []regname.Option
-	keychain   regauthn.Keychain
+	remoteOpts      []regremote.Option
+	refOpts         []regname.Option
+	keychain        regauthn.Keychain
+	roundTrippers   *RoundTripperStorage
+	transportAccess *sync.Mutex
 }
 
 // NewSimpleRegistry Builder for a Simple Registry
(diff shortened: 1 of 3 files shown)


pkg/imgpkg/registry/transport.go
```diff
@@ -0,0 +1,109 @@
...
+func NewRoundTripperStorage(baseRoundTripper http.RoundTripper) *RoundTripperStorage {
+	return &RoundTripperStorage{
+		baseRoundTripper: baseRoundTripper,
+		readWriteAccess:  &sync.Mutex{},
+		transports:       map[string]map[string]map[string]http.RoundTripper{},
+	}
+}
+
+// RoundTripper Retrieve the RoundTripper to be used for a particular registry and repository or nil if it cannot be found
+func (r *RoundTripperStorage) RoundTripper(registry, repository, method string) http.RoundTripper {
```
cppforlife: instead of strings can we take one of the ggcr objcects from where we can access registry and repo.

pkg/imgpkg/registry/transport.go
```diff
@@ -0,0 +1,109 @@
...
+	r.readWriteAccess.Lock()
+	defer r.readWriteAccess.Unlock()
```
cppforlife, line 37: nit: newline after defer, to make it look like it's applied to entire function

pkg/imgpkg/registry/transport.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+		return nil
+	}
+
+	return r.transports[registry][repository][method]
+}
+
+// Exists Checks if a RoundTripper exists for the registry and repository
+func (r *RoundTripperStorage) Exists(registry, repository, method string) bool {
```
cppforlife: can we combine Exists with RoundTripper above?

pkg/imgpkg/registry/transport.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, fmt.Errorf("Unable to create round tripper: %s", err)
+	}
+
+	if _, ok := r.transports[reg.RegistryStr()]; !ok {
+		r.transports[reg.RegistryStr()] = map[string]map[string]http.RoundTripper{}
+	}
+	s := strings.Split(scope, ":")
+	if len(s) != 3 {
+		panic(fmt.Sprintf("Internal inconsistency: expected scope '%s' to have 3 fields", scope))
+	}
```
cppforlife, line 77: what does the format of scope look like (lets add example comment). are scopes hard coded in the standard, or is that something that may change over time?

pkg/imgpkg/registry/registry.go
```diff
@@ -149,42 +167,87 @@ func (r SimpleRegistry) CloneWithSingleAuth(imageRef regname.Tag) (Registry, err
...
+	// the same RoundTripper multiple times
+	r.transportAccess.Lock()
+	defer r.transportAccess.Unlock()
+	rt := r.roundTrippers.RoundTripper(ref.Context().RegistryStr(), ref.Context().RepositoryStr(), scope)
+	if rt == nil {
+		resolvedAuth, err := r.keychain.Resolve(ref.Context())
+		if err != nil {
+			return nil, fmt.Errorf("Unable retrieve credentials for registry: %s", err)
+		}
+		rt, err = r.roundTrippers.CreateRoundTripper(ref.Context().Registry, resolvedAuth, scope)
```
cppforlife, line 209: im still a bit concerned about keychain tokens expiring. what was the situation with that again?

pkg/imgpkg/registry/transport.go
joaopapereira: done

pkg/imgpkg/registry/transport.go
joaopapereira: done

pkg/imgpkg/registry/transport.go
joaopapereira: I think we can remove it, I was using it before but apparently not anymore

pkg/imgpkg/registry/transport.go
joaopapereira: done

pkg/imgpkg/registry/registry.go
joaopapereira: Before doing any request to the registry ggcr will `ping` the registry to understand if it still has access to it, if not it will try and go through the oauth flow again.
You can follow the interaction starting on the Get function
Calling [makeFetcher](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/descriptor.go#L219)
Calls [transport.NewWithContext](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/transport.go#L38)
Which in turn calls [ping](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/transport/ping.go#L71)

cppforlife:
lgtm

PR #336, opened 2022-02-17
authn env vars when copying between two registries

alexgaganashvili:
I was thinking of suggesting using https://carvel.dev/imgpkg/docs/`latest`/auth/#via-environment-variables instead of https://carvel.dev/imgpkg/docs/`v0.24.0`/auth/#via-environment-variables in that note. But the redirect strips off `#via-environment-variables`


joaopapereira (approved):
LGTM

PR #338, opened 2022-02-17
allow using dockerhub proxy during testing

alexgaganashvili:


pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -534,9 +534,14 @@ func TestToRepoBundleCreatesValidLocationOCI(t *testing.T) {
 	fakeRegistry := helpers.NewFakeRegistry(t, logger)
 	defer fakeRegistry.CleanUp()
 
+	dockerhubProxy := "index.docker.io"
```
joaopapereira: This logic looks like it should be moved to `test/helpers/env.go` this is the struct that handles all the environment variable setup

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -615,7 +620,7 @@ func TestToRepoBundleCreatesValidLocationOCI(t *testing.T) {
 			APIVersion: "imgpkg.carvel.dev/v1alpha1",
 			Kind:       "ImageLocations",
 			Images: []bundle.ImageLocation{{
-				Image:    "index.docker.io/library/hello-world@sha256:ebf526c198a14fa138634b9746c50ec38077ec9b3986227e79eb837d26f59dc6",
+				Image:    dockerhubProxy + "/library/hello-world@sha256:ebf526c198a14fa138634b9746c50ec38077ec9b3986227e79eb837d26f59dc6",
```
joaopapereira: This maybe can be a call similar to:
```go
env.CompleteImageRef("hello-world@sha256:ebf526c198a14fa138634b9746c50ec38077ec9b3986227e79eb837d26f59dc6")
```
This function would add the `index.docker.io` or the proxy depending on the environment variable.

I am suggesting this because it feels like sprinkling this logic in different tests makes it harder to follow as a pattern while if we do this as part of a function call it might be easier to understand.
While I was writing this an idea came to my head, maybe the registry builder could be used to make this translation for us because in that case it would be transparent to the person writing the test and would be applicable to all the tests with a change in a single place (I did not validate the feasibility of this, it is just an idea if you want to explore it)

joaopapereira (changes requested):
Thanks for making this PR, I added some comments.

For extra credit, we have been thinking of eventually adding some tests that ensure that using a proxy is working as expected. If this is something that you would be interested in doing it would be appreciated. A couple of tests would be enough.

pkg/imgpkg/cmd/copy_repo_src_test.go
alexgaganashvili: Yes, good idea. Need to give it a thought how to make it transparent.

pkg/imgpkg/cmd/copy_repo_src_test.go
alexgaganashvili: I have moved that to helpers.go as functions.

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -62,6 +62,49 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+func TestExpectedRegistry(t *testing.T) {
```
joaopapereira: This test should be relocated to test/helpers since it is testing the implementation on that package

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -62,6 +62,49 @@ func TestMain(m *testing.M) {
...
+	os.Setenv("DOCKERHUB_PROXY", "my-dockerhub-proxy.tld/dockerhub-proxy")
+	assert.Equal(t, "my-dockerhub-proxy.tld/dockerhub-proxy", helpers.GetDockerHubRegistry())
+	os.Unsetenv("DOCKERHUB_PROXY")
+
+	if isSet {
+		os.Setenv("DOCKERHUB_PROXY", origProxyVal)
+	}
+}
+
+func TestExpectedImgRef(t *testing.T) {
```
joaopapereira: This test should be relocated to test/helpers since it is testing the implementation on that package

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -62,6 +62,49 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+func TestExpectedRegistry(t *testing.T) {
+	origProxyVal := ""
+
+	v, isSet := os.LookupEnv("DOCKERHUB_PROXY")
+	if isSet {
+		origProxyVal = v
```
joaopapereira: Just a suggestion, you could defer the setenv to the original value
```go
defer os.Setenv("DOCKERHUB_PROXY", origProxyVal)
```
This way you could remove the if in line 80.

If you do this you will ensure that the setenv always runs even if an error happens in the test

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -62,6 +62,49 @@ func TestMain(m *testing.M) {
...
+		os.Setenv("DOCKERHUB_PROXY", origProxyVal)
+	}
+}
+
+func TestExpectedImgRef(t *testing.T) {
+	origProxyVal := ""
+
+	v, isSet := os.LookupEnv("DOCKERHUB_PROXY")
+	if isSet {
+		origProxyVal = v
```
joaopapereira: same comment as the previous test

joaopapereira (changes requested):
Looks good if you could just move the tests to the other package I think we are ready to merge the PR

joaopapereira (approved):
LGTM

Issue #204, opened 2021-07-28
List bundle dependencies


ashvinmoro:
**Describe the problem/challenge you have**
***List dependencies:***
An imgpkg bundle can either be a standalone bundle or a package repository, with them referencing  dependent artifacts which can either be OCI artifacts or other imgpkg bundles. The imgpkg CLI today does not provide an option to query a bundle to list all dependent artifacts (regardless of the level of nesting) without doing a pull and parsing through the contents of the bundle.

This requires knowing the internals of the imgpkg bundle layout which is an implementation detail and causes too tight a coupling for clients consuming the CLI.

This feature is very helpful to inspect contents of the bundle metadata without requiring to download it, similar to how artifact manifest can be inspected using the crane command.

**Describe the solution you'd like**
Provide a CLI option and an API to allow clients consuming the bundles to query for all dependent artifacts and their types (bundle, repo, image) with their references locked to the relocated repository, similar to how “imgpkg pull” would fix the images lock.

Example:
`imgpkg -b <bundle-ref> --list-deps`
returns images referenced after the images are locked
`Images referenced in the ImagesLock of the bundle:`
`index.docker.io/test/package-repo@sha256:b5e1cfb8731bf5ba6a0fb7060c37bbca03e9fb12be56dbec4a1fe2bbdfff2bd3`
`index.docker.io/test/package-repo@sha256:42030f63b66e25a03c7d9ef28fa5369d230df5697b4ab2d94860e3f69026d18a`

**Anything else you would like to add:**
None

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

ashvinmoro:
cc: @ncdc

DennisDenuto:
Thanks for creating this issue.

> Provide a CLI option and an API... to query for... types

The CLI option can be added to the `pull` command. If this flag is present, the last section of output can be a flat list of all the dependent images that were pulled (also whether the dependent image is a bundle or an OCI image) Assuming we display whether the dependent image is a bundle or OCI image, would you still expect an API to 'query' the type?

Are you able to expand on the use case driving this request? or, why do you want a full list of dependent images per bundle?

[...]

PR #341, opened 2022-02-23
Describe command

joaopapereira:
This PR introduces the `describe` command in 2 different flavors, an API call, and a command-line option.
This PR does not implement the totality of #204 only provides the basic structure for the command and image information, at this point does not collect the information from the bundle metadata file, nor does it implement the filtering options that were designed in that story since we are not 100% sure if they are needed or helpful for our users.

What is implemented:
- describe command with output both in text and yaml formats
- move some utilities to internal package
- Fix bug that was copying annotations between bundles
- Create an API package

Example output with nested bundles
```
Bundle SHA: sha256:f35a6d5e5596919c6bd4f62164ee6f8ccd919d0d8a04b3a5fb382af33dd7da9d

Images:
  Image: localhost:5000/describe-test-collocated@sha256:5c2dafe3c70c13990190d643c91e9f67b8129b179257674888178868474f6511
  Type: Bundle
  Origin: localhost:5000/inner-bundle@sha256:5c2dafe3c70c13990190d643c91e9f67b8129b179257674888178868474f6511
  Images:
    Image: localhost:5000/describe-test-collocated@sha256:4c8b96d4fffdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
    Type: Image
    Origin: index.docker.io/dkalinin/k8s-simple-app@sha256:4c8b96d4fffdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
    Annotations:
      kbld.carvel.dev/id: docker.io/dkalinin/k8s-simple-app
  Image: localhost:5000/describe-test-collocated@sha256:4c8b96d4fffdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
  Type: Image
  Origin: index.docker.io/dkalinin/k8s-simple-app@sha256:4c8b96d4fffdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
  Annotations:
    kbld.carvel.dev/id: docker.io/dkalinin/k8s-simple-app

Succeeded
```

Related to #204


pkg/imgpkg/internal/util/level_logger.go
```diff
@@ -10,18 +10,19 @@ type LogLevel int
...
-	goui.UI
-
 	Errorf(msg string, args ...interface{})
 	Warnf(msg string, args ...interface{})
 	Debugf(msg string, args ...interface{})
 	Tracef(msg string, args ...interface{})
 }
 
 const (
+	// LogTrace most verbose level
```
cppforlife, line 20: seems a bit suspicious to put util into internal. dont we have classes that want to use logger (and if they do we really want ppl to impl their own for the interface)?

pkg/imgpkg/api/describe_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+			break
+		}
+	}
+	if bundle == nil {
+		panic("Internal consistency: bundle could not be found in list of bundles")
+	}
+
+	imagesRefs := bundle.ImagesRefs()
+	sort.Slice(imagesRefs, func(i, j int) bool {
+		return strings.Compare(imagesRefs[i].Image, imagesRefs[j].Image) <= 0
```
cppforlife: nit: `imagesRefs[i].Image < imagesRefs[j].Image`

pkg/imgpkg/api/describe_bundle.go
```diff
@@ -0,0 +1,155 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package api
```
cppforlife: should we move this into bundle package?

pkg/imgpkg/api/describe_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+type ImageInfo struct {
+	Image       string            `json:"image"`
+	Origin      string            `json:"origin"`
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// BundleContent Contents present in a Bundle
+type BundleContent struct {
+	Bundles []BundleDescription `json:"bundles,omitempty"`
+	Images  []ImageInfo         `json:"images,omitempty"`
```
cppforlife: does this include signature images and image-locations?

pkg/imgpkg/api/describe_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	if !isBundle {
+		return BundleDescription{}, fmt.Errorf("The image %s is not a bundle", bundleImage)
+	}
+
+	allBundles, _, err := bundle.AllImagesRefs(opts.Concurrency, opts.Logger)
+	if err != nil {
+		return BundleDescription{}, err
+	}
```
cppforlife: this to me feels like something that should be happening within the command itself.

pkg/imgpkg/internal/util/level_logger.go
joaopapereira: I did do a full swipe and moved all our utils into the internal packages as they are internal things that the users might not care, but I can pull the logger out

pkg/imgpkg/api/describe_bundle.go
joaopapereira: I thought about that but my objective was to create a clear interface that we could consider stable and tell people that they can use when interacting with imgpkg as a library.

pkg/imgpkg/api/describe_bundle.go
joaopapereira: Not right now. We can add them as well, maybe it makes sense since it might be useful for the callers

pkg/imgpkg/internal/util/level_logger.go
cppforlife: ok, sounds good, separate PR?

pkg/imgpkg/api/describe_bundle.go
joaopapereira: These checks and validations as well as the creation of the registry is inside this function to ensure that the caller of our API can easily call the function without having to do a big setup before hand

pkg/imgpkg/internal/util/level_logger.go
joaopapereira: I did it in a separate commit in this PR, but if you want I can separate this change from this PR

pkg/imgpkg/api/describe_bundle.go
cppforlife: > maybe it makes sense since it might be useful for the callers

i think we should since consumers may want to do something with _all_ images that are in the bundle, and by _all_ i mean anything that gets touched by imgpkg copy.

joaopapereira: I am not 100% sure about the location's image since it is something that is internal to imgpkg, but I will add the signatures.

pkg/imgpkg/api/describe_bundle.go
cppforlife: > tell people that they can use when interacting with imgpkg as a library.

i don't doubt that some users only need this feature (we know of one consumer), but entirety of pkg/imgpkg/bundle and pkg/imgpkg/registry is the client library i would expect consumers to use.

pkg/imgpkg/api/describe_bundle.go
cppforlife: it might be internal, but for example, if somebody needs to make sure that assets are all authorized, or something else, we definitely would want to include any "internal" images.

pkg/imgpkg/bundle/bundle_images_lock.go
```diff
@@ -130,10 +140,12 @@ func (o *Bundle) buildAllImagesLock(throttleReq *util.Throttle, processedImgs *p
 			bundles = append(bundles, nestedBundles...)
 
 			// Adds Image to the resulting ImagesLock
+			typedImgRef := NewContentImageRef(imgRef,
+				len(nestedBundles) > 0, // nestedBundles have Bundles when the image is a bundle
+			)
```
cppforlife: nit: this reads quite funky on multilines, mind squashing into a single line?

(same below for `processedImageRefs.AddImagesRef(`)

pkg/imgpkg/bundle/fetch_images.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"fmt"
+
+	regname "github.com/google/go-containerregistry/pkg/name"
+	"github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/internal/util"
+	"github.com/vmware-tanzu/carvel-imgpkg/pkg/imgpkg/lockconfig"
+)
+
+// FetchAllImagesRefs returns a flat list of nested bundles and every image reference for a specific bundle
+func (o *Bundle) FetchAllImagesRefs(concurrency int, ui util.UIWithLevels, sigFetcher SignatureFetcher) ([]*Bundle, error) {
```
cppforlife: util.* isnt accessible to external users (can we have a separate package under pkg/imgpkg/ so that we arent doing this accidently)

pkg/imgpkg/bundle/bundle.go
```diff
@@ -87,11 +107,7 @@ func (o *Bundle) findCachedImageRef(digestRef string) (ImageRef, bool) {
 }
 
 func (o *Bundle) allCachedImageRefs() []ImageRef {
```
cppforlife: i think we should remove this accessor since in bunch of places we are using o.cachedImageRefs directly.

pkg/imgpkg/signature/fetch_signatures.go
```diff
@@ -17,36 +19,65 @@ type Finder interface {
...
+			DigestRef: ref.Image,
+			Tag:       ref.Annotations["tag"],
+		})
+	}
+
+	return signatures, err
+}
+
+// FetchFromImageRef Retrieve the available signatures associated with the images provided
+func (s *Signatures) FetchFromImageRef(images []lockconfig.ImageRef) (map[string]lockconfig.ImageRef, error) {
```
cppforlife: FetchFromImageRef => FetchForImageRefs?

i would argue that we shouldnt be returning a map here. seems like an internal implementation. i dont see anyone consuming PrimaryLocation. (can still use a map to do uniquness check, but can return an array)

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -17,11 +17,42 @@ import (
...
+	// ContentImage Image that is part of the Bundle
+	ContentImage ImageType = "Image"
+	// SignatureImage Image that contains a signature
+	SignatureImage ImageType = "Signature"
+	// InternalImage Image that contains a signature
+	InternalImage ImageType = "Internal"
+)
+
+// NewContentImageRef Constructs a new ImageRef for Bundle or Image
+func NewContentImageRef(imgRef lockconfig.ImageRef, isBundle bool) ImageRef {
```
cppforlife: "naked" bool as an arg reads very confusing at the callee. i would recommend making this into:

~NewContentImageRef~
+NewBundleImageRef
+NewContentImageRef

(even in the enum you are making a distinction between ContentImage and BundleImage).

cppforlife: or just always use NewImageRefWithType, but remove bool from signature, so that isBundle is derived based on type.

pkg/imgpkg/bundle/images_refs.go
```diff
@@ -17,11 +17,42 @@ import (
...
 type ImageRef struct {
 	lockconfig.ImageRef
-	IsBundle *bool
+	IsBundle  *bool
+	Copiable  bool
+	ImageType ImageType
+}
+
+// ImageType defines the type of Image
+type ImageType string
```
cppforlife, line 26: one thing to note is that clients of imgpkg cannot assume that this is a final list of enum values. not sure how to best "enforce" that.

pkg/imgpkg/bundle/bundle.go
joaopapereira: done

pkg/imgpkg/signature/fetch_signatures.go
joaopapereira: done, I made it a slice all over

pkg/imgpkg/bundle/images_refs.go
joaopapereira: I added that information as part of the comment

pkg/imgpkg/bundle/images_refs.go
joaopapereira: done

PR #343, opened 2022-03-03
Apimachinery bump

joaopapereira:


hack/build.sh
```diff
@@ -16,7 +16,7 @@ go mod tidy
 # Specifically, the docker config.json is loaded before cli flags (and maybe even IaaS metadata services)
 git apply --ignore-space-change --ignore-whitespace ./hack/patch-k8s-pkg-credentialprovider.patch
 
-git diff --exit-code vendor/ || {
+git diff --exit-code vendor/github.com/vdemeester || {
```
cppforlife, line 19: we should still be checking for entire vendor directory.

joaopapereira: the problem with doing that is that locally any time you change a dependency and try to build imgpkg it fails and you need to stage or commit any change before you can compile. Which is not a good developer experience at all.

This check was added here for the purpose of checking that no changes were made to the package of the k8s credential providers since we are changing 1 line on it to turn off the docker config credentials loading. Personally, I think that we have to stop relying on the change we made to the vendored provider. This does not allow us to use `dependabot` to get dependencies updates since `dependabot` will try to restore the vendor folder to the "correct" state (in our case is without the change we made in the vendor library). So every time there is a bump on any library we have to create our own PR to do exactly what `dependabot`.

PR #361, opened 2022-03-23
Add in registry mutual tls

btrieger:
Some Docker Registries have mutual TLS on them meaning that you need to pass a client certificate and key in addition to the registry ca. I have added this functionality with 3 cli flags. --registry-mutual-tls(bool) default false , --registry-client-cert-path, and --registry-client-cert-key

### DIFF
--- a/pkg/imgpkg/cmd/registry_flags.go
+++ b/pkg/imgpkg/cmd/registry_flags.go
@@ -17,6 +17,10 @@ type RegistryFlags struct {
 	VerifyCerts bool
 	Insecure    bool
 
+	MutualTLS         bool
+	ClientCertificate string
+	ClientKey         string
+
 	Username string
 	Password string
 	Token    string
@@ -32,6 +36,10 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().StringSliceVar(&r.CACertPaths, "registry-ca-cert-path", nil, "Add CA certificates for registry API (format: /tmp/foo) (can be specified multiple times)")
 	cmd.Flags().BoolVar(&r.VerifyCerts, "registry-verify-certs", true, "Set whether to verify server's certificate chain and host name")
 	cmd.Flags().BoolVar(&r.Insecure, "registry-insecure", false, "Allow the use of http when interacting with registries")
+
+	cmd.Flags().BoolVar(&r.MutualTLS, "registry-mutual-tls", false, "Set whether or not to use mutual TLS. If true set registry-client-cert-path and registry-client-key-path")
+	cmd.Flags().StringVar(&r.ClientCertificate, "registry-client-cert-path", "", "Add Client Cert to authenticate against registry with (format: /tmp/foo)")
+	cmd.Flags().StringVar(&r.ClientKey, "registry-client-key-path", "", "Add Client Key to authenticate against registry with (format: /tmp/foo)")
 
 	cmd.Flags().StringVar(&r.Username, "registry-username", "", "Set username for auth ($IMGPKG_USERNAME)")
 	cmd.Flags().StringVar(&r.Password, "registry-password", "", "Set password for auth ($IMGPKG_PASSWORD)")
@@ -70,6 +78,10 @@ func (r *RegistryFlags) AsRegistryOpts() registry.Opts {
 		CACertPaths: r.CACertPaths,
 		VerifyCerts: r.VerifyCerts,
 		Insecure:    r.Insecure,
+
+		MutualTLS:         r.MutualTLS,
+		ClientCertificate: r.ClientCertificate,
+		ClientKey:         r.ClientKey,
 
 		Username: r.Username,
 		Password: r.Password,
--- a/pkg/imgpkg/registry/registry.go
+++ b/pkg/imgpkg/registry/registry.go
@@ -30,6 +30,10 @@ type Opts struct {
 
 	IncludeNonDistributableLayers bool
 
+	MutualTLS         bool
+	ClientCertificate string
+	ClientKey         string
+
 	Username string
 	Password string
 	Token    string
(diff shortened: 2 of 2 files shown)


btrieger:
I had a conversation with @joaopapereira on this and I have tested this on my local with nginx in front of a docker registry and am able to push images with mutual tls set to true and client and key provided. If I set it to false I get a 403 error which is what I receive if the client cert and key are not set. Let me know how else I can test this.

btrieger:
Expected Failure with mutual-tls false
```sh
./imgpkg copy --tar /Users/brieger/Documents/nginx.tar --to-repo 127.0.0.1:5043/nginx --registry-mutual-tls=false --registry-client-cert-path /Users/brieger/.docker/certs.d/127.0.0.1\:5043/client.cert --registry-client-key-path /Users/brieger/.docker/certs.d/127.0.0.1\:5043/client.key --registry-ca-cert-path /Users/brieger/.docker/certs.d/127.0.0.1\:5043/ca.crt --registry-username testuser --registry-password testpassword
copy | importing 1 images...

copy |
copy | done uploading images

imgpkg: Error: Error while preparing a transport to talk with the registry:
  Unable to create round tripper:
    GET https://127.0.0.1:5043/v2/: unexpected status code 400 Bad Request: <html>
<head><title>400 No required SSL certificate was sent</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<center>No required SSL certificate was sent</center>
<hr><center>nginx/1.21.6</center>
</body>
</html>
```
Expected Success with mutual tls true:
```sh
brieger-a01:carvel-imgpkg brieger$ ./imgpkg copy --tar /Users/brieger/Documents/nginx.tar --to-repo 127.0.0.1:5043/nginx --registry-mutual-tls=true --registry-client-cert-path /Users/brieger/.docker/certs.d/127.0.0.1\:5043/client.cert --registry-client-key-path /Users/brieger/.docker/certs.d/127.0.0.1\:5043/client.key --registry-ca-cert-path /Users/brieger/.docker/certs.d/127.0.0.1\:5043/ca.crt --registry-username testuser --registry-password testpassword
copy | importing 1 images...

426.03 MiB / 426.03 MiB [----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00% 1.53 GiB p/s
copy |
copy | done uploading images
426.03 MiB / 426.03 MiB [----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00% 1.53 GiB p/s

Succeeded
```

pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -33,6 +37,10 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&r.VerifyCerts, "registry-verify-certs", true, "Set whether to verify server's certificate chain and host name")
 	cmd.Flags().BoolVar(&r.Insecure, "registry-insecure", false, "Allow the use of http when interacting with registries")
 
+	cmd.Flags().BoolVar(&r.MutualTLS, "registry-mutual-tls", false, "Set whether or not to use mutual TLS. If true set registry-client-cert-path and registry-client-key-path")
```
cppforlife, line 40: by using flags we are making it only possible to use this for a single registry -- this may work well if fetching a single image, but is not great for anything beyond that (bundles for example may be referencing images in different registries).

can you mention what use case you are going after? what are you fetching (single image?)?

btrieger: ah I suppose yeah if we are pulling them from different registries. I suppose I could use a slice like we do for ca certs but then have to somehow keep the pairs together. I am fetching a bundle and pushing it to a single registry(that has mutual tls on it). Airgapped deployment of tanzu kubernetes grid in amazon-c2s

joaopapereira: We might need to allow something similar to https://carvel.dev/imgpkg/docs/v0.27.0/auth/#via-environment-variables as well in order to get this working for multiple registries.
This will have consequences as well on the Registry and when we are creating the transports because we would need to attach these certificates not on the base transport but on the a particular transport that is used for a particular registry

btrieger: @joaopapereira been searching around the codebase where is the _0, _1, _n for env support added for registry, password, and username. I only see it loading IMGPKG_USERNAME or IMGPKG_PASSWORD for example. 

Also, are you saying updating the newHTTPTransport function for each registry will not fix this as that is the base transport?

joaopapereira: Yes you are right, _0 and _1 are used to populate the keychain that is used when talking to the registry.

But in order to support MutalTLS for more than 1 registry we will need to extend the registry/transport.go structs to when they create a new round-tripper to select the appropriate keys based on the MutualTLS configuration.

Doesn't look very straightforward.
1. The keychain would have to signal that it uses MutualTLS to the `CreateRoundTripper` function
2. The baseRoundTripper attribute would have to be converted into a `http.Transport` or we can try to cast it on the fly so that the MutualTLS could be appropriately set on the transport provided to `transport.NewWithContext`
3. Keychain would have to know how to read the _0 for the MutualTLS

Let me know if this makes sense, I am also free to try to talk this through a little more

simonvpe:
My team is happy to use your patch @btrieger, so thank you for your work. I've submitted a [merge request](https://github.com/btrieger/carvel-imgpkg/pull/1) to make this work with v0.35.0.

```
❯ git log v0.35.0..HEAD
commit 3c1dc508e2b7488bb66ee1d3cdd63e5debe19a91 (HEAD -> mutual-tls, mine/mutual-tls)
Author: Simon Pettersson <simpet@netinsight.net>
Date:   Tue Feb 7 13:36:29 2023 +0100

    add mutual tls to options struct clone

commit d6df503753d81b31ecabf9d245430442e42d1ca9
Author: Brian Rieger <brieger@vmware.com>
Date:   Wed Mar 23 13:29:01 2022 -0500

    Add in registry mutual tls
```

PR #366, opened 2022-03-30
Set permission when pulling image

### LINKED ISSUE #349: Unpacked images should add back file permissions for group/other.

The `imgpkg push -i` option supports being able to create OCI artefact images containing arbitrary files. Thus it can serve as the equivalent of hosting a tarball on a website, but where things are stored as an image on an image registry.

As much as this appeals as a general solution for packaging up arbitrary files for distribution it suffers a problem.

The problem is that when you run `imgpkg push` the directories/files copied into the image only preserve user permissions and group/other permissions are discarded. As such, when using `imgpkg pull` to download and unpack the image, directories/files get unpacked with only user permission set. For example:

```
0 drwx------  4 graham  wheel  128  8 Mar 12:44 .
0 drwx------  4 graham  wheel  128  8 Mar 12:44 ..
8 -rw-------  1 graham  wheel   50  1 Jan  1970 Dockerfile
8 -rw-------  1 graham  wheel   13  1 Jan  1970 index.html
8 -rwx------  1 graham  wheel  128  1 Jan  1970 script.sh
```

This is always the case, and is nothing to do with a user using a restrictive umask. In other words this happens even if the user umask is 022, which should preserve group/other permissions for `r-x`.

An example of where this is problematic is where the image artefact is used to hold source code files that a user downloads and then needs to run a `docker build` in to create a runnable container image.

The problem in this case is that if files are copied into the container image using `COPY` in the `Dockerfile` they are done so with only user file permissions, but in the container image the files would be owned by root. If the container image is then run as non root it can't access the files copied into the container image.

Best solution may be to continue only storing the file permissions in the image, but when unpacking copy the user permissions to group/other, but where whether those permissions persist will depend on the users umask.

Using this approach, where a change is only made in `imgpkg pull` behaviour, any images created with an older `imgpkg` version would also work the same. So not dependent on what `imgpkg push` did.

So instead of the above result, if the umask had been 022, you would get:

```
0 drwxr-xr-x  4 graham  wheel  128  8 Mar 12:44 .
0 drwxr-xr-x  4 graham  wheel  128  8 Mar 12:44 ..
8 -rw-r--r--  1 graham  wheel   50  1 Jan  1970 Dockerfile
8 -rw-r--r--  1 graham  wheel   13  1 Jan  1970 index.html
8 -rwxr-xr-x  1 graham  wheel  128  1 Jan  1970 script.sh
```

This is how a Git server effectively works when a fresh copy is checked out.

If working in a controlled environment where you know what the umask would be, you can work around the issue if for example you know the umask would be 022, by running:

```
imgpkg pull -i ... -o ...
chmod -R go=u-w ...
```

This is okay if being done in a script, but if the user is directly running the `imgpkg pull` command, it is inconvenient to have to tell them to fix up permissions.

---
Vote on this request

joaopapereira:
Ensures that when an image is pulling all the files will contain the same permission for user, group and other. The permission is based on the user permission that is saved in the image

Fixes #349

### DIFF
--- a/pkg/imgpkg/image/dir_image.go
+++ b/pkg/imgpkg/image/dir_image.go
@@ -124,6 +124,10 @@ func (i *DirImage) extractTarEntry(header *tar.Header, input io.Reader) error {
 	path := filepath.Join(i.dirPath, header.Name)
 	mode := header.FileInfo().Mode()
 
+	// copy user permissions to group and other
+	userPermission := int64(mode & 0700)
+	permMode := os.FileMode(userPermission | userPermission>>3 | userPermission>>6)
+
 	err := os.MkdirAll(filepath.Dir(path), 0700)
 	if err != nil {
 		return err
@@ -131,13 +135,13 @@ func (i *DirImage) extractTarEntry(header *tar.Header, input io.Reader) error {
 
 	switch header.Typeflag {
 	case tar.TypeDir:
-		err := os.MkdirAll(path, mode)
+		err := os.MkdirAll(path, permMode)
 		if err != nil {
 			return err
 		}
 
 	case tar.TypeReg, tar.TypeRegA:
-		file, err := os.Create(path)
+		file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, permMode)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


pkg/imgpkg/image/dir_image.go
```diff
@@ -168,23 +172,21 @@ func (i *DirImage) extractTarEntry(header *tar.Header, input io.Reader) error {
...
-	err = lchmod(header, path, mode)
-	if err != nil {
-		return err
-	}
-
 	// must be done after everything
 	return lchtimes(header, path)
 }
 
 func lchmod(header *tar.Header, path string, mode os.FileMode) error {
```
cppforlife: shouldnt this be removed?

joaopapereira: removed

PR #389, opened 2022-05-11
Don't dirty the working directory with release checksums

benmoss:
This causes the binaries to record the dirty working directory in their version information

Also fixes hack/build.sh which had an undefined variable after https://github.com/vmware-tanzu/carvel-imgpkg/commit/714912cd8644c14f97251d716678de1da45037cc


joaopapereira (approved):
LGTM

PR #408, opened 2022-06-28
Let use know that imgpkg still needs to tag images after copy

### LINKED ISSUE #399: Show what happens between "done" and "Succeeded"
This is a very minor UX problem, but it's funny that imgpkg says "done" and then continues to run for a while before it says "Succeeded". 

Example: 
```
copy |
copy | done uploading images


[... this occupies a surprising amount of time, maybe ~15-30 seconds?]


Succeeded
```

I may be naive but when it says "done" i didn't expect it to still be running afterwards. 

maybe something like the existing "done uploading" line followed by "and now here's what i'm doing next"

This is not important, just capturing it as I was using imgpkg and surprised a little.

---
Vote on this request

joaopapereira:
Fixes #399


joe-kimmel-vmw:
Thank you!

PR #409, opened 2022-06-29
Bump go-containerregistry, remove k8s-pkg-credentialprovider

benmoss:
Move to cloud credential helpers recommended by GGCR:
https://github.com/google/go-containerregistry/tree/main/pkg/authn#emulating-cloud-provider-credential-helpers

We had to fix our test registry to reject mount requests properly with a 202 due to the changes in https://github.com/google/go-containerregistry/pull/1388

Manually validated against GCP and Azure ambient credentials, both of them work and fail gracefully when access to the metadata server is blocked.


pkg/imgpkg/registry/auth/env_keychain.go
```diff
@@ -54,7 +55,7 @@ func (k *EnvKeychain) Resolve(target regauthn.Resource) (regauthn.Authenticator,
 	}
 
 	for _, info := range infos {
-		registryURLMatches, err := credentialprovider.URLsMatchStr(info.URL, target.String())
+		registryURLMatches, err := urlsMatchStr(info.URL, target.String())
```
benmoss: This seems a little unfortunate but is what GGCR did as well https://github.com/google/go-containerregistry/blob/2b1087abc076a6e82339480ea191b7245962932c/pkg/authn/kubernetes/keychain.go#L187

pkg/imgpkg/registry/keychain_test.go
```diff
@@ -1,760 +0,0 @@
-// Copyright 2021 VMware, Inc.
```
benmoss, line 1: This test is basically just testing the libraries and the ordering of the keychain. It'd be nice to have some automated tests around the libraries behavior but this is the wrong level to test at. Testing the ordering of the keychain doesn't seem valuable.

pkg/imgpkg/registry/auth/env_keychain.go
joaopapereira: according to the comment that is how kubernetes auth keychain work as well

benmoss: Oh yeah, I omitted the full context there accidentally. Kubernetes [has the original version of this code](https://github.com/kubernetes/kubernetes/blob/b2ed6ca64f55452cdb4063d62920dddccd4344f7/pkg/credentialprovider/keyring.go#L184-L194), it was extracted out into vdemeester/k8s-pkg-credentialprovider awaiting https://github.com/kubernetes/enhancements/pull/1406, I'm not exactly sure where that ended up but now GGCR has copied that code into their codebase for their Kubernetes keychain implementation

benmoss: We might be able to simplify this code, I'm honestly not sure why it takes this much code to compare a URL and a registry string, but I just did the lazy option of copying all these interconnected functions

pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -40,29 +39,6 @@ func (r *RegistryFlags) Set(cmd *cobra.Command) {
 
 	cmd.Flags().DurationVar(&r.ResponseHeaderTimeout, "registry-response-header-timeout", 30*time.Second, "Maximum time to allow a request to wait for a server's response headers from the registry (ms|s|m|h)")
 	cmd.Flags().IntVar(&r.RetryCount, "registry-retry-count", 5, "Set the number of times imgpkg retries to send requests to the registry in case of an error")
-
-	cmd.Flags().String("registry-azure-cr-config", "", "Path to the file containing Azure container registry configuration information. ($IMGPKG_REGISTRY_AZURE_CR_CONFIG)")
```
joaopapereira, line 44: In the docs we should make sure that we remove the existence of this flag as well

joaopapereira (approved):
LGTM

pkg/imgpkg/cmd/registry_flags.go
benmoss: Need to figure out the implications for the other auth providers as well

cppforlife: is that somehow magically derived in ggcr or we just no longer support azure configuration?

pkg/imgpkg/registry/keychain.go
```diff
@@ -15,10 +20,14 @@ import (
 // keychains that contain credentials for 'any' target. i.e. env keychain takes precedence over the custom keychain.
 // Since env keychain contains credentials per HOSTNAME, and custom keychain doesn't.
 func Keychain(keychainOpts auth.KeychainOpts, environFunc func() []string) (regauthn.Keychain, error) {
-	iaasKeychain, err := auth.NewIaasKeychain(context.Background(), environFunc)
-	if err != nil {
-		return nil, err
+	keychain := []authn.Keychain{
+		auth.CustomRegistryKeychain{Opts: keychainOpts},
```
cppforlife: looks like CustomRegistryKeychain was moved up to the front. before it was last, was this intentional?

pkg/imgpkg/registry/auth/env_keychain.go
cppforlife: mind dropping them into a separate package in this file and adding a comment where they were copied from.

pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -70,6 +70,9 @@ func (r *RegistryFlags) AsRegistryOpts() registry.Opts {
 	if os.Getenv("IMGPKG_ANON") == "true" {
 		opts.Anon = true
 	}
+	if os.Getenv("IMGPKG_ENABLE_IAAS_AUTH") != "false" {
+		opts.EnableIaasAuthProviders = true
```
joaopapereira, line 74: Do we really still need this feature?

benmoss: @cppforlife thinks so, apparently users run into problems and want to be able to fallback to ignoring ambient credentials sometimes? I think especially since they are higher priority than the Docker config now, another piece of feedback

pkg/imgpkg/cmd/registry_flags.go
benmoss: The ACR library we're switching to (https://github.com/chrismellard/docker-credential-acr-env/) supports pulling authentication from the environment via `AZURE_CLIENT_ID`/`AZURE_CLIENT_SECRET`/`AZURE_TENANT_ID`. It's what's being used by [ko](https://github.com/google/ko/blob/691da5afb64e1d9204b250abc41fc738737b1ebb/pkg/commands/config.go#L47) and [cosign](https://github.com/sigstore/cosign/blob/516cc39f5582f46af9fc5b9081af08dba24441e7/cmd/cosign/cli/options/registry.go#L82-L90) so it seems like an acceptable choice

benmoss:
Maybe we should consider only using the Docker/env keychains by default, and enabling workload identity/ambient credentials via a flag

joaopapereira:
The main change that we are currently missing is:
- By default do not include all the keychains
- Create a new flag `--activate-keychain="gke"` with possible values [`gke`, `azure`, `aws`, `github`]
- rebase on top of develop branch

To keep backward compatibility `kapp-controller` would have to enable the IAAS keychains that are needed for the current installation.

joaopapereira:
going to close this PR since we merged #431 with the same commits. Thanks for the good work here @benmoss

Issue #216, opened 2021-08-11
Include more info in warning about non-distributable layer


gabegorelick:
**Describe the problem/challenge you have**

`imgpkg copy` will often log the following warning:

> copy | Warning: Skipped layer due to it being non-distributable. If you would like to include non-distributable layers, use the --include-non-distributable-layers flag

But this doesn't say anything about what the non-distributable layer is. Without more information, it's hard to determine whether I should pass `--include-non-distributable-layers` or if my bundle is messed up. And with a lot of images and a lot of layers, it can be hard to figure out which one is the offender.

**Describe the solution you'd like**
Include info about which image and which layer is non-distributable in the log message.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

DennisDenuto:
Thanks for the issue.

I think this is a nice improvement to the warning message.

Instead of both the image and layer/s, what do you think about specifying *only* the image? Since, with that information you can inspect the image manifest (giving you a list of the layers, of which you can see which are non-distributable)

Note:
Consider how to output a bundle with *lots* of images that have Non-distributable layers. (This warning might get really verbose)

gabegorelick:
> Instead of both the image and layer/s, what do you think about specifying only the image?

That may be OK. It's certainly a decent place to start.

joaopapereira:
This issue was solved in #410 and released in version 0.30.0

PR #410, opened 2022-07-01
Add image and layer information to warning of non distributable layers

joaopapereira:
Fixes issue #216


pkg/imgpkg/cmd/copy.go
```diff
@@ -309,76 +310,150 @@ func (c *CopyOptions) writeBundleLockOutput(bundle *bundle.Bundle) error {
 	return bundleLock.WriteToPath(c.LockOutputFlags.LockFilePath)
 }
 
-func processedImagesMediaType(processedImages *ctlimgset.ProcessedImages) []string {
-	everyMediaType := []string{}
+type nonDistributableLayers struct {
+	imgRef     string
+	imgFullRef string
+	layers     []string
```
cppforlife: nit: these fields should be "public" since they are accessed by bunch of places (and and not methods private to this struct). struct itself makes sense to keep private since it's private to the package.

pkg/imgpkg/cmd/copy.go
```diff
@@ -309,76 +310,150 @@ func (c *CopyOptions) writeBundleLockOutput(bundle *bundle.Bundle) error {
...
+}
+
+func getNonDistributableLayersFromImageDescriptors(ids *imagedesc.ImageRefDescriptors) []nonDistributableLayers {
```
cppforlife: it would be great to move these functions into its own file (since they are self-contained and mostly misc) so that they do not "pollute" main copy file.

cppforlife (commented):
2 minor nits, otherwise lgtm

pkg/imgpkg/cmd/copy.go
joaopapereira: done

pkg/imgpkg/cmd/copy.go
joaopapereira: done

Issue #268, opened 2021-10-06
[windows] imgpkg does not have a deterministic push across OS


DennisDenuto:
**What steps did you take:**

On Windows
```
> git clone https://github.com/vmware-tanzu/carvel-imgpkg
> cd carvel-imgpkg
> imgpkg push -i image-windows -f test/e2e/assets/simple-app --yes
Succeded pushing sha sha256:73e65af5e889da98ece0bd471b2cd83baac574e19943108212ba0417ff8f03c2
```

On Linux
```
> git clone https://github.com/vmware-tanzu/carvel-imgpkg
> cd carvel-imgpkg
> imgpkg push -i image-linux -f test/e2e/assets/simple-app --yes
Succeded pushing sha sha256:ceef30cbdce418efde0284f446df9cec9e535adcd6e1010dad30ddae1dc9367b

> imgpkg pull -i image-linux -o /tmp/linux
> imgpkg pull -i image-windows -o /tmp/windows
```

```
 ls -l /tmp/windows
total 48
-rw-------  1 dennisleon  wheel  11558 Dec 31  1969 LICENSE
-rw-------  1 dennisleon  wheel     98 Dec 31  1969 README.md
drwx------  2 dennisleon  wheel     64 Dec 31  1969 config
-rw-------  1 dennisleon  wheel     25 Dec 31  1969 config\config.yml
drwx------  2 dennisleon  wheel     64 Dec 31  1969 config\inner-dir
-rw-------  1 dennisleon  wheel      6 Dec 31  1969 config\inner-dir\README.txt

ls -l /tmp/linux
total 32
-rw-------  1 dennisleon  wheel  11357 Dec 31  1969 LICENSE
-rw-------  1 dennisleon  wheel     95 Dec 31  1969 README.md
drwx------  4 dennisleon  wheel    128 Oct  6 11:25 config
```

**What happened:**
Image pushed resulted in a different sha (and different contents) across windows and linux

**What did you expect:**
The same image pushed. Or a way to differentiate between an image built for windows vs linux. Using an Image Index perhaps?

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.19.0
- Docker registry used (e.g. `Docker HUB`): 
- OS (e.g. from `/etc/os-release`): Windows 10

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
[...]

PR #415, opened 2022-07-15
Ensure that pushed images can be pulled in different OS

### LINKED ISSUE #271: [windows] imgpkg pushing a bundle on windows cannot be pulled on linux
Related to: https://github.com/vmware-tanzu/carvel-imgpkg/issues/268

**What steps did you take:**

On windows
```
> imgpkg.exe push -f test\e2e\assets\bundle_file_permissions -b image
> imgpkg pull -b image -o ./perms
Succeeds
```

On Linux
```
> imgpkg pull -b image -o ./perms
imgpkg: Error: Reading path /tmp/perms/.imgpkg/images.yml: open /tmp/perms/.imgpkg/images.yml: no such file or directory
```

**What happened:**
linux is unable to find `images.yml` I think due to windows imgpkg using backslashes and linux expecting forward slashes.

**What did you expect:**
imgpkg pull to work on linux

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.19.0
- OS (e.g. from `/etc/os-release`): Windows 10

---
Vote on this request

joaopapereira:
Fixes #271


pkg/imgpkg/image/tar_image.go
```diff
@@ -128,9 +137,26 @@ func (i *TarImage) addFileToTar(fullPath, relPath string, info os.FileInfo, tarW
...
+
+		// Due to the difference of line endings on windows this next block will do a dos2unix of the file
+		// before writing it to the tar
+		b, err := ioutil.ReadAll(file)
+		if err != nil {
+			return err
+		}
+		b = regexp.MustCompile(`\r\n`).ReplaceAllLiteral(b, []byte{'\n'})
+		size = int64(len(b))
+		r = bytes.NewBuffer(b)
```
cppforlife: hmm, this doesnt feel right to me for 2 reasons:
- we are reading in entire file contents into memory
- we are modifying contents without explicit user ask (e.g. this will destroy binary files).

cppforlife:
im assuming https://github.com/vmware-tanzu/carvel-imgpkg/issues/271 is related to paths which i think is reasonable to normalize. 

regarding https://github.com/vmware-tanzu/carvel-imgpkg/issues/268, i think it's important for us not to mangle files that user gives us (unless explicitly requested; not sure though that's really job of imgpkg to "fix encodings" since we are content agnostic). whatever was put into a bundle, should come out with exactly same shas on the other end.

joaopapereira:
If we look at #268 right now imgpkg is creating images with different digest for the same content in windows and UNIX systems. This is due to the carriage returns.
An example of that would be if you do a git clone of a repository in UNIX and also in windows and you create an image from it, it should have the same sha but right now it does not.
So in some sense, I believe that imgpkg to keep the guarantee that for the same content we have the same sha we need to do the encoding fixing.

The binary format indeed is a problem. I will look into possibilities to ensure that we can check if a file is binary or not in a simple and less costly way.

cppforlife:
> If we look at https://github.com/vmware-tanzu/carvel-imgpkg/issues/268 right now imgpkg is creating images with different digest for the same content in windows and UNIX systems. This is due to the carriage returns.

it's not the same content though. if you shasum the content it will be different on unix vs windows, right? given that i wouldnt expect that imgpkg would not produce the same digest.

> An example of that would be if you do a git clone of a repository in UNIX and also in windows and you create an image from it, it should have the same sha but right now it does not.

git clone works on text, unlike imgpkg which works on files with arbitrary content. we have no idea what's inside these files and whether it's ok or not ok to modify them.

> The binary format indeed is a problem. I will look into possibilities to ensure that we can check if a file is binary or not in a simple and less costly way.

i think best that can be done with detection is guessing on X number of bytes, but that's pretty hacky.

question related to this topic: what does tar do? since imgpkg is pretty similar to tar in spirit.

joaopapereira:
> question related to this topic: what does tar do? since imgpkg is pretty similar to tar in spirit.

looked into tar and it does nothing, so I will revert my changes

PR #418, opened 2022-07-21
Add feature to allow resume of copy to tar

joaopapereira:


pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -1273,6 +1347,7 @@ func doesLayerExistInTarball(t *testing.T, path string, digest regv1.Hash) bool
 		}
 		require.NoError(t, err)
 		if hdr.Name == filePathInTar {
+
```
cppforlife: nit: extra newline

pkg/imgpkg/imagetar/tar_writer.go
```diff
@@ -167,14 +168,29 @@ func (w *TarWriter) writeLayers() error {
 		if isInflatable {
 			stream = nil
 		} else {
-			foundLayer, err := w.ids.FindLayer(imgLayer)
-			if err != nil {
-				return err
+			for _, layer := range w.layersOtherSource {
+				d, err := layer.Digest()
+				if err != nil {
+					return fmt.Errorf("retrieving digest: %s", err)
```
cppforlife: nit: carvel error messages start with capitalized letter

pkg/imgpkg/imagetar/tar_writer.go
```diff
@@ -37,11 +37,12 @@ type TarWriter struct {
 	opts                  TarWriterOpts
 	ui                    goui.UI
 	imageLayerWriterCheck ImageLayerWriterFilter
+	layersOtherSource     []regv1.Layer
```
cppforlife: nit: may be better, layersOtherSource => layersFromOtherSource

pkg/imgpkg/imagetar/tar_reader.go
```diff
@@ -28,6 +31,98 @@ func (r TarReader) Read() ([]imagedesc.ImageOrIndex, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		closer, err := verify.ReadCloser(r, size, h)
+		if err != nil {
+			return nil, err
+		}
+
+		output := make([]byte, size)
+		n, err := closer.Read(output)
```
cppforlife: not clear why we need to make a byte array of the size of the layer? how can we evaluate validity of a layer in a streaming manner.

pkg/imgpkg/imagetar/tar_reader.go
```diff
@@ -28,6 +31,98 @@ func (r TarReader) Read() ([]imagedesc.ImageOrIndex, error) {
...
+}
+
+func (r TarReader) getPresentLayersForIndex(idx v1.ImageIndex) ([]v1.Layer, error) {
+	var result []v1.Layer
+	dIdx, correct := idx.(*imagedesc.DescribedImageIndex)
+	if !correct {
+		panic("Internal inconsistency: unexpected index type")
+	}
+	for _, image := range dIdx.Images() {
+		missingLayers, err := r.getPresentLayersForImage(image)
```
cppforlife: are they missing layers or actually present layers?

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,64 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
+		if err != nil {
+			return nil, fmt.Errorf("Creating tmp folder: %s", err)
+		}
+		defer os.Remove(tmpFile.Name())
+
+		outputFile, err = os.Open(outputPath)
+		if err != nil {
+			return nil, fmt.Errorf("Opening file '%s': %s", outputPath, err)
+		}
+		io.Copy(tmpFile, outputFile)
```
cppforlife: io.Copy returns an error

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,64 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
+func (i TarImageSet) Export(foundImages *UnprocessedImageRefs, outputPath string, registry registry.ImagesReaderWriter, imageLayerWriterCheck imagetar.ImageLayerWriterFilter, resume bool) (d *imagedesc.ImageRefDescriptors, err error) {
 	ids, err := i.imageSet.Export(foundImages, registry)
 	if err != nil {
 		return nil, err
 	}
 
-	outputFile, err := os.Create(outputPath)
+	var outputFile *os.File
+	var layers []v1.Layer
+	var tmpFile *os.File
```
cppforlife, line 44: it's not clear by the name how tmpFile is used (this deserves a comment). it seems like we do quite a bit of juggling with it.

pkg/imgpkg/cmd/copy_repo_src_test.go
cppforlife: nit: extra new line

pkg/imgpkg/cmd/copy_repo_src_test.go
```diff
@@ -273,10 +276,81 @@ func TestToTarImage(t *testing.T) {
...
+					// Found configuration layer, it doesn't count as content blob
+					numberOfRequests--
+				}
+			}
+
+			return false
+		})
+
+		imageTarPath := filepath.Join(os.TempDir(), "img.tar")
+		defer os.Remove(imageTarPath)
```
cppforlife: if we are hard coding temp location, then we should remove it before it use. additionally lets namespace the name e.g. imgpkg-test-img.tar

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,64 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
+		tmpFile, err = ioutil.TempFile("", "tmp-extract")
+		if err != nil {
+			return nil, fmt.Errorf("Creating tmp folder: %s", err)
+		}
+		defer os.Remove(tmpFile.Name())
+
+		outputFile, err = os.Open(outputPath)
+		if err != nil {
+			return nil, fmt.Errorf("Opening file '%s': %s", outputPath, err)
+		}
```
cppforlife: it seems that first invocation of imgpkg would not be able to contain --resume, since output path might not exist. since this functionality would typically be used in a loop to retry i would recommend to allow absense of a file be allowed when --resume is specified.

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,64 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
+			if err1 != nil {
+				err = fmt.Errorf("original error: %s, post exit error: %s", err, err1)
+				return
+			}
+			lTmpFile, err1 := os.Open(tmpFile.Name())
+			if err1 != nil {
+				err = fmt.Errorf("original error: %s, post exit error: %s", err, err1)
+				return
+			}
+			io.Copy(outputFile, lTmpFile)
```
cppforlife: missing error check

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,64 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
+		}
```
cppforlife, line 112: both outputFile and lTmpFile are leaked.

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,64 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
 	if err != nil {
 		return nil, err
 	}
 
-	outputFile, err := os.Create(outputPath)
+	var outputFile *os.File
+	var layers []v1.Layer
+	var tmpFile *os.File
+	if resume {
+		tmpFile, err = ioutil.TempFile("", "tmp-extract")
```
cppforlife: lets include some identifiable information e.g. imgpkg-tar-imageset-...

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,64 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
-func (i TarImageSet) Export(foundImages *UnprocessedImageRefs, outputPath string, registry registry.ImagesReaderWriter, imageLayerWriterCheck imagetar.ImageLayerWriterFilter) (*imagedesc.ImageRefDescriptors, error) {
+func (i TarImageSet) Export(foundImages *UnprocessedImageRefs, outputPath string, registry registry.ImagesReaderWriter, imageLayerWriterCheck imagetar.ImageLayerWriterFilter, resume bool) (d *imagedesc.ImageRefDescriptors, err error) {
 	ids, err := i.imageSet.Export(foundImages, registry)
 	if err != nil {
 		return nil, err
 	}
 
-	outputFile, err := os.Create(outputPath)
+	var outputFile *os.File
+	var layers []v1.Layer
```
cppforlife: clearer name: layers => alreadyDownloadedLayers

pkg/imgpkg/imagetar/tar_reader.go
```diff
@@ -28,6 +31,98 @@ func (r TarReader) Read() ([]imagedesc.ImageOrIndex, error) {
 	return imagedesc.NewDescribedReader(ids, file).Read(), nil
 }
 
+// GetLayersPresentInFile retrieves all the layers that are present in a tar file
+func (r TarReader) GetLayersPresentInFile() ([]v1.Layer, error) {
```
cppforlife: suggestion on naming: GetLayersPresentInFile() => PresentLayers(). InFile is redundant since it's already on TarReader.

pkg/imgpkg/imagetar/tar_reader.go
```diff
@@ -28,6 +31,98 @@ func (r TarReader) Read() ([]imagedesc.ImageOrIndex, error) {
...
+			return nil, err
+		}
+		closer, err := verify.ReadCloser(r, size, h)
+		if err != nil {
+			return nil, err
+		}
+
+		output := make([]byte, size)
+		n, err := closer.Read(output)
+		if err != nil {
```
cppforlife, line 90: if err is not nil, then i dont think we should be considering this as a present layer.

pkg/imgpkg/imagetar/tar_reader.go
joaopapereira: In the feature's initial implementation, I searched for the missing layers. I forgot to rename these variables 😄

pkg/imgpkg/imagetar/tar_reader.go
joaopapereira: Changed the approach it doesn't buffer in imgpkg the read bytes

pkg/imgpkg/imageset/tar_image_set.go
joaopapereira: sounds good, updated the PR to continue like if --resume was not provided

pkg/imgpkg/imageset/tar_image_set.go
```diff
@@ -27,21 +29,82 @@ func NewTarImageSet(imageSet ImageSet, concurrency int, ui goui.UI) TarImageSet
...
+		// This will just follow the normal path of resume == false
+		outputFile, err = os.Open(outputPath)
+		if err == nil {
+			tmpFile, err = ioutil.TempFile("", "imgpkg-tar-imageset-")
+			if err != nil {
+				return nil, fmt.Errorf("Creating tmp folder: %s", err)
+			}
+			defer os.Remove(tmpFile.Name())
+
+			io.Copy(tmpFile, outputFile)
```
cppforlife: need to check io.Copy error (are there any other places?)

pkg/imgpkg/imagetar/tar_reader.go
```diff
@@ -28,6 +32,93 @@ func (r TarReader) Read() ([]imagedesc.ImageOrIndex, error) {
...
+			return nil, err
+		}
+		closer, err := verify.ReadCloser(r, size, h)
+		if err != nil {
+			return nil, err
+		}
+
+		n, err := io.Copy(io.Discard, closer)
+		if err != nil || n == 0 {
```
cppforlife: under what conditions `n == 0` happens vs err!=nil?

cppforlife: (not sure which case we are trying to address with n==0)

joaopapereira: Not sure if this was the result of a prior error I was getting when calling the reader or not. So I just kept it. But there were cases where it would not fail and read 0 bytes, nevertheless it might not be the case anymore since I fixed something else. I will try to remove it to see if it works without it.

joaopapereira: I just run the tests locally and it look like that we can remove the `n==0` portion

cppforlife:
lgtm.

PR #424, opened 2022-08-11
Add pull API

### LINKED ISSUE #390: When pulling image or bundle inform the caller if the content is cacheable or not
When another tool like vendir tries to pull a bundle or image using imgpkg, it should be informed if the contents are cacheable or not. This solution will allow tools to not rely as much on the registry and can store some of the image information on the disk.

I am proposing 2 possible UX implementations for this feature.

**Option 1:** Create a machine-readable output for the pull command
In this option add `--output-type` that can have the value `yaml` and would look something like the following example:
```yaml
image: localhost:5000/test@sha256:a9a409d96a51b16e3cbb214ad192004e57cf5c7c18ea66b9860c83395048e5eb
imageslock:
    path: /tmp/.imgpkg/images.yml
    updated: true
cacheable: true
```
or for recursive pulls
```yaml
image: localhost:5000/test@sha256:a9a409d96a51b16e3cbb214ad192004e57cf5c7c18ea66b9860c83395048e5eb
imageslock:
  path: /tmp/.imgpkg/images.yml
  updated: true
cacheable: true
inner-bundles:
- image: localhost:5000/test@sha256:a9a409d96a51b16e3cbb214ad192004e57cf5c7c18ea66b9860c83395048e555
  path: /tmp/.imgpkg/bundles/sha256-a9a409d96a51b16e3cbb214ad192004e57cf5c7c18ea66b9860c83395048e555
  imageslock:
    path: /tmp/.imgpkg/bundles/sha256-a9a409d96a51b16e3cbb214ad192004e57cf5c7c18ea66b9860c83395048e555imgpkg/images.yml
    updated: true
```

**Pros:**
- Provide a machine-readable output by calling the CLI
- YAML output is standard and it is easy to parse
**Cons:**
- Force other tools that call imgpkg as a CLI to know the version used (cannot use the flag on an older version of imgpkg)
- Other tools would have to call the CLI and parse the output

---
**Option 2:** Dedicated API to provide this information
In this option, the tools can include imgpkg as a library and that would allow them to have more control over processing the outputs.

One possible API could be:
```go
type BundleInfo struct {
  ImageRef string
  ImagesLockPath string
  Path string
  NestedBundles []BundleInfo
}

type Status struct {
  Cacheable bool
  ImageRef string
  IsBundle bool
  NestedBundles []BundleInfo
}

// Pull Download the contents of the image referenced by imageRef to the folder outputPath
func Pull(ctx context.Context, imageRef string, outputPath string, registryOpts registry.Opts) (Status, error)

// PullRecursive Downloads the contents of the Bundle referenced by imageRef to the folder outputPath.
// This functions should error out when imageRef does not point to a Bundle
func PullRecursive(ctx context.Context, imageRef string, outputPath string, registryOpts registry.Opts) (Status, error)
```

**Pros:**
- No need to do output processing
- More control over action to take when the API call ends
**Cons:**
- Other tools would have to include imgpkg as a library

**Note:** Adding some text to the current output was also considered. Parsing text is a process that can be error-prone and changing versions of the text can cause other tools that expect a particular text to fail. Due to this we advise against this implementation and didn't provide it as an option.

---

Looking at the pros and cons I believe the best solution would be to create an API that can be used by other tools. This will solve compatibility issues between these tools and imgpkg and would allow these tools to have better control over the errors that might surface from the pull.

From the following scenarios, let's check which ones imgpkg knows are not going to change and can be cacheable
- **Pull image using a tag**
    In this scenario, the contents are **not cacheable** because the tags can be moved to a different image
- **Pull image using the SHA**
    In this scenario, the contents are **cacheable**
- **Pull bundle using a tag**
    In this scenario, the contents are **not cacheable** because the tags can be moved to a different bundle
- **Pull bundle using the SHA and the bundle is fully relocated**
    In this scenario, the contents are **cacheable** because nothing will change on the contents of the bundle
- **Pull bundle using the SHA and the bundle is NOT fully relocated**
    In this scenario, the contents are **not cacheable** because is in the future the bundle becomes fully copied the ImagesLock file can be updated with the new location of the images
- **Pull bundle using the SHA and the bundle is partially copied**
    In this scenario, the contents are **not cacheable** because is in the future the remaining images might be copied and that will update the ImagesLock file

TLDR; currently imgpkg cannot partially copy a bundle.

---
Vote on this request

joaopapereira:
Introduce the new API for pull command
Enables the output as YAML for the pull command to ease the integration with other tools

Introduces the notion of a pulled image/bundle being cacheable

Fixes #390


pkg/imgpkg/bundle/bundle.go
```diff
@@ -31,6 +31,7 @@ type Logger interface {
 	Debugf(msg string, args ...interface{})
 	Tracef(msg string, args ...interface{})
 	Logf(msg string, args ...interface{})
+	UI() goui.UI
```
cppforlife: hmm, seems a bit weird for logger to carry ui. seems like we are spilling out internal implementation.

pkg/imgpkg/bundle/bundle.go
```diff
@@ -45,11 +46,23 @@ type ImagesMetadata interface {
 	FirstImageExists(digests []string) (string, error)
 }
 
+// GraphNode Node information of a Bundle
+type GraphNode struct {
```
cppforlife, line 48: GraphNode does not really indicate what it is. is it some type of content within a bundle? what would be a better name for this?

pkg/imgpkg/cmd/pull.go
```diff
@@ -49,6 +58,7 @@ func NewPullCmd(o *PullOptions) *cobra.Command {
 	o.BundleRecursiveFlags.Set(cmd)
 	o.LockInputFlags.Set(cmd)
 	cmd.Flags().StringVarP(&o.OutputPath, "output", "o", "", "Output directory path")
+	cmd.Flags().StringVarP(&o.OutputType, "output-type", "", "", "Will print machine readable output on standard out: [yaml]")
```
cppforlife: since we have a flag called --output that takes a directory path, it's pretty confusing to have a --output-type flag that controls unrelated functionality.

pkg/imgpkg/v1/pull.go
```diff
@@ -0,0 +1,207 @@
...
+type ImagesLockInfo struct {
+	Path    string `json:"path"`
+	Updated bool   `json:"updated"`
+}
+
+// BundleInfo Information related to the specific bundle
+type BundleInfo struct {
+	ImageRef      string          `json:"image"`
+	ImagesLock    *ImagesLockInfo `json:"imagesLock,omitempty"`
+	NestedBundles []BundleInfo    `json:"inner-bundles,omitempty"`
```
cppforlife: inconsistent casing of json names

pkg/imgpkg/v1/pull.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// InfoFromNestedBundle Given a Bundle struct it can generate the BundleInfo associated to the nested bundles
+func InfoFromNestedBundle(b *bundle.Bundle, updated bool) []BundleInfo {
```
cppforlife: it's not clear from the function signature what is `updated`

pkg/imgpkg/v1/pull.go
```diff
@@ -0,0 +1,207 @@
...
+	case isBundle && pullOptions.AsImage: // Trying to pull the OCI Image of a Bundle
+		st, err := pullImage(imageRef, outputPath, pullOptions, reg)
+		if err != nil {
+			return Status{}, err
+		}
+		st.IsBundle = true
+		return st, nil
+
+	case isBundle && pullOptions.IsBundle: // Trying to pull a Bundle
+		return pullBundle(imageRef, outputPath, pullOptions, reg, false)
```
cppforlife: it seems like we are now making an extra registry call to check if something is a bundle... bundle.NewBundle().IsBundle and we build another copy of the bundle inside pullBundle.

pkg/imgpkg/v1/pull.go
cppforlife: do we need to make this function public?

pkg/imgpkg/v1/pull.go
```diff
@@ -0,0 +1,207 @@
...
+	var nestedBundles []BundleInfo
+
+	for _, nBundle := range b.NestedBundles() {
+		nestedBundles = append(nestedBundles, NewBundleInfo(nBundle, updated))
+	}
+	return nestedBundles
+}
+
+// NewBundleInfo create a BundleInfo struct from bundle information
+func NewBundleInfo(node bundle.GraphNode, updated bool) BundleInfo {
```
cppforlife: do we need to make this function public?

pkg/imgpkg/cmd/pull.go
joaopapereira: I thought about that but we already were using output-type for the same thing in describe command. 
It would be strange to have the same behavior behind 2 different flags.

What is your suggestion?

pkg/imgpkg/bundle/bundle.go
joaopapereira: I had called it PullGraphNode before this commit but it felt too specific and we might use this structure in the future to keep some resemblance to the structure of the bundle. By structure, I am saying the bundles associated with the current bundle.

joaopapereira: Since it was in the `bundle` package I assumed that `bundle.GraphNode` would have enough information. What do you think?

pkg/imgpkg/cmd/pull.go
cppforlife: im not sure tbh. asking others in slack.

pivotaljohn: See also: [#carvel ~ 2022-08-16](https://kubernetes.slack.com/archives/CH8KCCKA5/p1660581869764349?thread_ts=1660581817.485209&cid=CH8KCCKA5)

cppforlife:
lgtm

PR #431, opened 2022-08-31
Bump go-containerregistry, remove k8s-pkg-credentialprovider

joaopapereira:
Move to cloud credential helpers recommended by GGCR:
https://github.com/google/go-containerregistry/tree/main/pkg/authn#emulating-cloud-provider-credential-helpers

We had to fix our test registry to reject mount requests properly with a 202 due to the changes in https://github.com/google/go-containerregistry/pull/1388

Manually validated against GCP and Azure ambient credentials, both of them work and fail gracefully when access to the metadata server is blocked.
Based on the PR #409

---

Apart from the removal of the dependency on k8s-pkg-credentialprovider this PR also changes the default behavior of imgpkg. The keychains no longer get loaded by default, in order to load a particular keychain the environment variable `IMGPKG_ACTIVE_KEYCHAINS` needs to be set. The Keychains that are supported are:
- `gke` Enabled access to the GKE authentication helpers
- `aks`  Enabled access to the AKS authentication helpers
- `ecr`  Enabled access to the ECR authentication helpers
- `github`  Enabled access to the Github authentication helpers

To enable multiple keychains you can provide them in a comma separated list like
`export IMGPKG_ACTIVE_KEYCHAINS=gke,aks,ecr,github`

This PR also starts the deprecation process for the environment variable `IMGPKG_ENABLE_IAAS_AUTH`. For now, if set to true, it is similar to doing `export IMGPKG_ACTIVE_KEYCHAINS=gke,aks,ecr,github`. In the future, this flag will no longer be used.


pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -94,6 +73,25 @@ func (r *RegistryFlags) AsRegistryOpts() registry.Opts {
 	if os.Getenv("IMGPKG_ANON") == "true" {
 		opts.Anon = true
 	}
+	iaasAuth, found := os.LookupEnv("IMGPKG_ENABLE_IAAS_AUTH")
+	if found {
+		if iaasAuth == "true" {
```
neil-hickey: do we wanna remove the casing before the check? Like strings.ToLower() or something. So that even if I put `True` it works? Or is that going against the current way we check the env_vars?

pkg/imgpkg/cmd/registry_flags.go
```diff
@@ -94,6 +73,25 @@ func (r *RegistryFlags) AsRegistryOpts() registry.Opts {
...
+			opts.EnableIaasAuthProviders = true
+		}
+	}
+
+	keychains, found := os.LookupEnv("IMGPKG_ACTIVE_KEYCHAINS")
+	if found {
+		if len(keychains) > 0 {
```
neil-hickey, line 83: nit: Any opposition to using `&&` to reduce the amount of branching going on?

pkg/imgpkg/registry/keychain.go
```diff
@@ -15,10 +20,39 @@ import (
...
+		keychain = append(keychain,
+			google.Keychain,
+			regauthn.NewKeychainFromHelper(ecr.NewECRHelper(ecr.WithLogger(io.Discard))),
+			regauthn.NewKeychainFromHelper(credhelper.NewACRCredentialsHelper()),
+			github.Keychain,
+		)
+	} else {
+		for _, activeKeychain := range keychainOpts.ActiveKeychains {
+			var k regauthn.Keychain
+			fmt.Println(activeKeychain)
```
neil-hickey: remove println?

test/helpers/registry/blobs.go
```diff
@@ -144,10 +143,14 @@ func (m *memHandler) Put(_ context.Context, _ string, h v1.Hash, rc io.ReadClose
 }
 
 // Mount is a no-op since all the blobs are store indexed by sha
-func (m *memHandler) Mount(_ context.Context, _, _ string, _ v1.Hash) error {
+func (m *memHandler) Mount(_ context.Context, repo, from string, h v1.Hash) error {
+	// Mount(ctx context.Context, repo, from string, h v1.Hash) error
```
neil-hickey: is this comment supposed to be here?

test/helpers/registry/blobs.go
```diff
@@ -347,12 +350,12 @@ func (b *blobs) handle(resp http.ResponseWriter, req *http.Request) *regError {
...
-				return regErrInternal(err)
+			if err == nil {
+				resp.Header().Set("Docker-Content-Digest", h.String())
+				resp.WriteHeader(http.StatusCreated)
+				return nil
 			}
-			resp.Header().Set("Docker-Content-Digest", h.String())
-			resp.WriteHeader(http.StatusCreated)
-			return nil
+			// check err type??
```
neil-hickey, line 357: a TODO? Or just thought? Looks like we are ignoring any error that the Mount() comes back with now.

neil-hickey (approved):
LGTM. Some minor nits :)

pkg/imgpkg/cmd/registry_flags.go
joaopapereira: We were using != "false" so I think you are right, doing ToLower makes sense. Fixed

pkg/imgpkg/cmd/registry_flags.go
joaopapereira: No reason, changed to &&

test/helpers/registry/blobs.go
joaopapereira: no, removed

test/helpers/registry/blobs.go
joaopapereira: This is just a temporary pkg that we are waiting it to be merged in ggcr so we can remove it from imgpkg

PR #438, opened 2022-09-19
Bump go to 1.19.1

joaopapereira:
Bump other libraries that have CVE associated with them (these CVE's do not impact imgpkg)


joaopapereira:
The external tests will fail because github does not test new versions of github workflows when pushing code

PR #455, opened 2022-11-30
Fixing CVE's

rohitagg2020:
Fixing CVE's


joaopapereira (approved):
LGTM
The workflows that are failing will keep on failing because the event `pull_request_target` does not allow new workflows to run against the repository.

PR #457, opened 2022-12-02
V0.31.1 bump deps

joaopapereira:


neil-hickey (approved):
LGTM. Ran through trivy and osspi tools, both say it's clear of CVEs 🥳

PR #484, opened 2023-02-13
Add new constructor to the registry

### LINKED ISSUE #481: imgpkg Registry API does not honor secret provided via keychain
**What steps did you take:**

I have upgraded `imgpkg` to the latest version (v0.35.0) in my application. After the upgrade unit test for pulling images from an authenticated registry is failing with HTTP status 401. 

**What happened:**

HTTP GET request by `imgpkg` is missing `Authorization` from the request header when registry auth data is provided via keychain.

**What did you expect:**

I expect `imgpkg` registry API to use the user-provided keychain for authentication.

I believe the bug is in this line of code where `imgpkg` is using its default keychain for authentication. https://github.com/carvel-dev/imgpkg/blob/682c97c7d1f7327434fc6cb525839ed8b9575eff/pkg/imgpkg/registry/registry.go#L265 
**Environment:**

- imgpkg version (use `imgpkg --version`): v0.35.0
- Docker registry used (e.g. `Docker HUB`): Harbor, GCR, Docker Hub

---
Vote on this request

joaopapereira:
- Update the pull API to allow the users to provide an already create Registry instead of the API creating one

Fixes #481

### DIFF
--- a/pkg/imgpkg/registry/registry.go
+++ b/pkg/imgpkg/registry/registry.go
@@ -131,6 +131,15 @@ type SimpleRegistry struct {
 	transportAccess *sync.Mutex
 }
 
+// NewBasicRegistry does not provide any special behavior and all the options as passed as is to the underlying library
+func NewBasicRegistry(regOpts ...regremote.Option) (*SimpleRegistry, error) {
+	return &SimpleRegistry{
+		remoteOpts:      regOpts,
+		roundTrippers:   NewNoopRoundTripperStorage(),
+		transportAccess: &sync.Mutex{},
+	}, nil
+}
+
 // NewSimpleRegistry Builder for a Simple Registry
 func NewSimpleRegistry(opts Opts) (*SimpleRegistry, error) {
 	httpTran, err := newHTTPTransport(opts)
@@ -141,7 +150,7 @@ func NewSimpleRegistry(opts Opts) (*SimpleRegistry, error) {
 }
 
 // NewSimpleRegistryWithTransport Creates a new Simple Registry using the provided transport
-func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper, regOpts ...regremote.Option) (*SimpleRegistry, error) {
+func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper) (*SimpleRegistry, error) {
 	var refOpts []regname.Option
 	if opts.Insecure {
 		refOpts = append(refOpts, regname.Insecure)
@@ -165,9 +174,6 @@ func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper, regOp
 	var regRemoteOptions []regremote.Option
 	if opts.IncludeNonDistributableLayers {
 		regRemoteOptions = append(regRemoteOptions, regremote.WithNondistributable)
-	}
-	if regOpts != nil {
-		regRemoteOptions = append(regRemoteOptions, regOpts...)
 	}
 	tries := opts.RetryCount
 	if tries == 0 {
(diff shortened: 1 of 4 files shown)


pkg/imgpkg/registry/registry.go
```diff
@@ -203,6 +209,14 @@ func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper, regOp
 // the image provided
 // A Registry need to be provided as the first parameter or the function will panic
 func (r SimpleRegistry) CloneWithSingleAuth(imageRef regname.Tag) (Registry, error) {
+	if r.keychain == nil { // If no keychain is present it assumes NewBasicRegistry was used to create the Registry. So we short circuit this execution
```
jtigger, line 214: From a domain perspective, what does it mean for `r.keychain` to be `nil`?

I believe it means that either no keychain is required _or_ the keychain will be created using the `...regremote.Option`s.

As long as those two are the only scenarios where `r.keychain` can be `nil`, this checks out.

Are these the only scenarios where `r.keychain` can be `nil`?

pkg/imgpkg/registry/registry.go
```diff
@@ -262,6 +281,10 @@ func (r *SimpleRegistry) transport(ref regname.Reference, scope string) (http.Ro
 	defer r.transportAccess.Unlock()
 	rt := r.roundTrippers.RoundTripper(ref.Context(), scope)
 	if rt == nil {
+		if r.keychain == nil {
```
jtigger, line 291: ditto, here.  We're basically saying: a) either no auth is required, or b) the credentials _and transport_ are presumed to be configured by some other mechanism (namely through the GGCR options) than what the `SimpleRegistry` typically provides.

That all true?

pkg/imgpkg/registry/registry.go
```diff
@@ -203,6 +209,14 @@ func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper, regOp
 // the image provided
```
jtigger: Seems like the godoc is slightly out-of-date, now.

In the case where there's no keychain configured, we don't replace it at all.

Does this work?

> // CloneWithSingleAuth produces a copy of this Registry whose keychain has exactly one auth — the one that can be used to access `imageRef`. If no keychain is explicitly configured on this Registry, the copy is a BasicRegistry.

pkg/imgpkg/registry/registry.go
```diff
@@ -203,6 +209,14 @@ func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper, regOp
...
 // A Registry need to be provided as the first parameter or the function will panic
 func (r SimpleRegistry) CloneWithSingleAuth(imageRef regname.Tag) (Registry, error) {
+	if r.keychain == nil { // If no keychain is present it assumes NewBasicRegistry was used to create the Registry. So we short circuit this execution
+		return &SimpleRegistry{
+			remoteOpts:      r.remoteOpts,
+			roundTrippers:   NewNoopRoundTripperStorage(),
+			transportAccess: &sync.Mutex{},
+		}, nil
```
jtigger: Replace with a call to `NewBasicRegistry()`?

jtigger (approved):
This all checks out.

I've dropped some gut-check questions, stating some assumptions I have. If they are valid, we're GTG.

One minor suggestion in there to call the `NewBasicRegistry()` constructor when we clone.

All other aspects LGTM. 👍

pkg/imgpkg/registry/registry.go
joaopapereira: The only situation where this can effectively happen is when you create the `SimpleRegistry` using `NewBasicRegistry` because the other constructors always create a keychain

pkg/imgpkg/registry/registry.go
joaopapereira: Yes correct

PR #485, opened 2023-02-13
Speed up multi-image operations

### LINKED ISSUE #483: Frequent calls to docker-credential-desktop and osascript when doing a copy
**What steps did you take:**
I was attempting to copy the image from these instructions: https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/1.4/tap/install-aws.html

The copy was from a password-protected repository at `registry.tanzu.vmware.com` to a non-password-protected docker container registry.

**What happened:**

This is 214 images and approx 7GB of data.  I noticed that it took a long time (more than 2 minutes) to collect all the information for the clone, and while that was happening, the terminal title was flashing between `osascript`, `sw_version` and a few other commands.  A `ps` revealed the following:

```bash
28423 ttys011    0:01.87 imgpkg copy --concurrency=1 -b registry.tanzu.vmware.com/tanzu-application-platform/tap-packages:1.4.0 --to-repo registr
31366 ttys011    0:00.02 docker-credential-desktop get
31370 ttys011    0:00.09 osascript -e user locale of (get system info)
```

As an approximate guess, this looks like approx `(31370 - 28423) / 4 > 700` invocations of `docker-credential-desktop get`, which seems excessive.  This seems to continue; after 5 minutes, there seem to have 1200 invocations (or approx 2/second).  Nothing else seems to be spawning processes on my machine (idle except for typing this bug report into Chrome).  Overall, it took 8 minutes to collect the information for these 214 images, and then approx 1 minute to verify that all the images had been copied to the destination registry (I'd prevously copied most images, so this _should_ have been a metadata-only comparison).

**What did you expect:**
One invocation of `docker-credential-desktop get` per registry, or perhaps one per minute.

This is running 0.35.0; looking through previous bugs, #334 looks related but clearly doesn't seem to have helped in this case.

Running with `--debug` doesn't show a lot of requests to the `/v2` endpoint for authentication as described in #290 

**Environment:**

- imgpkg version (use `imgpkg --version`):
```
imgpkg version v0.35.0
```
Self-install of Docker `registry` image from https://github.com/evankanderson/k8s-private-local-registry

```$ uname -v
Darwin Kernel Version 22.2.0: Fri Nov 11 02:08:47 PST 2022; root:xnu-8792.61.2~4/RELEASE_X86_64
```

---
Vote on this request

evankanderson:
Fixes #483

It turns out that if you pass a fresh keychain into ggcr, it will attempt to resolve the keychain each time, and doesn't keep a per-registry cache.  Add a per-registry authentication cache to the per-registry transport cache.

On a "verify-only" copy of 214 images and 7.06GiB, this reduced time from about 20 minutes to about 3 minutes. Will go back and grab "before" timing in a moment.

```
Succeeded

real	2m53.102s
user	0m2.926s
sys	0m2.721s
```

**Note: I've never touched this code before, so you may want to give this PR extra scrutiny**

### DIFF
--- a/pkg/imgpkg/registry/registry.go
+++ b/pkg/imgpkg/registry/registry.go
@@ -126,6 +126,7 @@ type SimpleRegistry struct {
 	remoteOpts      []regremote.Option
 	refOpts         []regname.Option
 	keychain        regauthn.Keychain
+	authn           map[string]regauthn.Authenticator
 	roundTrippers   RoundTripperStorage
 	transportAccess *sync.Mutex
 }
@@ -195,6 +196,7 @@ func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper, regOp
 		refOpts:         refOpts,
 		keychain:        keychain,
 		roundTrippers:   NewMultiRoundTripperStorage(baseRoundTripper),
+		authn:           map[string]regauthn.Authenticator{},
 		transportAccess: &sync.Mutex{},
 	}, nil
 }
@@ -219,6 +221,7 @@ func (r SimpleRegistry) CloneWithSingleAuth(imageRef regname.Tag) (Registry, err
 		refOpts:         r.refOpts,
 		keychain:        keychain,
 		roundTrippers:   NewSingleTripperStorage(rt),
+		authn:           map[string]regauthn.Authenticator{},
 		transportAccess: &sync.Mutex{},
 	}, nil
 }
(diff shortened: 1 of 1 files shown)


evankanderson:
The same run using the 0.35.0 imgpkg version:

```
Succeeded

real	20m19.801s
user	3m12.668s
sys	5m33.005s
```

joaopapereira (approved):
LGTM

PR #513, opened 2023-04-18
Add preserve-permission flag to push command

### LINKED ISSUE #498: imgpkg removes group permissions when storing files in OCI image, which causes permission problems with the Jammy stack
**What happened:**
Jammy-base builders require read permissions is set for group. It uses different users for build and run, so group read permissions are required to access files.

- Using imgpkg to copy a jammy-base image
- Imgpkg removes group permissions when storing files in OCI image, permissions are set based on unmask
- node/python/gradle workload deployments with "--local-path" will fail. These projects require read access to file system resources.

**What did you expect:**
The permissions should be retained so that Jammy-base images and their workloads can run. 

imgpkg should implement a new flag called --preserve-permissions  that will disable the current behavior for imgpkg that would remove all the permissions. This will ensure we don't cause a breaking change for existing user workflows.

joaopapereira:
Fixes #498


pkg/imgpkg/image/tar_image.go
```diff
@@ -141,12 +151,16 @@ func (i *TarImage) addFileToTar(fullPath, relPath string, info os.FileInfo, tarW
 	if runtime.GOOS == "windows" {
 		relPath = strings.ReplaceAll(relPath, "\\", "/")
 	}
+	filePermission := int64(info.Mode() & 0700)
```
neil-hickey, line 154: Why do we need to make it '7xx' permissions by default exactly ?

I assume by `bitwise AND'ing` we are making it at least 0700 here

pkg/imgpkg/image/dir_image.go
```diff
@@ -135,6 +135,11 @@ func (i *DirImage) extractTarEntry(header *tar.Header, input io.Reader) error {
 	userPermission := int64(mode & 0700)
 	permMode := os.FileMode(userPermission | userPermission>>3 | userPermission>>6)
 
+	// if group/all permission is set assume that the image was created to keep the full original permission set
```
neil-hickey: This comment doesn't really tell me why we need to do this.

> assume that the image was created to keep the full original permission set

What does that mean exactly?

pkg/imgpkg/image/dir_image_test.go
```diff
@@ -61,4 +66,39 @@ func TestDirImage(t *testing.T) {
...
 		}
 	}
+
+	t.Run("When extracting tar that contains files and folders with permissions", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skip("permissions values are different in windows because of umask")
+		}
+		img, err := image.NewFileImage(filepath.Join("test_assets", "img_tar_with_permissions.tar"), nil)
+		require.NoError(t, err)
+		folder, err := os.MkdirTemp("", "test")
```
neil-hickey: non-blocked suggestion.

We might wanna use `T.TempDir()` instead of os.MkdirTemp and defer removal. It actually does it for you.

neil-hickey (approved):
LGTM! I added a question or two around the `07xx` permissions we are enforcing by default. Wondering why we are doing that.

pkg/imgpkg/image/tar_image.go
praveenrewar: > I assume by bitwise AND'ing we are making it at least 0700 here

I am not really good at this, but I think we are making it _at most_ 0700, which does make sense (it would be at least 0700 if we were doing a `bitwise OR`)

pkg/imgpkg/image/dir_image.go
praveenrewar: ^ +1

praveenrewar (approved):
LGTM!

pkg/imgpkg/image/tar_image.go
joaopapereira: The bitwise here ensures that we only capture the part of the permissions associated with the User permissions. So what this does basically is to remove Group and All Permissions

neil-hickey: Ahhh yes, the opposite :D Need to bitwise OR my brain this morning to make sure its all on

PR #521, opened 2023-05-18
Add max-depth flag to describe

netiperher:
When a bundle contains a large set of nested bundles the describe command on the root bundle can take a long time. 

This change introduces a `--max-depth` flag to the describe command in order to limit the depth of recursion that is performed on the bundle. This is similar to the `--no-recursion` flag discussed in #204 .

### DIFF
--- a/pkg/imgpkg/v1/describe.go
+++ b/pkg/imgpkg/v1/describe.go
@@ -61,6 +61,7 @@ type Description struct {
 type DescribeOpts struct {
 	Logger                 bundle.Logger
 	Concurrency            int
+	MaxDepth               int
 	IncludeCosignArtifacts bool
 }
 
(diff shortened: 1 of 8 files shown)


pkg/imgpkg/v1/describe.go
```diff
@@ -98,7 +99,7 @@ func DescribeWithRegistryAndSignatureFetcher(bundleImage string, opts DescribeOp
 		return Description{}, fmt.Errorf("Only bundles can be described, and %s is not a bundle", bundleImage)
 	}
 
-	allBundles, err := newBundle.FetchAllImagesRefs(opts.Concurrency, opts.Logger, sigFetcher)
+	allBundles, err := newBundle.FetchAllImagesRefs(opts.Concurrency, opts.MaxDepth, opts.Logger, sigFetcher)
```
joaopapereira, line 102: Before we call `FetchAllImagesRefs` do you mind checking that the depth is >= 0? Just to ensure people do not try to get a negative number in

joaopapereira (changes requested):
The PR looks pretty good, just have 1 change that I would like to see before we merge this PR

PR #527, opened 2023-05-30
Bump golang 1.20.4, go-containerregistry, protobuf in v0.33.x line

sethiyash:


joaopapereira (approved):
The change looks good, let us wait for the tests to be green and we can merge it

PR #531, opened 2023-05-31
prefix version with v in building binaries while verifying artifacts

sethiyash:
`v` is missing while providing version for building binaries in verify artifacts stage

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,7 +31,7 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/')
         with:
           version: 1.16.2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/')
         with:
           version: 1.16.2
-          args: release --rm-dist --debug ${{ env.SKIP_PUBLISH }}
+          args: release --clean --debug ${{ env.SKIP_PUBLISH }}
```
sethiyash, line 34: `--rm-dist` has been deprecated in favor of `--clean`

PR #536, opened 2023-06-12
Bumping docker and docker/distribution dependency in v0.36.x line

sethiyash:
- Bumping docker and docker/distribution dependency in v0.36.x line


joaopapereira (approved):
LGTM

PR #412, opened 2022-07-09
Add ability to set labels on image push

old-ocean-creature:
Add ability to set labels on image push

PR #552, opened 2023-07-26
Add support for OCI labels on image/bundle push

### LINKED ISSUE #153: Add annotations/labels to bundle
OCI images would be a great place to store machine learning models. ImgPkg works as expected with storing a generic model file, however there are times in which you may want to select between different model versions based on metadata such as the training score or dataset used. 

I would like to add basic metadata to bundles to help select them. Looking for similar functionality to https://docs.docker.com/engine/reference/builder/#label or https://github.com/opencontainers/image-spec/blob/master/annotations.md. Would be nice to also have a way to search or filter them when looking through the repository. 

---
Vote on this request

phenixblue:
This adds the ability for users to specify OCI labels that will be added to the generated OCI artifact (bundle/image) during a `push` operation.

Fixes #153

Example: 

```shell
$ imgpkg push -b index.docker.io/jmsearcy/example-bundle:v1.0.3 -l foo.bar=baz -f ../../tmp/imgpkg-test
```

The resulting OCI artifact can be inspected as such to show the labels added:

```shell
$ regctl image inspect jmsearcy/example-bundle:v1.0.3

{
  "created": "0001-01-01T00:00:00Z",
  "architecture": "",
  "os": "",
  "config": {
    "Labels": {
      "dev.carvel.imgpkg.bundle": "true",
      "foo.bar": "baz",
    }
  },
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:f77fc57aa0878a70576e6b45c74ce5a7070219687b92fe073f632d71b7b107f8"
    ]
  },
  "history": [
    {
      "created": "0001-01-01T00:00:00Z",
      "created_by": "imgpkg",
      "author": "imgpkg"
    }
  ]
}
```

### DIFF
--- a/pkg/imgpkg/bundle/contents.go
+++ b/pkg/imgpkg/bundle/contents.go
@@ -49,6 +49,9 @@ func (b Contents) Push(uploadRef regname.Tag, labels map[string]string, registry
 		return "", err
 	}
 
+	if labels == nil {
+		labels = map[string]string{}
+	}
 	labels[BundleConfigLabel] = "true"
 
 	return plainimage.NewContents(b.paths, b.excludedPaths, b.preservePermissions).Push(uploadRef, labels, registry, logger)
--- a/pkg/imgpkg/bundle/contents.go
+++ b/pkg/imgpkg/bundle/contents.go
@@ -49,6 +49,10 @@ func (b Contents) Push(uploadRef regname.Tag, labels map[string]string, registry
 		return "", err
 	}
 
+	if labels == nil {
+		labels = map[string]string{}
+	}
+
 	labels[BundleConfigLabel] = "true"
 	return plainimage.NewContents(b.paths, b.excludedPaths, b.preservePermissions).Push(uploadRef, labels, registry, logger)
 }
--- a/pkg/imgpkg/bundle/contents.go
+++ b/pkg/imgpkg/bundle/contents.go
@@ -43,13 +43,13 @@ func NewContents(paths []string, excludedPaths []string, preservePermissions boo
 }
 
 // Push the contents of the bundle to the registry as an OCI Image
-func (b Contents) Push(uploadRef regname.Tag, registry ImagesMetadataWriter, logger Logger) (string, error) {
+func (b Contents) Push(uploadRef regname.Tag, labels map[string]string, registry ImagesMetadataWriter, logger Logger) (string, error) {
 	err := b.validate()
 	if err != nil {
 		return "", err
 	}
 
-	labels := map[string]string{BundleConfigLabel: "true"}
+	labels[BundleConfigLabel] = "true"
 	return plainimage.NewContents(b.paths, b.excludedPaths, b.preservePermissions).Push(uploadRef, labels, registry, logger)
 }
 
--- a/pkg/imgpkg/bundle/contents.go
+++ b/pkg/imgpkg/bundle/contents.go
@@ -49,11 +49,8 @@ func (b Contents) Push(uploadRef regname.Tag, labels map[string]string, registry
 		return "", err
 	}
 
-	if labels == nil {
-		labels = map[string]string{}
-	}
-
 	labels[BundleConfigLabel] = "true"
+
 	return plainimage.NewContents(b.paths, b.excludedPaths, b.preservePermissions).Push(uploadRef, labels, registry, logger)
 }
 
--- a/pkg/imgpkg/bundle/contents_test.go
+++ b/pkg/imgpkg/bundle/contents_test.go
@@ -44,7 +44,7 @@ images:
 			t.Fatalf("failed to read tag: %s", err)
 		}
 
-		_, err = subject.Push(imgTag, make(map[string]string), fakeRegistry, util.NewNoopLevelLogger())
+		_, err = subject.Push(imgTag, map[string]string{}, fakeRegistry, util.NewNoopLevelLogger())
 		if err != nil {
 			t.Fatalf("not expecting push to fail: %s", err)
 		}
--- a/pkg/imgpkg/bundle/contents_test.go
+++ b/pkg/imgpkg/bundle/contents_test.go
@@ -44,7 +44,7 @@ images:
 			t.Fatalf("failed to read tag: %s", err)
 		}
 
-		_, err = subject.Push(imgTag, fakeRegistry, util.NewNoopLevelLogger())
+		_, err = subject.Push(imgTag, make(map[string]string), fakeRegistry, util.NewNoopLevelLogger())
 		if err != nil {
 			t.Fatalf("not expecting push to fail: %s", err)
 		}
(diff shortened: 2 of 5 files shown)


phenixblue:
Related to #412

pkg/imgpkg/bundle/contents_test.go
```diff
@@ -44,7 +44,7 @@ images:
 			t.Fatalf("failed to read tag: %s", err)
 		}
 
-		_, err = subject.Push(imgTag, fakeRegistry, util.NewNoopLevelLogger())
+		_, err = subject.Push(imgTag, make(map[string]string), fakeRegistry, util.NewNoopLevelLogger())
```
joaopapereira:
```suggestion
		_, err = subject.Push(imgTag, map[string]string{}, fakeRegistry, util.NewNoopLevelLogger())
```
for consistency with next change

pkg/imgpkg/bundle/contents.go
```diff
@@ -43,13 +43,14 @@ func NewContents(paths []string, excludedPaths []string, preservePermissions boo
...
 // Push the contents of the bundle to the registry as an OCI Image
-func (b Contents) Push(uploadRef regname.Tag, registry ImagesMetadataWriter, logger Logger) (string, error) {
+func (b Contents) Push(uploadRef regname.Tag, labels map[string]string, registry ImagesMetadataWriter, logger Logger) (string, error) {
 	err := b.validate()
 	if err != nil {
 		return "", err
 	}
 
-	labels := map[string]string{BundleConfigLabel: "true"}
+	labels[BundleConfigLabel] = "true"
```
joaopapereira, line 55: To cross out T's here, do you mind checking if the `labels` is not nil? If it is, we can create a map locally.

pkg/imgpkg/cmd/push.go
```diff
@@ -56,11 +59,20 @@ func (po *PushOptions) Run() error {
...
+	if err != nil {
+		return err
+	}
+
 	var imageURL string
 
 	isBundle := po.BundleFlags.Bundle != ""
 	isImage := po.ImageFlags.Image != ""
 
+	if po.LabelFlags.Labels == nil {
```
joaopapereira: Do we need this check here if we set a default in the LabelsFlags?

pkg/imgpkg/cmd/label_flags.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+type LabelFlags struct {
+	Labels map[string]string
+}
+
+func (l *LabelFlags) Set(cmd *cobra.Command) {
+	cmd.Flags().StringToStringVarP(&l.Labels, "labels", "l", make(map[string]string), "Set labels on image")
```
joaopapereira: How much more complicated do you think it would be if we allowed the following
`imgpkg push -f folder -i registry.io/my-img -l my.label=10 -l my.otherlabel="something interesting" -l yet.another.label="this is the real deal"`?
I think we might have to make this a `StringSliceVarP` and we would have to process each value to split by `=`.

pkg/imgpkg/cmd/push_test.go
```diff
@@ -235,6 +239,86 @@ func TestImageAndBundleLockError(t *testing.T) {
 	}
 }
 
+func TestLabels(t *testing.T) {
```
joaopapereira, line 242: The only other test that I think we can do is when you provide a multi work label value like: `my.label="something that has multiple words"`

joaopapereira (changes requested):
The change looks good to me. Added some questions and suggestions.

pkg/imgpkg/bundle/contents_test.go
phenixblue: Ah, yeah, I had that originally and was testing some things and forgot to change it back.

pkg/imgpkg/bundle/contents.go
phenixblue: I have a check for that in the PushOptions `Run()` method in the [/pkg/imgpkg/cmd/push.go](https://github.com/phenixblue/imgpkg/blob/e03fe72406ac4f3b634395a727a9a5ccb25fce20/pkg/imgpkg/cmd/push.go#L72) file that gets called before either `pushBundle()` or `pushImage` do, which is the caller to this area of code. Do you think it's better to explicitly have the check in each `Push()` method for Images and Bundles?

pkg/imgpkg/cmd/push.go
phenixblue: I didn't think so at first, but there were some nil map assignment errors in some of the tests when I didn't have this. Having this here also prevented having to do the check for both Images and Bundles separately since they seem to both start from this code path.

pkg/imgpkg/cmd/label_flags.go
phenixblue: I think that works as-is right now, or maybe I'm not understanding what you're suggesting.

```shell
$ ./imgpkg push -i jmsearcy/test-imgpkg-image -f ./pkg/imgpkg/cmd/test_assets/image_with_config -l my.label=10 -l my.otherlabel="something interesting" -l yet.another.label="this is the real deal"

dir: .
file: config.yml
Pushed 'index.docker.io/jmsearcy/test-imgpkg-image@sha256:acafc2d67e1fa9093d08c75399b5e60f0e60851741bb3e604195db19b3a09a40'
Succeeded
```

```shell
$ regctl image inspect jmsearcy/test-imgpkg-image

{
  "created": "0001-01-01T00:00:00Z",
  "architecture": "",
  "os": "",
  "config": {
    "Labels": {
      "my.label": "10",
      "my.otherlabel": "something interesting",
      "yet.another.label": "this is the real deal"
    }
  },
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:284871f9fdadd60646be5ee42a95afa16c9352ef41ef5c1490dce02aa08bb338"
    ]
  },
  "history": [
    {
      "created": "0001-01-01T00:00:00Z",
      "created_by": "imgpkg",
      "author": "imgpkg"
    }
  ]
}
```

joaopapereira: My bad. I assume that you needed a SliceVar to be able to use multiple flags

pkg/imgpkg/bundle/contents.go
joaopapereira: My main concern is that we assume that we can just update the variable in this particular place, so I would prefer the check to live at this level. The next time `labels` is used is inside NewFileImage function, and we just check for `len` of it, and if it is `nil` it will not fail.

pkg/imgpkg/cmd/push_test.go
phenixblue: Done!

pkg/imgpkg/bundle/contents.go
phenixblue: OK. I will move the check for nil map to here.

pkg/imgpkg/cmd/push.go
phenixblue: Removing this from here in favor of having it in the bundle `Push()` method as per the[ above thread]([pkg/imgpkg/bundle/contents.go](https://github.com/carvel-dev/imgpkg/pull/552/files/a6db6ee2ab2f394a760c1ca5259b8a6500381255#diff-2d42d9f47abf1127955d8688255d683771a9c39d03d08d3f84a44c43a599c32f))

joaopapereira (approved):
LGTM

Issue #544, opened 2023-06-21
Copying a bundle with the `image-is-bundle-check` being false doesn't work as expected.


praveenrewar:
**What steps did you take:**

```bash
 ## Case 1
 $ imgpkg pull -b this-is-a-bundle-for-sure
 
 ## Case 2
 $ imgpkg pull -b this-is-a-bundle-for-sure --image-is-bundle-check=false
```

When we run the above 2 commands, the expectation would be to have the same result, but right now because of [this](https://github.com/carvel-dev/imgpkg/blob/1a02bc4ced5e9b0d1cd13ae080a1f830ef8a4ee8/pkg/imgpkg/cmd/pull.go#L84) and [this](https://github.com/carvel-dev/imgpkg/blob/1a02bc4ced5e9b0d1cd13ae080a1f830ef8a4ee8/pkg/imgpkg/v1/pull.go#L105), in the second scenario we will pull it as an image, which might be a bit weird because we are trying to pull a bundle using the bundle flag.

**Anything else you would like to add:**
The value of the flag `image-is-bundle-check` shouldn't effect copying a bundle or we shouldn't allow using the flag itself.

**Environment:**

- imgpkg version (use `imgpkg --version`):
- Docker registry used (e.g. `Docker HUB`): 
- OS (e.g. from `/etc/os-release`):

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

joaopapereira:
Hey
I think we should not allow this flag to be used with `-b`. It should throw an error pointing the user to use the `-i` flag.

PR #563, opened 2023-08-19
Error when --image-is-bundle-check flag is set to false while using -b flag 

ashpect:
This solves #544 

## Issue 
The command 
```
imgpkg pull -b index.docker.io/<user>/<repo>:<tag> --image-is-bundle-check=false -o <location>
```
successfully runs and will pull it as an image, which is a bit weird because we are trying to pull a bundle using the bundle flag.

## Fix
Now running the command throws this error 
```
imgpkg: Error: Setting --image-is-bundle-check flag to false is not allowed when using -b flag.
```

### DIFF
--- a/pkg/imgpkg/cmd/pull.go
+++ b/pkg/imgpkg/cmd/pull.go
@@ -130,5 +130,9 @@ func (po *PullOptions) validate() error {
 	if po.BundleRecursiveFlags.Recursive && len(po.ImageFlags.Image) > 0 {
 		return fmt.Errorf("Cannot use --recursive (-r) flag when pulling a bundle")
 	}
+
+	if !po.ImageIsBundleCheck && len(po.BundleFlags.Bundle) != 0 {
+		return fmt.Errorf("Cannot set --image-is-bundle-check while using -b flag")
+	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/imgpkg/cmd/pull.go
```diff
@@ -130,5 +130,9 @@ func (po *PullOptions) validate() error {
 	if po.BundleRecursiveFlags.Recursive && len(po.ImageFlags.Image) > 0 {
 		return fmt.Errorf("Cannot use --recursive (-r) flag when pulling a bundle")
 	}
+
+	if !po.ImageIsBundleCheck && len(po.BundleFlags.Bundle) != 0 {
+		return fmt.Errorf("Setting --image-is-bundle-check flag to false is not allowed when using -b flag.")
```
praveenrewar: Should we just say that the flag cannot be used while using `-b`? The default value of the flag is true, so technically that will always be set, but let's not confuse users by saying that they can set it to true but cannot set it to false.

ashpect: Updated

praveenrewar (commented):
Thank you for working on this @ashpect. The change looks good to me. Would you like to add a test for this as well?

ashpect:
Yes, definitely.

ashpect:
Hello @praveenrewar. I added a test for the fix. 
Please review it whenever you get the time. 🙌 


Screenshot 2023-08-24 at 11 07 14 PM

joaopapereira (approved):
LGTM

PR #569, opened 2023-08-28
Updated test to build binary locally based on variable value

kumaritanushree:
Updated test to build binary locally based on variable value. Default it is set to true. If you do not want to build binary through test-all.sh run: `export BUILD_LOCAL=false`

### DIFF
--- a/hack/test-all.sh
+++ b/hack/test-all.sh
@@ -2,7 +2,11 @@
 
 set -e -x -u
 
-./hack/build.sh
+# By default it will generate binary of imgpkg to run test. 
+# export BUILD_BINARY=false to skip binary generation
+if [ ${BUILD_BINARY:-true} == true ]; then
+    ./hack/build.sh
+fi
 
 export IMGPKG_BINARY="$PWD/imgpkg${IMGPKG_BINARY_EXT-}"
 


hack/test-all.sh
```diff
@@ -2,7 +2,10 @@
 
 set -e -x -u
 
-./hack/build.sh
+if [ ${BUILD_LOCAL:-true} == true ]; then
```
rohitagg2020: Rename it to `BUILD_BINARY`

hack/test-all.sh
```diff
@@ -2,7 +2,10 @@
...
+    echo "build-local"
```
rohitagg2020: Remove echo

hack/test-all.sh
```diff
@@ -2,7 +2,10 @@
 
 set -e -x -u
 
-./hack/build.sh
```
rohitagg2020, line 5: Add a comment which says how to use it.

kumaritanushree: done

hack/test-all.sh
kumaritanushree: done

hack/test-all.sh
kumaritanushree: done

joaopapereira (approved):
LGTM

PR #579, opened 2023-09-19
Bumping go version to 1.21.1

sethiyash:
- Bumping go version to 1.21.1


joaopapereira:
These tests are failing because we do not allow the Workflow files to be changed when we are doing end-to-end tests.

praveenrewar (approved):

PR #612, opened 2023-11-24
Saving bundles as an intermediate ocitar

ashpect:
Implements [this](https://github.com/carvel-dev/carvel/pull/692) proposal.
Implemented except writing the tests, waiting for initial review, after which e2e tests can be written.

Screenshot 2024-05-23 at 10 14 29 PM


pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -60,49 +63,73 @@ func (c CopyRepoSrc) CopyToTar(dstPath string, resume bool) error {
...
+				return nil, err
+			}
+			err = image.ExtractOciTarGz(c.OciFlags.OcitoReg, tempDir)
+			if err != nil {
+				return nil, fmt.Errorf("Extracting OCI tar: %s", err)
+			}
+			processedImages, err = c.tarImageSet.Import(tempDir, importRepo, c.registry, true)
+			if err != nil {
+				return nil, fmt.Errorf("Importing OCI tar: %s", err)
+			}
```
joaopapereira, line 89: Since this is similar to the else case, maybe we can move this outside the if statement

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -60,49 +63,73 @@ func (c CopyRepoSrc) CopyToTar(dstPath string, resume bool) error {
...
 			return nil, fmt.Errorf("Cannot use tar source (--tar) with tar destination (--to-tar)")
 		}
 
-		processedImages, err = c.tarImageSet.Import(c.TarFlags.TarSrc, importRepo, c.registry)
+		if c.OciFlags.IsOci() {
+			tempDir, err := os.MkdirTemp("", "imgpkg-oci-extract-")
+			if err != nil {
+				return nil, err
+			}
+			err = image.ExtractOciTarGz(c.OciFlags.OcitoReg, tempDir)
```
joaopapereira, line 82: Let us make sure we clean up after ourselves
```suggestion
                         defer os.RemoveAll(tempDir)
			err = image.ExtractOciTarGz(c.OciFlags.OcitoReg, tempDir)
```

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -60,49 +63,73 @@ func (c CopyRepoSrc) CopyToTar(dstPath string, resume bool) error {
...
 			return nil, err
 		}
 
-		var parentBundle *ctlbundle.Bundle
-		foundRootBundle := false
-		for _, processedImage := range processedImages.All() {
-			if processedImage.ImageIndex != nil {
-				continue
-			}
+		// This is added to not read the lockfile and change the ref for oci-flag. Will be removed once we add an inflate option to copy the refs.
```
joaopapereira: Is this commented out because if we copy from oci-tar to the repository, we do not have all the images? If that is the case, we should change the comment to highlight something like:
```
// When copying images from an oci-tar to a repository, imgpkg will not try to access the origin repositories and will only copy the OCI Image to the registry, similar to the behavior we currently have on the `imgpkg push` command.
```

pkg/imgpkg/cmd/copy_repo_src.go
```diff
@@ -133,6 +160,11 @@ func (c CopyRepoSrc) CopyToRepo(repo string) (*ctlimgset.ProcessedImages, error)
 		return nil, fmt.Errorf("Tagging images: %s", err)
 	}
 
+	err = os.RemoveAll(tempDir)
+	if err != nil {
+		fmt.Println("Error cleaning up temporary directory:", err)
+	}
+
```
joaopapereira: Added a prior comment that I think it would be better to use than this part of the code. Because if anything fails in the meanwhile we will leave the folder back
```suggestion
```

pkg/imgpkg/cmd/oci_flags.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2023 The Carvel Authors.
```
joaopapereira:
```suggestion
// Copyright 2024 The Carvel Authors.
```

pkg/imgpkg/image/tar_image.go
```diff
@@ -180,3 +181,119 @@ func (i *TarImage) isExcluded(relPath string) bool {
 	}
 	return false
 }
+
+// CreateOciTarFileAndDeleteFolder creates a oci tar file from the source folder and deletes the source folder. Used while pushing the oci-tar
+func CreateOciTarFileAndDeleteFolder(source, target string) error {
+
```
joaopapereira:
```suggestion
```

pkg/imgpkg/image/tar_image.go
```diff
@@ -180,3 +181,119 @@ func (i *TarImage) isExcluded(relPath string) bool {
 	}
 	return false
 }
+
+// CreateOciTarFileAndDeleteFolder creates a oci tar file from the source folder and deletes the source folder. Used while pushing the oci-tar
```
joaopapereira: Why are we deleting the source folder?
If I do `imgpkg push -b repo/app1-config -f config/ --to-oci-tar /path/to/file.tar` is my `config` folder going to be deleted? We should not do that

pkg/imgpkg/image/tar_image.go
```diff
@@ -180,3 +181,119 @@ func (i *TarImage) isExcluded(relPath string) bool {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ExtractOciTarGz extracts the oci tar file to the extractDir
+func ExtractOciTarGz(inputDir, extractDir string) error {
+
```
joaopapereira:
```suggestion
```

pkg/imgpkg/imagedesc/image_intermediate.go
```diff
@@ -0,0 +1,156 @@
+// Copyright 2023 The Carvel Authors.
```
joaopapereira:
```suggestion
// Copyright 2024 The Carvel Authors.
```

joaopapereira (changes requested):
Thanks for the PR, made some comments in the code.
On a more generic note:
- We need to add some tests to this new feature
- Can we use ggcr to read/write the tarballs for us, since this is a regular OCI tar we should be able to use something like https://github.com/google/go-containerregistry/tree/main/pkg/v1/tarball to do this work. 
- In terms of form, I think I would prefer us to provide variables like `ociTarPath` to the Push functions instead of providing them to the constructors since that has been the pattern we have been using. The idea here is that when you call the function you choose the Repository/location where you want to store the result. Another option that might make more sense here is instead of calling NewContents().Push maybe we should have a new function function called NewContents().PushToTar() since we do not need the majority of the code in Push because we are not building tags or writing them

pkg/imgpkg/cmd/copy_repo_src.go
ashpect: The else case is calling the import function with false as a parameter while here its true which in turns calls the readoci tar different to read tar depending on the bool in the import function.
Screenshot 2024-05-25 at 1 45 45 PM

renuy:
@ashpect , could i request you to rebase and send for review again. (unfortunately, it will not be a straight forward rebase)
Thank you!

ashpect:
Yes, sure. I will get to it this upcoming week.

Issue #636, opened 2024-02-22
imgpkg v0.40.0 breaks with relocated image


mstergianis:
**What steps did you take:**
1. Create a bundle that references an image. It could be anything, I chose nginx for my example
2. Relocate (`imgpkg copy`) that bundle to a second private registry. The scenario I'm simulating here is moving something from an company's internal registry, to a company's public yet authenticated registry.
3. log out of the first registry. So as to simulate a customer that only has access to the public registry
4. imgpkg describe the image that is on the public registry

Here's a script that roughly sums up what I did to replicate this from scratch
```shell
mkdir -p bundle/.imgpkg/
cat >bundle/.imgpkg/images.yml <<EOF
apiVersion: imgpkg.carvel.dev/v1alpha1
images:
- annotations:
    kbld.carvel.dev/id: tdp-configurator
  image: FIRST-PRIVATE-REGISTRY/SOME-URI@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef

kind: ImagesLock
EOF

imgpkg push -b FIRST-PRIVATE-REGISTRY/SOME-BUNDLE-URI -f bundle/

imgpkg copy -b FIRST-PRIVATE-REGISTRY/SOME-BUNDLE-URI@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef --to-repo SECOND-PRIVATE-REGISTRY/SOME-BUNDLE-URI

docker logout FIRST-PRIVATE-REGISTRY

imgpkg describe -b SECOND-PRIVATE-REGISTRY/SOME-BUNDLE-URI@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
# imgpkg: Error: Error:
#   GET https://FIRST-PRIVATE-REGISTRY/v2/SOME-BUNDLE-URI/manifests/sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef:
#     UNAUTHORIZED: unauthorized to access repository: SOME-BUNDLE-URI, action: pull: unauthorized to access repository: SOME-BUNDLE-URI, action: pull in getting remote access of image FIRST-PRIVATE-REGISTRY/SOME-BUNDLE-URI@sha256:0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
```

**What happened:**
`imgpkg describe` fails because it tries to list the layers of an image it can't access

**What did you expect:**
[...]

PR #637, opened 2024-02-22
Collect description information for images from correct registry

joaopapereira:
In this PR we only fix the fact that we are using the correct registry to retrieve information from.

Related to #636


praveenrewar (approved):
LGTM!
@joaopapereira Could you rebase the PR.

PR #638, opened 2024-02-23
When 2 images with same SHA are reference by 2 nested bundles

joaopapereira:
When there are 2 nested bundles that reference the same image (AKA the same SHA) these images are considered as separated images internally. This behavior causes problems when imgpkg is trying to create the Locations image because it searches for the images by SHA, creating a dupplicated entry.

To fix this we decided to force imgpkg to select only one of the entries to add to the Locations list.

This could be averted as well if we stored only 1 copy of these images in the UnprocessedImagesRef struct the main issue is that this is going to hide from imgpkg that the same image is referenced from 2 different sources, this might cause problems in the long run, specially if one day we decide to implement the ability for users to copy bundles to different locations in the destination repository.


praveenrewar (approved):
LGTM!
Ideally it would be good if there are tests for this, but since we have tested it manually and since it's a blocker, we will merge this and make a release, and add a test later on.

PR #639, opened 2024-02-26
Added flag --layers to enable/disable functionality of showing each layers of image

kumaritanushree:


pkg/imgpkg/cmd/describe.go
```diff
@@ -81,7 +84,7 @@ func (d *DescribeOptions) Run() error {
 	ttyEnabledLogger := util.NewUILevelLogger(logLevel, util.NewLoggerNoTTY(d.ui))
 	if d.OutputType == "text" {
 		p := bundleTextPrinter{logger: ttyEnabledLogger}
-		p.Print(description)
+		p.Print(description, d.Layers)
```
joaopapereira: We are not providing this option for the yaml output?

pkg/imgpkg/cmd/describe.go
```diff
@@ -138,14 +141,16 @@ func (p bundleTextPrinter) printerRec(description v1.Description, originalLogger
 		indentLogger.Logf("- Image: %s\n", b.Image)
 		indentLogger.Logf("  Type: Bundle\n")
 		indentLogger.Logf("  Origin: %s\n", b.Origin)
-		indentLogger.Logf("  Layers:\n")
-		for _, d := range b.Layers {
-			indentLogger.Logf("    - Digest: %s\n", d.Digest)
+		if layers {
```
joaopapereira: instead of having a variable present can we jus change this if to check on `len(b.Layers)`?

joaopapereira (changes requested):
Added some comments.

pkg/imgpkg/cmd/describe.go
kumaritanushree: In yaml with empty object it will not show the layers in output only.

pkg/imgpkg/cmd/describe.go
kumaritanushree: updated

joaopapereira (approved):
LGTM

PR #664, opened 2024-05-14
Limit session IDs to a non-negative, pseudo-random int32

### LINKED ISSUE #663: Cannot build for `GOARCH=386`
**What steps did you take:**

Build for, say, `windows/386`:

```
GOOS=windows GOARCH=386 go build ./...
```

**What happened:**

Int overflow. 🤖 💀 

```
# carvel.dev/imgpkg/pkg/imgpkg/registry
pkg/imgpkg/registry/registry.go:202:36: cannot use 9999999999 (untyped int constant) as int value in argument to rand.Intn (overflows)
```

**What did you expect:**

It builds w/o problems.

`imgpkg` is not distributed for `*/386`. That's not something we need to change. However, since `carvel.dev/imgpkg` can be consumed as a library, it should be compatible.

**Environment:**

- imgpkg version (use `imgpkg --version`): n/a
- Docker registry used (e.g. `Docker HUB`): n/a
- OS: `windows/386`

---
Vote on this request

mamachanko:
Technically, this reduces the set from which session IDs are drawn by (9.999.999.999 - 2.147.483.647) = 7.852.516.352, say 75%.

Practically, ~2e10 possible session IDs should be large enough to avoid collisions in the real-world.

This fixes an int overflow for 386 architectures.

Fixes #663

### DIFF
--- a/pkg/imgpkg/registry/registry.go
+++ b/pkg/imgpkg/registry/registry.go
@@ -199,7 +199,7 @@ func NewSimpleRegistryWithTransport(opts Opts, rTripper http.RoundTripper) (*Sim
 
 	sessionID := opts.SessionID
 	if sessionID == "" {
-		sessionID = fmt.Sprint(rand.Intn(9999999999))
+		sessionID = fmt.Sprint(rand.Int31())
 	}
 	baseRoundTripper = NewImgpkgRoundTripper(baseRoundTripper, sessionID)
 


joaopapereira (approved):
I am ok with the change. If we see any issue with this session ID in the future, we can always go for an Alphanumerical session ID, which will increase the number of possibilities again.

PR #668, opened 2024-05-30
Copy api

joaopapereira:
There is a breaking change on the `v1.SignatureFetcher` I decided to do this so that we do not have another interface to do the same thing. Nevertheless we can create a new Interface and eventually deprecate this one.


pkg/imgpkg/v1/copy.go
```diff
@@ -17,61 +17,57 @@ import (
 	regname "github.com/google/go-containerregistry/pkg/name"
 )
 
-type SignatureRetriever interface {
-	Fetch(images *imageset.UnprocessedImageRefs) (*imageset.UnprocessedImageRefs, error)
-}
+const rootBundleLabelKey string = "dev.carvel.imgpkg.copy.root-bundle"
```
praveenrewar, line 20: This doesn't seem to be used anywhere 🤔

pkg/imgpkg/v1/test_assets/bundle_with_mult_images/.imgpkg/bundle.yml
```diff
@@ -0,0 +1,9 @@
+apiVersion: imgpkg.carvel.dev/v1alpha1
+kind: Bundle
+metadata:
+  name: basic
+authors:
+- name: Carvel Team
+  email: carvel@vmware.com
```
praveenrewar: nit: carvel@cncf.com

praveenrewar (commented):
(Still going through but mostly looks pretty good)

pkg/imgpkg/v1/copy.go
joaopapereira: That is a visualization problem in GitHub :( 
Because git considers the file a move, it only shows in the diff the line changes that are different from the original file. This const is used in 4 places inside the v1/copy.go file

praveenrewar (approved):

praveenrewar:
@joaopapereira I think a rebase will be required to trigger the tests again (without the harbor test).

PR #669, opened 2024-06-06
Remove Harbor tests

joaopapereira:
For now we are removing these Harbor tests because the registry we were using is being decomissioned and there is no Harbor replacement. In the future if we see problems raise with Harbor we will need to host a Harbor somewhere and add these tests back.


praveenrewar (approved):
LGTM! Since the tests are being run from the develop branch we will have to merge the PR with pending test.

PR #728, opened 2025-02-28
Fix resume functionality

joaopapereira:


pkg/imgpkg/imagetar/tar_reader.go
```diff
@@ -33,101 +39,132 @@ func (r TarReader) Read() ([]imagedesc.ImageOrIndex, error) {
...
 		}
-		r, err := layer.Compressed()
-		if err != nil {
+		r.lock.Lock()
+		if _, ok := layerCache[h.String()]; ok {
+			errChan <- nil
+			r.lock.Unlock()
 			continue
 		}
+		layerCache[h.String()] = nil
```
100mik, line 94: Won't this be nil by default if there is not hit for the value?

pkg/imgpkg/imagetar/tar_reader.go
```diff
@@ -33,101 +39,132 @@ func (r TarReader) Read() ([]imagedesc.ImageOrIndex, error) {
...
+			layerCache[h.String()] = layer
+			r.lock.Unlock()
+			errChan <- nil
+		}()
 	}
-	return result, nil
+
+	for range layers {
+		if err := <-errChan; err != nil {
+			return err
```
100mik, line 135: Is there any value in aggregating errors in case of multiple ones here?

pkg/imgpkg/imagetar/tar_reader.go
joaopapereira: Yes, this is telling the rest of the loop that there is already 1 thread handling this layer so it can be skipped.

pkg/imgpkg/imagetar/tar_reader.go
joaopapereira: we are failing fast here but I'm not sure if it is worth it or not to wait for all the layers.

praveenrewar (approved):

PR #734, opened 2025-03-18
Increase copy resume performance

joaopapereira:
With this change, we will use the same file provided to be resumed. We do this to increase the performance, previously we were copying the file to the side, which sounds safer but the cost of the
copy vs move is too high.
This change saved almost 20 minutes from a copy that was taking around 1h10.


devacts (approved):
LGTM 👍

Issue #757, opened 2025-10-28
Ability to shallow-copy a bundle (image-is-bundle-check for copy)


rosenhouse:
**Describe the problem/challenge you have**
I'd like to be able to use `imgpkg` to make an exact, shallow copy of a bundle.

Currently if I do
```
imgpkg copy -i example.com/my-src-bundle... --to-repo private.registry.corp/shallow-copy-of-my-bundle
```
then I see

```
imgpkg: Error: Expected bundle flag when copying a bundle (hint: Use -b instead of -i for bundles)
```

and if I add in `--image-is-bundle-check=false` then I get an error about an unknown flag.

**Describe the solution you'd like**
If I add in `--image-is-bundle-check=false` then the above command should succeed (analogous to `pull`).

The resulting artifact should be copied, digest-exact, to the destination repo.

All other flags, like `--tar` / `--to-tar` and `--cosign-signatures` should work as expected.

**Anything else you would like to add:**
I'd be happy to contribute a pull-request for this feature. 

But tell me if it's a bad idea!

rosenhouse:
Fixed by #758

PR #756, opened 2025-10-27
Replace gcr with ghcr to run external registry tests

devacts:

PR #758, opened 2025-10-28
Allow shallow-copy of a bundle

rosenhouse:
This addresses #757 by adding an new `--ignore-bundle-check` flag to `imgpkg copy`.  That enables shallow-copying a bundle, by treating it like an opaque image.

e.g.
```
imgpkg copy --ignore-bundle-check -i 'example.com/my-bundle:foo' --to-repo 'destination.corp/a-shallow-copy'
```

It also works with `--to-tar` (destination) and `--tar` (source) and with and without `--cosign-signatures` although I didn't add tests specifically for those bits.

Update: originally this flag was `--image-is-bundle-check` (default true) to match the analogous flag on `pull`.  After review, it is now named `--ignore-bundle-check` (default false) to make it more clear.

### DIFF
--- a/pkg/imgpkg/cmd/copy.go
+++ b/pkg/imgpkg/cmd/copy.go
@@ -23,14 +23,14 @@ import (
 type CopyOptions struct {
 	ui ui.UI
 
-	ImageFlags         ImageFlags
-	ImageIsBundleCheck bool
-	BundleFlags        BundleFlags
-	LockInputFlags     LockInputFlags
-	LockOutputFlags    LockOutputFlags
-	TarFlags           TarFlags
-	RegistryFlags      RegistryFlags
-	SignatureFlags     SignatureFlags
+	ImageFlags        ImageFlags
+	IgnoreBundleCheck bool
+	BundleFlags       BundleFlags
+	LockInputFlags    LockInputFlags
+	LockOutputFlags   LockOutputFlags
+	TarFlags          TarFlags
+	RegistryFlags     RegistryFlags
+	SignatureFlags    SignatureFlags
 
 	RepoDst string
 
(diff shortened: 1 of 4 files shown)


rosenhouse:
~Failing tests might get fixed by #756 ?~ Update: rebased on main, hoping that fixes CI.

pkg/imgpkg/cmd/copy.go
```diff
@@ -73,6 +74,7 @@ func NewCopyCmd(o *CopyOptions) *cobra.Command {
 	}
 
 	o.ImageFlags.SetCopy(cmd)
+	cmd.Flags().BoolVar(&o.ImageIsBundleCheck, "image-is-bundle-check", true, "Error when image is a bundle (disable shallow-copying bundles via -i)")
```
joaopapereira: Not sure I like the phrasing here, I know that we used the same verbiage in other places but it feels hard to understand. What do you think about `ignore-bundle-check` and for the helper text maybe `When using -i imgpkg will not check if the image is a bundle. This will allow shallow-copying the bundle image.`?

pkg/imgpkg/v1/copy.go
```diff
@@ -220,7 +221,7 @@ func getProvidedSourceImages(origin CopyOrigin, reg registry.Registry, opts Copy
 		if err != nil {
 			return nil, nil, err
 		}
-		if ok {
+		if ok && !opts.AllowShallowCopyBundle {
```
joaopapereira: Given that `v1` serves as an API I think we should have a test to ensure the API is working as expected and not only rely on the e2e tests.

joaopapereira (changes requested):
Thanks for the PR. I added 2 comments, do you think you can address them?

pkg/imgpkg/cmd/copy.go
rosenhouse: Yes, that is clearer.  Updated.

pkg/imgpkg/v1/copy.go
rosenhouse: Yes.  I've added cases in `pkg/imgpkg/v1/copy_test.go` within the `ToTar` test , to cover the success case and the missing-flag case.  Let me know if I should add similar cases to `ToRepo` or elsewhere.

joaopapereira (approved):
LGTM.
The only thing that I am not a fan of is for the API to return errors that only make sense for the CLI, but I guess that is a bigger issues that we have to address in other places as well.

rosenhouse:
Thanks!

PR #763, opened 2025-11-03
Remove ghcr workflow run on PR trigger

devacts:
The external registry check on ghcr can be triggered

1. manually using and workflow_dispatch
2. on push to develop

The manual workflow_dispatch for `test-gh-external-registries-ghcr` check has succeeded
https://github.com/carvel-dev/imgpkg/actions/runs/19057882451/job/54431827100

After this merge it will run only on push to develop and manual trigger

### DIFF
--- a/.github/workflows/gh-test-external-reg-ghcr.yml
+++ b/.github/workflows/gh-test-external-reg-ghcr.yml
@@ -12,8 +12,6 @@ on:
   push:
     branches:
       - develop
-  pull_request:
-    types: [opened, reopened, synchronize, edited]
 
 permissions:
   contents: read


joaopapereira:
I am lost here. Can you explain why we are removing this workflow?

devacts:
> I am lost here. Can you explain why we are removing this workflow?

One of the challenges with ghcr packages is that it downgrades the access for a regualar ephemeral github token to read only from forked PRs even though the workflow config has `package:write` permissions
To overcome that I had configured a PAT with package write access but that secret is not accessible from forked PR (found out after merging the last fix)

One final fix could be add an option to dispatch workflow trigger using an [issue comment](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#issue_comment)

But temporarily, to unblock some of the existing PRs from forks which are failing ghcr external registry test, I am making this change to remove the test on pull request trigger only
The workflow can still be triggered manually using workflow_dispatch option from github action UI.

Screenshot 2025-11-04 at 12 35 57 PM

this passed because it ran on develop branch of default repo and has access to PAT.

Also the test will run on push to develop trigger (i.e. once the PR is merged to develop)

joaopapereira (approved):
Lgtm

PR #767, opened 2026-01-15
Bump golang to 1.25.5

CodesbyUnnati:
Bumping Golang to 1.25.5 to fix the following CVEs:

```
┌─────────────────────┬────────────────┬──────────┬────────┬───────────────────┬───────────────┬─────────────────────────────────────────────────────┐
│       Library       │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version │                        Title                        │
├─────────────────────┼────────────────┼──────────┼────────┼───────────────────┼───────────────┼─────────────────────────────────────────────────────┤
│ golang.org/x/crypto │ CVE-2025-47914 │ MEDIUM   │ fixed  │ v0.35.0           │ 0.45.0        │ golang.org/x/crypto/ssh/agent: in                   │
│                     │                │          │        │                   │               │ golang.org/x/crypto/ssh/agent                       │
│                     │                │          │        │                   │               │ https://avd.aquasec.com/nvd/cve-2025-47914          │
│                     ├────────────────┤          │        │                   │               ├─────────────────────────────────────────────────────┤
│                     │ CVE-2025-58181 │          │        │                   │               │ golang.org/x/crypto/ssh: in golang.org/x/crypto/ssh │
│                     │                │          │        │                   │               │ https://avd.aquasec.com/nvd/cve-2025-58181          │
└─────────────────────┴────────────────┴──────────┴────────┴───────────────────┴───────────────┴──────────────
```
Bumping golangci-lint to v2.4.0 introduced a bunch of lint failures that required base code updation and updating/removing some lint syntax.
- Remove typecheck that is now default checked by the linter.
- Updating new-from-rev SHA to point to develop branch of imgpkg to resolve a bunch of lint errors that required a bunch of base code updates.

This PR also fixes a TestToTarImage failure on Windows that appeared after upgrading golangci-lint to v2.4.0.

Error: The test was failing with an "[access denied](https://github.com/carvel-dev/imgpkg/actions/runs/21201927734/job/60989278686#step:6:1595)" error during the "Resume" test case. This happened because:

- The test opened the generated tarball (bundle.tar) to verify its contents.
- While that file was still open (locked) for reading, the test attempted to run CopyToTar with Resume=true.
- The Resume operation tries to move/rename the file, but Windows strictly blocks moving files that are currently open by another process(reader).

Fix: I updated the test to verify a temporary copy of the tarball instead of the original.

- We verify the layers using this temporary copy and then immediately delete it.
- We also updated the test to store layer data (Digest/Size) in a simple struct instead of keeping the Layer objects in memory (which kept file handles open).
- This ensures the original file is completely unlocked when the "Resume" operation tries to move it.


CodesbyUnnati:
Closing in favour of https://github.com/carvel-dev/imgpkg/pull/768

PR #790, opened 2026-07-20
Include bundle metadata in describe output

### LINKED ISSUE #632: Command `describe` does not include `metadata`.
**What steps did you take:**
As per https://github.com/carvel-dev/imgpkg/issues/124#issuecomment-823325183, I expect when we do `imgpkg describe -b ...`, if the bundle have `metadata` fields, it it will be printed as written in `imgpkg/bundle.yml`.

I put some data:
```
$ cat .imgpkg/bundle.yml 
metadata:
  oslManifestURL: ...
```
before I do `imgpkg push -b ...`, but it's not appearing when I do `imgpkg describe -b ...`:
```
$ imgpkg describe -b ... --output-type yaml
sha: sha256:bb59ca320f34063174bc9c00bc6f5b795d223c3b9b965851038ce85ebe4eb1d2
content:
...
image: ...
metadata: {}
origin: ...

Succeeded
```

**What happened:**
The `metadata` information is not printed.

**What did you expect:**
The `metadata` information is printed.

The [code](https://github.com/carvel-dev/imgpkg/blob/870f8123934c43005e118c873ec447590309ab2a/pkg/imgpkg/v1/describe.go#L134) seems to never populate this, just put an empty map as place holder.

Can this please be rectified, so I can retrieve this information vie `impkg describe -b ...`?

If I do `imgpkg pull -b ...`, the retrieved `.imgpkg/bundle.yml` has the information, but I don't want to have to pull it as the bundle can be huge, it's just a waste of time and local storage to pull the whole bundle just for me to get the `.imgpkg/bundle.yml`.

**Environment:**

- imgpkg version (use `imgpkg --version`): 0.34.1
- Docker registry used (e.g. `Docker HUB`): Harbor, Artifactory
- OS (e.g. from `/etc/os-release`): Various (Ubuntu, CentOS, photon).

---
Vote on this request

clcc2019:
## Summary

- read `.imgpkg/bundle.yml` directly from each bundle image
- populate metadata, authors, and websites in `imgpkg describe` output
- preserve the existing empty metadata behavior when `bundle.yml` is absent

Fixes #632

## Testing

- `go test ./pkg/imgpkg/v1 -run '^TestDescribeBundleMetadata$' -count=1`\n- `go test ./pkg/imgpkg/v1 -run '^TestDescribeBundle$' -count=1`\n- `git diff --check`

### DIFF
--- a/pkg/imgpkg/bundle/bundle_metadata.go
+++ b/pkg/imgpkg/bundle/bundle_metadata.go
@@ -1,0 +1,97 @@
+// Copyright 2026 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package bundle
+
+import (
+	"archive/tar"
+	"fmt"
+	"io"
+	"path/filepath"
+
+	regv1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"sigs.k8s.io/yaml"
+)
+
+// BundleAuthor contains author information from a bundle's metadata file.
+type BundleAuthor struct {
+	Name  string `json:"name,omitempty"`
+	Email string `json:"email,omitempty"`
+}
+
+// BundleWebsite contains a website from a bundle's metadata file.
+type BundleWebsite struct {
+	URL string `json:"url,omitempty"`
+}
+
+// BundleMetadata contains the user-provided information in .imgpkg/bundle.yml.
+type BundleMetadata struct {
+	Metadata map[string]string `json:"metadata,omitempty"`
+	Authors  []BundleAuthor    `json:"authors,omitempty"`
+	Websites []BundleWebsite   `json:"websites,omitempty"`
+}
+
+// Metadata reads the metadata associated with the bundle.
+func (o *Bundle) Metadata() (BundleMetadata, error) {
+	img, err := o.checkedImage()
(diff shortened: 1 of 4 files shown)


joaopapereira (changes requested):
Need some linter clearing and also test are failing, do you mind fixing that?

Text describe still never prints metadata (bundleTextPrinter); only YAML does. Issue #632 used `-o yaml`, so OK for the fix, but I would prefer to have consistent UX output.

We are missing a test for “no bundle.yml → empty metadata” (claimed in description).



I was contemplating another question about this PR and that is that we have to read the image 2 times to get the bundle.yaml file and the images.yaml, when we already have the image downloaded locally in the SingleLayerReader. What I am debating here is if this is going to make the Read slower and that will affect the overall performance of imgpkg so that when we do describe it is faster. But as I was writing this I think I convinced myself that it is better to have a slower describe than a slower pull or copy. nevertheless I think it makes sense to maybe have SingleLayerReader to be able to have another function called ReadBundleInfo or something of the sorts and have a function called `func (o *SingleLayerReader) read(img regv1.Image, filename string) ([]byte, error)` that basically we can extract the logic of reading a file from a layer and after that `Read` would call o.read(img, ImagesLockFile) while the ReadBundleInfo would call the o.read(img, BundleMetadataFile)

PR #791, opened 2026-08-03
Reject a tar manifest whose descriptor carries no refs

arpitjain099:
`DescribedImage.Ref` and `DescribedImageIndex.Ref` both do this:

```go
func (i DescribedImage) Ref() string { return i.desc.Refs[0] }
```

`Refs` is unmarshalled from the `manifest.json` inside a tar (`TarReader.getIdsFromManifest` -> `NewImageRefDescriptorsFromBytes`), so it is whatever the file says. A descriptor with `"Refs": []`, or with the field absent, panics as soon as anything reads the ref:

```
panic: runtime error: index out of range [0] with length 0
```

`TarReader.Read` alone does not trigger it, but `PresentLayers` and the copy path both call `Ref()` on every item, so `imgpkg copy --tar` on a truncated or hand-edited bundle crashes rather than reporting a bad tar. Someone consuming an air-gapped bundle they did not build is exactly the case where the file cannot be assumed well-formed.

Neither `Ref()` has an error to return, and changing them to hand back an empty string only moves the failure to `regname.NewDigest("")` with a worse message. Since every descriptor written by `buildImage`/`buildImageIndex` carries exactly one ref, this checks the invariant while parsing the manifest and names the descriptor that is missing it:

```
Expected image descriptor sha256:aaaa to have at least one ref
```

Images and indexes nested inside an index are walked too, since `buildIndex` calls `NewDescribedImage` on those the same way.

Three cases added in `pkg/imgpkg/imagedesc`. Before the change, parsing a descriptor with no refs succeeds and `Ref()` panics; after it, parsing fails with the message above. `go test ./pkg/imgpkg/...` passes apart from `TestLabels` in `pkg/imgpkg/cmd`, which also fails on a clean checkout here because `IMGPKG_E2E_IMAGE` and `IMGPKG_E2E_RELOCATION_REPO` are not set.

### DIFF
--- a/pkg/imgpkg/imagedesc/image_ref_descriptors.go
+++ b/pkg/imgpkg/imagedesc/image_ref_descriptors.go
@@ -51,7 +51,47 @@ func NewImageRefDescriptorsFromBytes(data []byte) (*ImageRefDescriptors, error) 
 		return nil, err
 	}
 
+	for _, desc := range descs {
+		switch {
+		case desc.Image != nil:
+			if err := validateImageRefs(*desc.Image); err != nil {
+				return nil, err
+			}
+		case desc.ImageIndex != nil:
+			if err := validateImageIndexRefs(*desc.ImageIndex); err != nil {
+				return nil, err
+			}
+		}
+	}
+
 	return &ImageRefDescriptors{descs: descs}, nil
+}
+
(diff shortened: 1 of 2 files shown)


joaopapereira (changes requested):
The tests look like they did not run, so we need to make sure we do that.
Also in our tests we use the `require` library instead of using `t.Fatal`.

PR #795, opened 2026-09-22
Compare and cache registry hostnames case-insensitively

sameerforge:
## Summary

Registry hostnames are DNS names (RFC 1035/1123) and must be treated as case-insensitive. Several comparison and cache-key sites in `pkg/imgpkg/registry` and `pkg/imgpkg/imageset` instead relied on Go's default case-sensitive string equality, `filepath.Match`, and map keys. When a hostname is supplied with different casing across two independent sources — an env var vs. a CLI/image ref, an images-lock file vs. a `--to-repo` flag, or two references to the same registry typed with different case — imgpkg would silently:

- fail to match configured credentials and fall back to anonymous auth,
- skip the cross-repo blob-mount optimization and re-upload blobs unnecessarily, or
- miss the authenticator/transport cache and redundantly re-resolve credentials and rebuild the round tripper.

None of these are hard failures with an obvious error message — they're intermittent, casing-dependent behavior differences, which makes them easy to miss in testing and painful to diagnose in the field.

## Changes

- **`pkg/imgpkg/registry/auth/credentialprovider/keyring.go`** — `URLsMatch` now lowercases both operands before `filepath.Match`. This is the shared primitive `EnvKeychain.Resolve` (`env_keychain.go`) calls through, so it fixes env-var-based credential matching against a differently-cased image reference host.
- **`pkg/imgpkg/imageset/image_set.go`** — `imageBlobsCanBeMounted` now uses `strings.EqualFold` instead of `!=` when comparing the source and destination registry hosts, so cross-repo mount is no longer skipped purely due to casing.
- **`pkg/imgpkg/registry/registry.go`** — `SimpleRegistry.transport` now lowercases the registry key used for the in-memory `authn` cache.
- **`pkg/imgpkg/registry/transport.go`** — `MultiRoundTripperStorage.RoundTripper` (read) and `CreateRoundTripper` (write) now lowercase the registry/repository keys consistently, so two case-variant references to the same physical registry share one cache entry instead of two.

All changes are internal comparison/cache-key normalizations. No CLI flag, environment variable name, or on-disk lock-file format changes — fully backward compatible.

## Test plan

- `go build ./...` — clean
- `go vet ./pkg/imgpkg/registry/... ./pkg/imgpkg/imageset/...` — clean
- `go test ./pkg/imgpkg/registry/... ./pkg/imgpkg/imageset/...` — full suite passes
- Verified pre-existing (unrelated) local failures in `TestInsecureRegistryFlag` reproduce identically on `develop` with these changes stashed, confirming they're caused by a missing `docker-credential-desktop` binary on the test machine, not by this change

## Related

Found via an internal case-sensitivity compliance audit covering FQDN/hostname, username, certificate alias, UUID, and DNS-name handling across the registry/auth code paths; this PR addresses all 5 open findings from that audit (all in the hostname/FQDN category — the other entity types had no findings in this codebase).


pkg/imgpkg/registry/transport.go
```diff
@@ -61,24 +61,29 @@ func (r *MultiRoundTripperStorage) RoundTripper(repo regname.Repository, scope s
 	// Maybe we should check to make sure only 1 repository is present in the scopes
 	method := s[2]
 
-	if _, ok := r.transports[repo.RegistryStr()]; !ok {
+	// Registry hostnames are DNS names (RFC 1035/1123) and must be used as case-insensitive cache keys,
+	// consistently with CreateRoundTripper below.
+	registryKey := strings.ToLower(repo.RegistryStr())
+	repositoryKey := strings.ToLower(repo.RepositoryStr())
```
himsngh: same here?

pkg/imgpkg/registry/registry.go
```diff
@@ -285,7 +286,8 @@ func (r *SimpleRegistry) writeOpts(ref regname.Reference) ([]regremote.Option, e
 // transport Retrieve the RoundTripper that can be used to access the repository
 func (r *SimpleRegistry) transport(ref regname.Reference, scope string) (http.RoundTripper, regauthn.Authenticator, error) {
 	registry := ref.Context()
-	registryKey := registry.Name()
+	// Registry hostnames are DNS names (RFC 1035/1123) and must be used as case-insensitive cache keys.
+	registryKey := strings.ToLower(registry.Name())
```
himsngh: This is a similar issue to what @joaopapereira has pointed out here https://github.com/carvel-dev/kapp-controller/pull/1863/changes/8fe7831be1a405dd3b778dd1f0064533cc8af11a#r4083736466?

sameerforge: Good catch! Fixed in [720d6e66](https://github.com/carvel-dev/imgpkg/pull/795/commits/720d6e660ebc28195b9173aad0e5789ddcae7f1c) — only registry hostnames should be case-insensitive per RFC 1035; repository paths are case-sensitive (aligns with the kapp-controller fix).

pkg/imgpkg/registry/transport.go
sameerforge: Fixed in [720d6e66](https://github.com/carvel-dev/imgpkg/pull/795/commits/720d6e660ebc28195b9173aad0e5789ddcae7f1c)

himsngh (approved):
LGTM!

